Issue #63, opened 2017-10-16
Mysterious startup settings change


shaomeng:
The msg box in the screenshot pops out mysteriously...

I just opened vapor, loaded the Katrnia data set, and clicked on the "Navigation" tab, then it shows up! 

![screenshot from 2017-10-16 11-03-44](https://user-images.githubusercontent.com/814885/31625074-fb1eea26-b261-11e7-8ac1-203438fee656.png)

shaomeng:
fixed by commit d4c9283

PR #105, opened 2017-10-18
quick hack on issue #63

shaomeng:
Please review this fix, and if everything looks right, I'll merge it!


shaomeng:
Ok, this fix seems really simple so I'll merge it.

But this pull request, then review, then merge steps should be really useful!

PR #116, opened 2017-10-19
fix issue 38

shaomeng:


shaomeng:
This issue seems to be there still with 1st time start... Let me investigate a little more...

PR #124, opened 2017-10-19
we created a new file. It fixed #123

### LINKED ISSUE #123: test: we need an awesome file!

shaomeng:
fix #123

### DIFF
--- a/awesome.file
+++ b/awesome.file
@@ -1,0 +1,1 @@
+I'm awesome


awesome.file
```diff
@@ -0,0 +1 @@
+I'm awesome
```
shaomeng: why use a global var here?

Issue #120, opened 2017-10-19
Vapor reports error if user cancels opening a session.


shaomeng:
Click File->Open to try to open a session, but bail out by clicking "Cancel."

Vapor reports an error at this point, but it shouldn't.

shaomeng:
This is duplicate as #99 .

Interestingly, #99 was fixed, but the fixes are gone mysteriously... Need a git guru to help figure it out!

PR #125, opened 2017-10-19
detect null string in file selectors

shaomeng:
This pull request fixes bug #120


shaomeng:
Given both @StasJ and @sgpearse  are happy with this pull request? 
I'll merge it in by the end of the day.

Issue #55, opened 2017-10-16
Creating multiple visualizers crashes the application


clyne:
Creating multiple visualizers crashes the application

To reproduce:
1. Load Duku.nc
2. Create and enable a TwoDData renderer
3. Create a new visualizer
4. In the new visualizer create and enable a second TwoDData render instance. The application will crash

Platforms:
Tested on Mac

PR #133, opened 2017-10-22
Fix Issue #55

StasJ:
…sualizers


clyne:
This commit has been made irrelevant by feature_88

clyne:
Stas, it looks like we still have the version of ControlExec::NewVisualizer() that takes a second argument, 'withDatasets'. It doesn't look like withDatasets is used inside of NewVisualizer or needed, and should be removed.

thanks - jc

PR #174, opened 2017-10-28
Increment

clyne:


clyne:
There is still a problem with the ECHAM data set. The latitude coordinate is reversed, running from north to south. vapor want's coordinate values to always be montonically increasing

Issue #178, opened 2017-10-30
cf2vdc: assert in RunWriteThreadCompressed


clyne:
cf2vdc is throwing an assert in RunWriteThreadCompressed

To reproduce:
1. Run cf2vdfccreate on /glade/p/VAST/VAPOR/Data/CAM/AdamPhillips/F2000C5_new120_025/FV_768x1152.bilinear.F2000C5_new120_025.cam.h0.2004-01.nc:

cf2vdccreate FV_768x1152.bilinear.F2000C5_new120_025.cam.h0.2004-01.nc fv.nc

2. Convert the data set with cf2vdc:

cf2vdc FV_768x1152.bilinear.F2000C5_new120_025.cam.h0.2004-01.nc fv.nc

Plaforms tested:
?

PR #181, opened 2017-10-30
Fix Issue #178

StasJ:


StasJ:
Confirmed that the dataset listed in the issue now converts and can be opened/rendered in vapor.

clyne:
Stas, this all looks fine, but I think that there is more going on here. The problem is non-determinstic: if you run the command several times it will assert with different variable. Yes, the code is threaded, but it's threaded only within each variable: i.e. each variable is processed in parallel, but there is a synchronization step in between different variables. 

It would be could to run all of the data sets again before closing this bug. FYI I've fixed some of the other bugs that were being reported, so you should sync with master first.

Issue #220, opened 2017-11-06
Contours: Only timestep 0 gets drawn


sgpearse:
Contours is currently only rendering timestep 0.

To Reproduce:
1) Load KatrinaMaster.nc
2) Spin up a new Contour instance
3) Enable it with default settings and advance in time.  Only timestep 0 is rendered.

PR #222, opened 2017-11-06
Fix for broken timestepping in Contours

sgpearse:
Fix for the following:

https://github.com/NCAR/VAPOR/issues/220

### DIFF
--- a/include/vapor/ContourRenderer.h
+++ b/include/vapor/ContourRenderer.h
@@ -87,11 +87,20 @@ public:
 	bool cacheIsValid(int timestep); 
 private:
 
-	int GetCurrentTimestep() {return 0;}
+	int GetCurrentTimestep();
 
 	//! Invalidate the cache of contours at a single time step
 	//! \param[in] timestep is the Time Step to be invalidated.
 	void invalidateLineCache(int timestep);
+
+	// Check to make sure a point is within a variable's bounds
+	// \param[in] x coordinate
+	// \param[in] y coordinate
+	// \param[in] z coordinate
+	// \param[in] Vector describing minimum data bounds
+	// \param[in] Vector describing maximum data bounds
+	bool pointWithinBounds(double x, double y, double z,
+		vector<double> varMin, vector<double> varMax);
 	
 	//! Rebuild the entire line cache at the current time step
 	//! \param[in] dmgr Current DataMgr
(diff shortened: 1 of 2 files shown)


sgpearse:
> The derived GetCurrentTimestep() method seems to be doing exactly what the baseclass version is doing. Is it necessary?

Whoops.  I was observing how the TwoDDataRenderer acquires the current timestep and it didn't occur to me that this was in the base class.

> Why is pointWithinBounds needed? Grid::GetValue() should return a MissingValue if the point is outside of the grid, and you are only selecting points that are inside of the user-selected bounding box (GetBox()).

I didn't articulate this earlier in your office; that the box will extend outside of the variable domain as time changes in Katrina.  There's currently an _assertion_ that triggers if we request a value outside of the domain, so I figured I was doing something illegal, and made a new function to enforce that rule.

If the Box is supposed to be clipped to the variable domain (as I believe it should), then this sounds like a new bug as well.

clyne:
> On Nov 6, 2017, at 6:12 PM, sgpearse <notifications@github.com> wrote:
> 
> The derived GetCurrentTimestep() method seems to be doing exactly what the baseclass version is doing. Is it necessary?
> 
> Whoops. I was observing how the TwoDDataRenderer acquires the current timestep and it didn't occur to me that this was in the base class.
> 
> Why is pointWithinBounds needed? Grid::GetValue() should return a MissingValue if the point is outside of the grid, and you are only selecting points that are inside of the user-selected bounding box (GetBox()).
> 
> I didn't articulate this earlier in your office; that the box will extend outside of the variable domain as time changes in Katrina. There's currently an assertion that triggers if we request a value outside of the domain, so I figured I was doing something illegal, and made a new function to enforce that rule.
> 
> If the Box is supposed to be clipped to the variable domain (as I believe it should), then this sounds like a new bug as well.

We should not be throwing an assert if you request a value outside of the domain. This will break things elsewhere. It would be best to fix the root problem, rather than provide a workaround in one of the renderers. Do you have a test case that reproduces this?


> 
> —
> 

John Clyne
clyne@ucar.edu

sgpearse:
To reproduce,

Method 1 (same as bug [220](https://github.com/NCAR/VAPOR/issues/223))
1) On the master branch, create a Barb renderer with KatrinaMaster.nc
2) Advance to timestep 7

Method 2
1) On branch issue_220, do a pull
2) Create a Contour renderer with KatrinaMaster.nc
3) Advance to any timestep other than 0

Issue #273, opened 2017-11-08
plot utility: throws exception when multiple data sets loaded


clyne:
The plot utility won't launch and vapor throws an exception when multiple data sets are loaded

To reproduce:
1. Import the WRF data set Data/CoupledWrfRoms/wrfout_d01_1982-01-31_10_00_00
2. Import the CF data set Data/CoupledWrfRoms/kiost_roms_nwp12_avg_y1982-002.nc 
3. Launch the Tools/Plot utility

PR #364, opened 2017-11-10
work around issue #273

shaomeng:
The cause of issue #273 isn't multiple data sets, but data sets with no 3D variables.


shaomeng:
I'm going to merge this pr and continue with my bug fixes.

Issue #414, opened 2017-11-16
Statistics: Timestep Selection


sgpearse:
I believe that Statistics is missing a timestep in the min/max selector for time.

To reproduce:
1) Run Vapor with KatrinaMaster.nc
2) Note that KatrinaMaster.nc holds 10 timesteps, 0 to 9
3) Open Statistics, the timestep selectors max out at 8

shaomeng:
Submitted a PR to work around it in the Statistics tool, but there's further an improvement needed for DataMgr, as described in issue #429 .

shaomeng:
should be fixed by PR #435

Issue #429, opened 2017-11-18
Function name and results don't match


shaomeng:
https://github.com/NCAR/VAPOR/blob/bf2d043fd8562097a75aed76e18de4f585f56940/include/vapor/DataMgr.h#L376-L379

The above function, `GetNumTimeSteps()`, by its name and the comments, should return 10 for a data set with 10 time steps.
However, it actually returns 9, as we can see from the implementation:
https://github.com/NCAR/VAPOR/blob/bf2d043fd8562097a75aed76e18de4f585f56940/lib/vdc/DataMgr.cpp#L608-L611.

PR #430, opened 2017-11-18
fix issue #414

shaomeng:


clyne:
Sam, the reason this bug is occurring is that one of the overloaded DataMgr::GetNumTimeSteps() is returning the number of time steps, minus one. The correct fix would be revert the changes in this commit, and change DataMgr::GetNumTimeSteps() so that it returns the correct value. Note that DataMgr::GetNumTimeSteps() is overloaded, and one version returns the correct value, and one does not.

shaomeng:
I've filed an issue (#429) on it, and will go ahead and correct it. At the same time I'm going to close this PR.

PR #505, opened 2017-12-06
Fix for issue 489

sgpearse:


sgpearse:
Good point, should be all set now.

PR #508, opened 2017-12-08
Refactorization of the ContourAppearanceSubtab

sgpearse:
My notes on these changes may help to understand what my intention was here.  See the link below.

https://docs.google.com/a/ucar.edu/document/d/1w-7O75x8eauIITKpIh_sGDXvIa_5HVO0qUp0mIVs6-o/edit?usp=sharing

### DIFF
--- a/apps/vaporgui/ContourAppearanceGUI.ui
+++ b/apps/vaporgui/ContourAppearanceGUI.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>403</width>
-    <height>1042</height>
+    <height>1070</height>
    </rect>
   </property>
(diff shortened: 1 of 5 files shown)


sgpearse:
The current changes should allow the user to apply any spacing value that they want, regardless of the data extents, or transfer function extents.

This required subclassing the Combo class with the SpacingCombo class.  The Combo class imposes bounds on its accepted values.  Since we want to let the user enter values that can go past reasonable bounds, we needed reimplementations of a couple functions to break these rules.

Moderate testing has been performed, and there is one bug that I'm looking at.

I think this is closer to an ideal solution.  If not, let's keep iterating.

PR #531, opened 2018-01-05
Contour free form

sgpearse:


sgpearse:
This branch includes commits that I've made in the past for reference.  Currently, the user may select the Contour Minimum, Count, and Spacing arbitarily without any corrections to ensure that all contour values are being rendered.

PR #532, opened 2018-01-05
Render holder refactor

sgpearse:
Refactored the RenderHolder to use the new VaporTable widget, which is included in this pull request.


clyne:
Looks good, a couple of minor comments:

1. Follow our naming convention for public methods on VaporTable (e.g. rowCount -> RowCount, etc.)

2. Rather than providing public access to what should probably be internal members via the cellWiget() method, make cellWidth a private member function (don't expose the VaporTable internals to the world) and add the ability to highlight the current row to the table itself as a configuration option.

sgpearse:
I've modified the TransformTable to use VaporTable in the current pushed repository of renderHolderRefactor.

Also...

> Follow our naming convention for public methods on VaporTable (e.g. rowCount -> RowCount, etc.)

No problem.

> Rather than providing public access to what should probably be internal members via the cellWiget() method, make cellWidth a private member function (don't expose the VaporTable internals to the world) and add the ability to highlight the current row to the table itself as a configuration option.

Also not a problem.  I implemented this outside of VaporTable because the only table that highlights rows/columns is the one in the RenderHolder, but it's no big deal to incorporate that feature into VaporTable.

As for making the CellWidget method private:  Is this granting the user too much power over the table?  Our RangeCombos provide public functions to get sliders and line edits, so I thought that retrieving cell-widgets would be desirable for the same reason.

PR #537, opened 2018-01-17
Issue 530

sgpearse:
GeometryWidget now supports single-point-mode when given the appropriate flags in Reinit()


clyne:
Need to resolve Git conflict

PR #563, opened 2018-03-07
Issue 303 axis annotations

sgpearse:
Revision 2 of axis annotations.

Fixes include:
✓ Remove setVizFeatureTextChanged slot
✓ No more openGL calls in EventRouters
✓ Remove vizfeatureReturnPressed slot, replaced where appropriate
✓ Fix axis arrow location line-edits
✓ Update guis for arrows and frames
✓Refactored TimeAnnotations
✓Now using ErrorReporter::MSG_ERR

Postponed actions include:
**X** Make copyRegionFromRenderer() a protected base class method - I plan on creating a promoted widget for this
**X** Make updateColorHelper() a protected base class method - I plan on promoting this too


sgpearse:
John, here are my notes on the talking-points that you brought up in the previous pull request.  Your comments are in the quoted text.

**Refactorization on 2/23/2018:**
✓ Remove setVizFeatureTextChanged slot
✓ Remove vizfeatureReturnPressed slot, replace where appropriate
✓ Fix axis arrow location line-edits
✓ Redo for arrows, frames and time annotations
Still an error being reported from DataMgrUtils?  Could not convert point x y to Lon/Lat.
Is ErrorReporter::MSG_ERR the correct display mechanism?  I think so.
Make ColorPicker promoted widget


**Notes 2/21/2018**
Make a rough refactorization pass on VizFeatureEventRouter

 

> remove void VizFeatureEventRouter::confirmText();
> remove void VizFeatureEventRouter::_confirmText();
> remove code after return in VizFeatureEventRouter::_updateTab()

Agreed.  Should I take a pass at refactoring the entire thing in general?  I did notice other nuisances during my work.
 
DONE

> make VizFeatureEventRouter::copyRegionFromRenderer() a protected, reusable
> method on EventRouter or RenderEventRouter base class?
> ditto VizFeatureEventRouter::updateCopyRegionCombo() 

That's one option.  I had another in mind, which is to make a promoted Copy Region widget, which is currently embedded in the GeometryWidget.

Question: At what point do we set this kind of logic in a promoted widget, vs embedding the logic in the event router?

Do as is reported in issue summary.  Update issue summary to fix AxisAnnotation copy code.
 
> Error handling wrong (checks boolean, not negint), and a no-op in
> VizFeatureEventRouter::convertPCSToLonLat()
> ditto VizFeatureEventRouter::convertLonLatToPCS()

Fixed!

> Call error report display mechanism.
 
Probably should not make openGL calls in eventrouters (glGetDoublev)
 
How come?  Where should we do it?

Hard code it.  This may be benign.  Rule of thumb: only call gl functions from within a GL context (GLWidget)

> Isn't VizFeatureEventRouter::setColorHelper() a "Get" accessor?

Make setColorHelper and updateColorHelper() protected base class
methods? Don't we replicate this in almost every event router?

Make ColorPicker promoted widget

We do not use this in any EventRouter.  This is a wrapper for QColorDialog, which is used in:
ColorbarWidget.cpp
ControlPointEditor.cpp
TFWidget.cpp
VizFeatureEventRouter.cpp

getColorHelper and setColorHelper are legacy functions that just 1) set a QLineEdit's color, and 2) prompt user to select a color.  They probably should be protected EventRouter methods though.
 
Scan the code, see if we'll ever use it.  Add if we do.  Also change setColorHelper to getColorHelper.

> foo.{cpp,h}
> What are these?
> 

Mistakes
 

> VizFeatureRenderer.cpp
> VizFeatureRenderer::drawRegionBounds(): is this deprecated? Can it be
> removed?

This has not been implemented yet.  Currently assigned to stas.

Delete!

> error handling in VizFeatureRenderer::convertPointToLonLat()

 
Fixed.  Needed to check for rc<0, not !rc.

PR #568, opened 2018-03-12
Issue 303 axis annotations

sgpearse:


sgpearse:
*note to self - removed AnnotationsRenderer.cpp/h

sgpearse:
Should be good now :)

PR #574, opened 2018-03-16
Implemented 573

StasJ:
QPushButtonWithDoubleClick is required to be a public class. This is an ugly workaround but it is required by Qt.


shaomeng:
This `QPushButtonWithDoubleClick` seems to be something we can reuse at a later time. Could we put it in its own file?

PR #667, opened 2018-04-17
Stas

StasJ:
Fixes/Changes:

- Fixed mismatched new[] / delete[]
- Added GetVarDimLens to VDC c interface
- Template pattern definitions for DC::_getVarTemplate were missing. This would could cause the linker to fail depending on how it was compiled.


shaomeng:
This PR has such a cool name!

Issue #645, opened 2018-04-16
Settings: un-careful lock window dimensions cause vapor unusable. 


shaomeng:
To reproduce: 
1. Open VAPOR, go to Settings, check "lock window dimensions."
2. Change the width and height to be 400. 
3. Restart VAPOR.
4. VAPOR becomes too small to use, and it's not able to resize.
5. Also because it's too small, you cannot change the "lock window dimensions" settings anymore. 
6. VAPOR on your machine is 400x400 forever! 

![screenshot from 2018-04-16 15-47-47](https://user-images.githubusercontent.com/814885/38837007-9fa96c00-418d-11e8-80e9-cc5762880167.png)

Issue #620, opened 2018-04-13
Unreasonable setting value restrictions


shaomeng:
I found the following setting value restrictions *unreasonable*:

1. https://github.com/NCAR/VAPOR/blob/1432d5aab8269f59d6263654887e52ebb2d30570/apps/vaporgui/SettingsEventRouter.cpp#L66-L68
Why 0 thread is allowed? Why aren't more than 8 threads allowed? 

2. https://github.com/NCAR/VAPOR/blob/1432d5aab8269f59d6263654887e52ebb2d30570/apps/vaporgui/SettingsEventRouter.cpp#L70-L72
Why 0MB cache allowed? Why don't allow more than 100GB cache? 

3. https://github.com/NCAR/VAPOR/blob/1432d5aab8269f59d6263654887e52ebb2d30570/apps/vaporgui/SettingsEventRouter.cpp#L74-L80
Why 0 is allowed as window dimensions? What will happen if a user enters 0?

sgpearse:
make maxInt values for all maxima

PR #668, opened 2018-04-17
Resolve issue #645

shaomeng:
Also,

I added a new class, `QIntValidatorWithFixup`, that can automatically fix input that is out of range.


shaomeng:
This PR also partially fixes issue #620

Issue #643, opened 2018-04-16
Settings: Autosave file dialog should verify that a regular file with appropriate extension


clyne:
The Autosave file location dialog allows the user to change the session file path to an arbitrary path. It should require a file ending with a ".vs3" suffix because the file filter for reading session files requires this extension.

shaomeng:
A regression is introduced... I'm re-opening this issue.

clyne:
Sam, please provide a precise step-by-step process that demonstrates the problem you're seeing. It seems to work fine for me.

shaomeng:
John, I'm actually curious how it works fine for you. 
Here are the steps to repeat it:
1. Open Katrina, go to "Settings" tab.
2. Change the default session file from `VaporAutoSave.vs3` to `VaporAutoSave.v3`.
4. A new `VaporAutoSave.v3` file appears on the disk.

clyne:
This bug still exists. I'm reopening it. The steps provided by Sam still result in a file with an invalid extension

shaomeng:
Aha, I think I fixed the `Save Session File` button, but didn't fix `Auto Save Session File`. Let me re-fix...

PR #676, opened 2018-04-18
fix issue #643

shaomeng:


shaomeng:
I'm going to merge this one since my next bug fix depends on this one.

Issue #614, opened 2018-04-12
Settings: default thread number needs to be more intelligent. 


shaomeng:
The default thread number seems to be hard coded as 2. 
A more desirable behavior is to set as the number of CPU cores. 
(The screenshot shows this issue on cisl-vapor machine.)

![screenshot from 2018-04-12 11-30-59](https://user-images.githubusercontent.com/814885/38693885-83df6f92-3e45-11e8-943a-4aa11757ae42.png)

shaomeng:
Why do we allow 0 threads, and don't allow more than 8 ???
https://github.com/NCAR/VAPOR/blob/1432d5aab8269f59d6263654887e52ebb2d30570/apps/vaporgui/SettingsEventRouter.cpp#L67

shaomeng:
The default value is 0 right now.

PR #680, opened 2018-04-18
fix issue #614

shaomeng:
fix issue #614

### DIFF
--- a/apps/vaporgui/SettingsGUI.ui
+++ b/apps/vaporgui/SettingsGUI.ui
@@ -471,7 +471,7 @@
          <item>
           <widget class="QLabel" name="label_9">
            <property name="text">
-            <string>Number of threads (0 for num. of CPU cores)</string>
+            <string>Number of threads (0 for available num. of CPU cores)</string>
            </property>
           </widget>
          </item>
--- a/apps/vaporgui/SettingsParams.cpp
+++ b/apps/vaporgui/SettingsParams.cpp
@@ -411,8 +411,7 @@ void SettingsParams::SetCurrentPrefsPath(string pth){
 }
 
 int SettingsParams::GetNumThreads() const {
-	long val = GetValueLong(_numThreadsTag, 2);
-	if (val < 0) val = 0;
+	long val = GetValueLong(_numThreadsTag, 0);
 	return((int) val);
 }
 
@@ -421,7 +420,6 @@ void SettingsParams::SetNumThreads(int val) {
 }
 
 int SettingsParams::GetFontSize() const {
-//	return (int)GetValueDouble(_fontSizeTag, 
 	return 24;
 }
 
--- a/apps/vaporgui/SettingsParams.cpp
+++ b/apps/vaporgui/SettingsParams.cpp
@@ -411,8 +411,9 @@ void SettingsParams::SetCurrentPrefsPath(string pth){
 }
 
 int SettingsParams::GetNumThreads() const {
-	long val = GetValueLong(_numThreadsTag, 0);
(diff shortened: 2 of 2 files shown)


shaomeng:
John, 
- For `SettingsParams::GetFontSize()` I only removed a line of comment, and did not change its behavior.
- For `SettingsParams::GetNumThreads()`, it shouldn't be a Parameters class's job to examine if a parameter is valid; its sole job should be to record and report parameters. That said, I introduced an improved `QIntValidatorWithFixup` that won't allow negative values to be set at the first place:https://github.com/NCAR/VAPOR/blob/15dfde457db82a2accec108292c5763b2e774fc1/apps/vaporgui/QIntValidatorWithFixup.cpp#L3-L18

shaomeng:
Two points on the responsibilities of a Params class.
- Even if it's the responsibility of a Params class to validate a parameter, it should be validated when the `SetNumOfXXX()` function is called, but not `GetNumOfXXX()` is called. 
- Further, I still think it's a bad idea to ask a Params class to validate a parameter, because of the following reason.

Imagine the propagation when  user sets a parameter. A QT widget (for example a `QEditBox`) first receives a value, most likely a string. Then this string is parsed into a number in a Vapor class (e.g. a `SettingsTab`). Then a Vapor class passes this number to a params class (e.g. a `SettingsParams`). 
If you only validate this parameter in a params class, then once you decide to correct this value, the corrected value needs to travel this entire pipeline back to `SettingsTab` and `QEditBox`, which involves a good number of functions calls. 
Alternatively, if you validate it by`QEditBox` at the very beginning, everybody else is simply working with a valid value. 

In the case of `RenderParams::GetConstantColor()`, why do you allow `RenderParams::SetConstantColor()` to set some invalid color at the first place? If it's a user input, why do you allow a user to input 2.0 at the first place?

clyne:
> On Apr 19, 2018, at 10:38 AM, Samuel Li <notifications@github.com> wrote:
> 
> Two points on the responsibilities of a Params class.
> 
> 	• Even if it's the responsibility of a Params class to validate a parameter, it should be validated when the SetNumOfXXX() function is called, but not GetNumOfXXX() is called.

That might be true if the params state wasn’t maintained externally (in session files). Hence, unlike a class object that maintains it’s state in memory and the C++ language can be used to enforce the validity of member variables, the state (think of class member data) of the params classes can be changed by things other than accessor Set methods. For example, undo/redo, or loading a session file. 

> 	• Further, I still think it's a bad idea to ask a Params class to validate a parameter, because of the following reason.
> Imagine the propagation when user sets a parameter. A QT widget (for example a QEditBox) first receives a value, most likely a string. Then this string is parsed into a number in a Vapor class (e.g. a SettingsTab). Then a Vapor class passes this number to a params class (e.g. a SettingsParams).
> If you only validate this parameter in a params class, then once you decide to correct this value, the corrected value needs to travel this entire pipeline back to SettingsTab and QEditBox, which involves a good number of functions calls.
> Alternatively, if you validate it byQEditBox at the very beginning, everybody else is simply working with a valid value.

I’m not suggesting that validation only occurs in the params class. However, if the params Get methods perform some rudimentary validation it simplifies their use, by reducing the amount of error checking that needs to be performed elsewhere in the code.

> 
> In the case of RenderParams::GetConstantColor(), why do you allow RenderParams::SetConstantColor() to set some invalid color at the first place? If it's a user input, why do you allow a user to input 2.0 at the first place?
> 

We don’t. SetConstantColor does clamp values. See my comments above. 
> —
> 

John Clyne
clyne@ucar.edu

PR #682, opened 2018-04-19
Issue 663

sgpearse:
Fixed Bug 663, regarding AxisAnnotations being applied to extents that changed when a renderer was enabled.  The coupling between AxisAnnotations and DataMgrs has also been removed.  We now support only one axis annotation for all datasets, that uses the currently selected projection string for lat/lon conversions.


sgpearse:
Should be good to go now.

Issue #679, opened 2018-04-18
File/folder selection dialog fails to examine the existence/permission of users' input


shaomeng:
We use many file/folder selection dialogs to take user's input to specify various files and directories. For example, to specify a captured image, or an autosave session file.

However, we don't examine if the user's input is valid. Then you can do a number of crazy things:

1. you can choose a file that you don't have read/write permission.
2. you can choose a directory that you don't have read/write permission.

shaomeng:
Update on 4/24/2018:
The selection dialog doesn't even seem to work in at least one occasion: 
![apr-24-2018 17-32-42](https://user-images.githubusercontent.com/814885/39219078-a41c67ac-47e5-11e8-93d7-ecfcdcb7730d.gif)

PR #688, opened 2018-04-20
Issue 679

sgpearse:
Fix for issue #679

### DIFF
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -1152,7 +1152,10 @@ void MainForm::sessionOpen(QString qfileName)
 	if (! qfileName.endsWith(".vs3")){
 		return;
 	}
-	FileOperationChecker::FileGoodToRead(qfileName);
+
+	if (!FileOperationChecker::FileGoodToRead(qfileName))
+		MSG_ERR(FileOperationChecker::GetLastErrorMessage().toStdString());
+
 
 	string fileName = qfileName.toStdString();
 	sessionOpenHelper(fileName);
@@ -1180,7 +1183,8 @@ void MainForm::_fileSaveHelper(string path)
 		return;
 	}
 
-	FileOperationChecker::FileGoodToWrite(fileName);
+	if (!FileOperationChecker::FileGoodToWrite(fileName))
+		MSG_ERR(FileOperationChecker::GetLastErrorMessage().toStdString());
 
 	SettingsParams* sParams = GetSettingsParams();
 	sParams->SetSessionDir(path);
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -1178,14 +1178,14 @@ void MainForm::_fileSaveHelper(string path)
 	}
     if( path.empty()) return;
 
-	if( _controlExec->SaveSession(path) < 0 )
-    {
-		MSG_ERR("Saving session file failed");
+	if (!FileOperationChecker::FileGoodToWrite(fileName)) {
+		MSG_ERR(FileOperationChecker::GetLastErrorMessage().toStdString());
 		return;
 	}
 
-	if (!FileOperationChecker::FileGoodToWrite(fileName)) {
-		MSG_ERR(FileOperationChecker::GetLastErrorMessage().toStdString());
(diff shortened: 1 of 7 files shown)


shaomeng:
I included within the FileOperationChecker to report an error because there are multiple ways to fail a file operation. The caller would not know exactly why the operation failed if he only receives a false return value. 
Moreover, I cannot think of a case where a file operation is about to fail, and the caller does NOT want to inform the user. 
Thoughts? @clyne @sgpearse

clyne:
> On Apr 20, 2018, at 12:26 PM, Samuel Li <notifications@github.com> wrote:
> 
> I included within the FileOperationChecker to report an error because there are multiple ways to fail a file operation. The caller would not know exactly why the operation failed if he only receives a false return value.

You can have the checker method return an error string or status flag as an argument. 

> Moreover, I cannot think of a case where a file operation is about to fail, and the caller does NOT want to inform the user.

Sure. You may want to use utilities to explore multiple options. Is there a readable file of some sort in the current working directory? No? How about in the home directory? 

In general things that return status are more useful if they don’t enforce an error reporting mechanism that may not be wanted in all use cases. That’s the model that we use throughout the codebase. 

jc

> Thoughts? @clyne @sgpearse
> 
> —
> 

John Clyne
clyne@ucar.edu

shaomeng:
John, that is a good point. I've changed the behavior of `FileOperationChecker` so that it 1) does not pop up a window, and 2) it records the latest error message.

sgpearse:
Should be good now.

sgpearse:
Should be good now.

apps/vaporgui/MainForm.cpp
```diff
@@ -1178,6 +1184,11 @@ void MainForm::_fileSaveHelper(string path)
 		return;
 	}
 
+	if (!FileOperationChecker::FileGoodToWrite(fileName)) {
```
shaomeng: This "FileGoodToWrite()" check happens AFTER `_controlExec` tries to write the file (L1181). It should, however, happen BEFORE `_controlExec` attempts to write.

apps/vaporgui/SettingsEventRouter.cpp
```diff
@@ -197,43 +198,110 @@ void SettingsEventRouter::_changesPerSaveChanged() {
...
 		QString::fromStdString( sParams->GetAutoSaveSessionFile() ),
-        tr("Vapor 3 Session Files (*.vs3)")
-    );
+		tr("Vapor 3 Session Files (*.vs3)")
+	);
 
-	if (! fileName.isEmpty()) 
-    {
+	bool goodToWrite = FileOperationChecker::FileGoodToWrite(fileName); 
+	if (! fileName.isEmpty() && goodToWrite) {
```
shaomeng: Testing both if `fileName` is empty and if it's good to write is redundant. That is, if a fileName were empty, it won't pass the "GoodToWrite" test.

lib/render/Visualizer.cpp
```diff
@@ -998,11 +998,13 @@ int Visualizer:: captureImage(string filename)
     else // write png files
     {
 		FILE* test = fopen((const char*)filename.c_str(), "wb");
-        fclose( test );
-		if (!test) {
+		if (!test)
```
shaomeng: `FileOperationChecker` should be used here to simplify and modulate code.

sgpearse: FileOperationChecker is in the GUI, and the Visualzier is in the render library.  Can't use it yet!

apps/vaporgui/SettingsEventRouter.cpp
```diff
@@ -197,43 +198,109 @@ void SettingsEventRouter::_changesPerSaveChanged() {
...
+	);
 
-	if (! fileName.isEmpty()) 
-    {
+	bool goodToWrite = FileOperationChecker::FileGoodToWrite(fileName); 
+	if (goodToWrite) {
 		sParams->SetAutoSaveSessionFile(fileName.toStdString());
 		_saveSettings();
 	}
+	else (!goodToWrite) {
```
shaomeng: I don't think this line will pass the compiler...

lib/render/Visualizer.cpp
shaomeng: that's true...

PR #748, opened 2018-04-26
fix warnings on linux

shaomeng:
This PR has the following changes:
1. It removes compiler flags that suppress certain warnings.
2. It adds global flags "-Wall -Wno-sign-compare" that reports warnings.
3. It fixes most VAPOR warnings on linux using GCC.

What is left:
1. A few warnings that need attention from Scott. I'll submit issues once this PR is approved.
2. Clang has additional warnings. We'll examines them in the next steps.


CMakeLists.txt
```diff
@@ -35,6 +35,13 @@ elseif (WIN32)
 	add_definitions (-DWIN32)
 endif()
 
+# compiler warning flags
+if (NOT WIN32)
+  SET( COMPILE_FLAGS "-Wall -Wno-sign-compare")
+  SET( CMAKE_CXX_FLAGS  "${COMPILE_FLAGS}" )
+  SET( CMAKE_C_FLAGS  "${COMPILE_FLAGS}" )
```
StasJ: We don't have any c code

apps/vaporgui/main.cpp
```diff
@@ -166,7 +166,6 @@ if (getenv("VAPOR_DEBUG"))
     fontFile = fontFile + "//arimo.ttf";
 
 	QFontDatabase fdb; 
-	int result = fdb.addApplicationFont(QString::fromStdString(fontFile));
```
StasJ: No

include/vapor/UnstructuredGridCoordless.h
```diff
@@ -137,9 +137,6 @@ void GetIndices(
   virtual const void *address() const {return this; };
 
   virtual bool equal(const void* rhs) const {
-	const ConstCoordItrUCoordless *itrptr = 
-		static_cast<const ConstCoordItrUCoordless *> (rhs);
-
```
StasJ: What?

lib/render/ImageRenderer.cpp
```diff
@@ -201,8 +200,6 @@ int ImageRenderer::GetMesh( DataMgr *dataMgr,
 		// Get the width and height of the image texture. These
 		// will be used to set the width and height of the mesh.
 		//
-		unsigned char *texture = (unsigned char *) _getTexture(dataMgr);
```
StasJ: This will break things.

CMakeLists.txt
shaomeng: hmm, how about this one:
https://github.com/NCAR/VAPOR/blob/53b1fa78b25c1f84e1c1ca8680c8b849de71e79f/lib/vdc/kdtree.c

StasJ: Fair point

apps/vaporgui/main.cpp
shaomeng: good call! 

Fixed in commit 104b214 .

include/vapor/UnstructuredGridCoordless.h
shaomeng: Agreed. 

That fix will go into a different PR though.

lib/render/ImageRenderer.cpp
shaomeng: More specifically?

shaomeng: fixed in commit 4e9502a

apps/vaporgui/ErrorReporter.cpp
```diff
@@ -197,7 +197,7 @@ string ErrorReporter::GetSystemInformation()
 	ret += "OS: " + string(info.sysname) + " " + string(info.release) + " " + string(info.version) + "\n";
 	ret += "Distro:\n";
 	char buffer[128];
-    	FILE *pipe = popen("lsb_release", "r"), pclose;
+    	FILE *pipe = popen("lsb_release", "r");
```
clyne: pipe is not being closed

lib/wasp/WASP.cpp
```diff
@@ -455,12 +456,15 @@ vector <size_t> vdiff(vector <size_t> a, vector <size_t> b) {
 	for (int i=0; i<a.size(); i++) c[i] = a[i] - b[i];
 	return(c);
 }
+#endif
 
 // Determine POD type 
 //
+#ifdef VAPOR3_0_0_ALPHA
```
StasJ: If this is commented out a call to NetCDFType((float)3.0f) will be promoted to NetCDFType(double) without warning

shaomeng: fixed in commit 4d331b0

apps/vaporgui/ErrorReporter.cpp
shaomeng: That's a bug of the code. I intended to only address warnings in this PR, unless you think I should fix bugs along the way.

shaomeng:
Don't merge yet, more fixes coming...

apps/vaporgui/ErrorReporter.cpp
shaomeng: since this bug is so easy to fix, I fixed via commit 499f0ba

PR #793, opened 2018-05-03
Enabled anti-aliasing for contours

StasJ:
## Without anti-aliasing
(This is exacerbated by aliasing from github resizing the image)

screen shot 2018-05-03 at 2 22 37 pm

## With anti-aliasing
screen shot 2018-05-03 at 2 22 45 pm


StasJ:
To get high quality anti-aliased contours, the renderer needs to be treated as a transparent surface.

PR #802, opened 2018-05-04
Issue 643

shaomeng:
These changes should address issue #643.


shaomeng:
" is there something special about the file dialogs here that warrants special attention? "
-- Well, the fact that we enforce a certain extension causes complication. This enforcement largely requires our implementation with just a little bit help from QT. This is what you see in this PR. Is this the special attention you're referring to? 

"Didn't you write a FileOperationChecker class that we could use here?"
-- The FileOperationChecker class only tells you right or wrong, but does NOT fix any problem for you. The majority of code here is to fix any potential problems. 

"The _chooseAutoSaveFile() and _autoSaveFileChanged() are largely cut-n-paste. Can't we use a helper function and eliminate code duplication?" 
-- A helper function would be possible; but that helper function is only reused by 2 other functions, which are these two. Is it worth doing so?

shaomeng:
John, this branch is ready to merge with the latest commit.

Issue #721, opened 2018-04-25
Scene transforms unable to view.


shaomeng:
See screenshot:

![screenshot 2018-04-25 15 36 16](https://user-images.githubusercontent.com/814885/39274258-9c918c10-489e-11e8-9c38-0509c54002a0.png)

shaomeng:
This issue is fixed, thus closing.

shaomeng:
The height of this table is sufficiently big, but the width isn't, so you cannot see the values in the last cell. Moreover, there is actually quite sufficient real estate in the GUI in terms of width...

![screenshot from 2018-05-04 12-27-46](https://user-images.githubusercontent.com/814885/39645877-ec9893d0-4f96-11e8-9705-7fee077020c9.png)

Issue #718, opened 2018-04-25
Barb: Geometry displays incomplete on Ubuntu


shaomeng:
The numbers aren't able to read; see attached.

![screenshot 2018-04-25 15 13 47](https://user-images.githubusercontent.com/814885/39273226-6bbba31c-489b-11e8-8af8-bb1fb8b7a709.png)

shaomeng:
This bug is resolved, thus closing.

shaomeng:
I think this was resolved at one point, but it's back! 

![screenshot from 2018-05-04 12-27-46](https://user-images.githubusercontent.com/814885/39646182-d2c29b12-4f97-11e8-9032-19dbbf327532.png)

Issue #803, opened 2018-05-04
Settings: GUI needs more polish on Ubuntu


shaomeng:
Currently a line of text isn't displayed in complete. 

![screenshot from 2018-05-04 16-12-00](https://user-images.githubusercontent.com/814885/39655323-2ae21aa2-4fb6-11e8-8e5b-d9d927dd84c7.png)

shaomeng:
Update on 6/11/2018:
This behavior remains the same.

shaomeng:
Cannot test due to issue #1082

sgpearse:
Fixed.

PR #832, opened 2018-05-07
Further refinement to the Ubuntu's GUI

sgpearse:
Fixes for #718 and #721


shaomeng:
Scott, since you're already dealing with these GUI issues, would it make sense to also look at issue #803 ?

PR #895, opened 2018-06-07
Issue 551

sgpearse:
Pulled the CopyRegion option out of the GeometryWidget and made it into its own PromotedWidget, named the CopyRegionWidget.


sgpearse:
Gotcha.  Should be good now.

Issue #912, opened 2018-07-09
Compiler warning introduced by a recent PR. 


shaomeng:
The following line has introduced a compiler warning on Ubuntu. 
Per our code policy, PRs should **NOT** introduce warnings. 

https://github.com/NCAR/VAPOR/blob/040ecf6fd59eb883d4889ebb9a25e10359870d54/apps/vaporgui/Manip.cpp#L1072

PR #913, opened 2018-07-10
Issue 829 - Better scaling for barbs

sgpearse:


shaomeng:
1. Scott, is this a minor improvement PR that only requires one reviewer, or a bigger PR that would require all members to review?

2. There are new compiler warnings in this branch, in addition to the ones reported in issue #912  and issue #911 . I'd suggest we achieve warning free on the master branch first before we accept any new PR.

clyne:
Sam, I will add reviewers (testers) if I think it necessary.

jc

> On Jul 10, 2018, at 11:57 AM, Samuel Li <notifications@github.com> wrote:
> 
> 	• Scott, is this a minor improvement PR that only requires one reviewer, or a bigger PR that would require all members to review?
> 
> 	• There are new compiler warnings in this branch, in addition to the ones reported in issue #912 and issue #911 . I'd suggest we achieve warning free on the master branch first before we accept any new PR.
> 
> —
> 

John Clyne
clyne@ucar.edu

sgpearse:
This is passing my tests except with CAM data.

Instead of having the scale default to 1, I changed it to range between 1-100, with the default being 50.  Let me know what you think.  If the default was 1, would we want the maximum to be 2?  Let me know if 1-100 feels right.

PR #969, opened 2018-08-22
New DVR

shaomeng:


shaomeng:
Hi John,

A few comments:

1. I feel the 80 character rule is really limiting for modern c++ with name spaces, qualifiers, etc. I honest think a hard 80 character limit would reduce readability because 1) many function/variables would be in their acronyms instead of fully-spelt names, and 2) most functions are broken into multiple lines. 

Could we use a compromise of 100 characters instead? Most of my code are complying within 100 characters, for example:
https://github.com/NCAR/VAPOR/blob/b07d642800077bff0e2e77b8a470c1d0b998af6e/lib/render/DVRenderer.cpp#L219

2. I'll go use our standard error handling. 

3. I can break drawVolumeFaces into smaller pieces.

4. I'll figure out a way to disabling it on macs.

shaomeng:
@clyne The required changes are made, and the branch is ready to merge!

clyne:
There are diagnostics in DVRender.cpp (search for cout).

drawVolumeFaces() is still a beast of a function weighing in at over 150 lines. How about putting the render pass setup code in separate functions?

shaomeng:
@clyne 
1. The diagnostics are now removed.
2. Do we need to worry about it now? The vast majority code in  `drawVolumeFaces()` passes in uniforms to the shader, and they would be reduced if the promise of a new shader manager holds true...

clyne:
Sent from my iPhone

On Aug 23, 2018, at 5:59 PM, Samuel Li <notifications@github.com> wrote:

@clyne <https://github.com/clyne>

   1. The diagnostics are now removed.
   2. Do we need to worry about it now? The vast majority code in
   drawVolumeFaces() passes in uniforms to the shader, and they would be
   reduced if the promise of a new shader manager holds true...


Yes please

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/NCAR/VAPOR/pull/969#issuecomment-415609585>, or mute
the thread

shaomeng:
@clyne Ok, the size of `_drawVolumeFaces()` is also addressed.

Issue #954, opened 2018-08-15
TFWidget performance issue


shaomeng:
Summary: 
- TFWidget takes too much time gathering information of a color map. 
Take a look at TFWidget in the DVR. 

Specifics:
- For a 256^3 data set, at each paint cycle, gathering the color map takes ~30% time, while the rest of the rendering routines take ~70% time. 
- For a 512^3 data, this ratio becomes ~20% and 80%.
- For a 64^3 data, this ratio becomes ~60% and 40%.
- Given the workload difference, gathering a color map shouldn't take that much time.

shaomeng:
John commented that it's the `makeLut()` step taking much time. As a result, John suggested to add a comparison method for a `MapperFunction`, so for the same `MapperFunction`, there's no need to `makeLut()` again.

shaomeng:
`makeLut()` doesn't seem to impose a big performance hit in practice, thus I'm closing it.

PR #975, opened 2018-08-27
Issue 954 mapper performance

sgpearse:
Fix for #954


lib/params/MapperFunction.cpp
```diff
@@ -133,6 +137,8 @@ MapperFunction::MapperFunction(
 
 	m_opacityMaps = new ParamsContainer(*(rhs.m_opacityMaps));
 	m_opacityMaps->SetParent(this);
+
+	_dirtyBit = true;
 }
 
 MapperFunction &MapperFunction::operator=( const MapperFunction& rhs ) {
```
shaomeng: What's the rational for this assign constructor to always construct _dirtyBit = true, instead of get the value from rhs?

lib/params/MapperFunction.cpp
```diff
@@ -300,25 +310,43 @@ void MapperFunction::hsvValue(float value, float *h, float *s, float *v) const
...
+  	  m_colorMap->color(v).toRGB(&clut[4*i]);
+  	  clut[4*i+3] = getOpacityValueData(v);
+  	}
+	memcpy(_clut, clut, sizeof(float)*_numEntries*4);
+  }
+  else {
+	memcpy(clut, _clut, sizeof(float)*_numEntries*4);
   }
+
+  _dirtyBit = false;
```
shaomeng: Should this `_dirtyBit = false;` assignment be in the bracket of line 329?

lib/params/OpacityMap.cpp
```diff
@@ -620,3 +632,11 @@ double OpacityMap::maxValue() const {
 	double relmax = GetValueDouble(_relMaxTag, 1.0);
 	return (minmaxmap[0] + relmax*(minmaxmap[1]-minmaxmap[0]));
 }
+
+bool OpacityMap::GetDirtyBit() const {
+	return _dirtyBit;
+}
+
+void OpacityMap::ResetDirtyBit() {
```
shaomeng: The name of Reset should be used when there's an apparent default value, so reset something clearly means it goes to the default value.
In this case, _dirtybit doesn't have an apparent default value, thus the name `ResetDirtyBit()` doesn't tell what this function does. 
Consider changing it to `SetDirtyBitFalse()` or taking in a value: `ResetDirtyBit( bool newValue )`.

clyne:
What are we trying to accomplish with this PR? It looks like a lot of complexity just to avoid calculation of a lookup table. How expensive is that operation?

shaomeng:
I profiled the current master branch and this feature branch. I'm measuring within a `_paintGL()` function call, how much percentage time is spent on querying this LUT. The test operation is DVR on a 512^3 data.

The first screenshot is the master branch result. 
screenshot from 2018-08-27 15-27-11


The 2nd screenshot is the feature branch result:
screenshot from 2018-08-27 15-27-03


This is some serious performance difference, and for this particular test on cisl-vapor, there's some noticeable interactivity improvement.

lib/params/MapperFunction.cpp
sgpearse: Fair point.

lib/params/MapperFunction.cpp
sgpearse: No.  Once we regenerate the lookup table we reset the dirty bit.

clyne:
OK. But how much time is it taking (in seconds, for example)? 

> On Aug 27, 2018, at 3:30 PM, Samuel Li <notifications@github.com> wrote:
> 
> I profiled the current master branch and this feature branch. I'm measuring within a _paintGL() function call, how much percentage time is spent on querying this LUT. The test operation is DVR on a 512^3 data.
> 
> The first screenshot is the master branch result.
> 
> 
> The 2nd screenshot is the feature branch result:
> 
> 
> This is some serious performance difference, and for this particular test on cisl-vapor, there's some noticeable interactivity improvement.
> 
> —
> 

John Clyne
clyne@ucar.edu

shaomeng:
- Cisl-vapor: mostly between 0.006 and 0.008. 
- Geyser: mostly between 0.012 and 0.014.

clyne:
Then I would say that it has no discernible impact on frame rate and we should leave it alone. The params classes maintain state in an Xml database and caching results internally is tricky and error prone.

> On Aug 27, 2018, at 11:54 PM, Samuel Li <notifications@github.com> wrote:
> 
> 	• Cisl-vapor: mostly between 0.006 and 0.008.
> 	• Geyser: mostly between 0.012 and 0.014.
> —
> 

John Clyne
clyne@ucar.edu

shaomeng:
The impact on frame rate, it's actually quite discernible...
But of course, we'll need to make sure the fix works correctly, and doesn't break other parts. 

With the very fast single-core CPU on cisl-vapor, here are frame rates with and without the optimization.

screenshot from 2018-08-28 11-59-56

screenshot from 2018-08-28 11-59-50

PR #1016, opened 2018-10-08
Issue 962 probe

sgpearse:
This is the implementation of our Slice Renderer (formerly the Probe).


apps/vaporgui/VariablesWidget.cpp
```diff
@@ -101,18 +107,30 @@ void VariablesWidget::Reinit(
...
+		if (dimFlags & THREED) {
 			_activeDim = THREEDIMS;
+            orientationFrame->hide();
+        }
 		else
 			_activeDim = TWODIMS;
+            //orientationFrame->show();
 	}
 
+    // If the renderer is only 3D, hide the 2D orientation selector
```
clyne: Remove  commented out code. Use ifdefs if the code needs to re retained.

lib/render/SliceRenderer.cpp
```diff
@@ -0,0 +1,374 @@
...
+    p->GetBox()->GetExtents(_cacheParams.boxMin, _cacheParams.boxMax);
+
+    MapperFunction* tf = p->GetMapperFunc(_cacheParams.varName);
+    tf->makeLut(_cacheParams.tf_lut);
+    _cacheParams.tf_minMax = tf->getMinMaxMapValue();
+    
+    if (_textureData) delete [] _textureData;
+    _textureData = new unsigned char[_textureWidth * _textureHeight * 4];
+
+    int rc = _saveTextureData();
```
clyne: If saveTextureData() generates an error condition that calls SetErrMsg() then saveCacheParams should pass the error upwards (it can't return void).

lib/render/SliceRenderer.cpp
```diff
@@ -0,0 +1,374 @@
...
+            if (bin >= _cacheParams.tf_lut.size())
+                bin = _cacheParams.tf_lut.size()-4;
+            
+            unsigned char red, green, blue, alpha;
+            red   = _cacheParams.tf_lut[bin + 0]*255;
+            green = _cacheParams.tf_lut[bin + 1]*255;
+            blue  = _cacheParams.tf_lut[bin + 2]*255;
+            alpha = _cacheParams.tf_lut[bin + 3]*255;
+
+            _textureData[index + 0] = red;
```
clyne: You should use an "indirect" texture for the slicer: store the data value in the texture and let the shader do a lookup into 
the color map. That way you don't have to rebuild the texture every time the color map changes. Also, the data can be stored directly as floats. See Sam's volume renderer for an example.

Alternatively, it might be acceptable to cache the sampled slice data and generate the texture from the cached data to avoid querying the grid each time the colormap changes. The Grid::GetValue method can be expensive.

lib/render/SliceRenderer.cpp
```diff
@@ -0,0 +1,374 @@
...
+        _renderXZ(min, max);
+    else if (orientation == YZ)
+        _renderYZ(min, max);
+
+    lgl->DisableTexture();
+
+    return 0;
+}
+
+void SliceRenderer::_renderXY(
```
clyne: Couldn't renderXY, renderXZ and renderYZ be replaced with a single render() function if you simply did

max[i] := min[i]

when i is the axis corresponding to the plane normal?

sgpearse:
@clyne Ready for review/merge.

clyne:
Couple of things:

1. the slice renderer doesn't play nice with other renderers. For example, the Image render doesn't work if the slice renderer is used. Load the MOM Data/ MOM4/00010101/00010101.ocean_month.NWPp2.Clim.nc data set. Enable the slice renderer with any variable. Then enable the slice renderer. The map should be visible. It's not. Disable the slice renderer. The map is still not visible. 

2. There are still performance issues. The SliceRenderer::_saveTextureData() is making calls to functions inside of doubly nested loop to compute sample coordinates. Though this may make the code a little cleaner, if the calculations are performed directly the results are ~40% faster.

clyne:
I'm adding @stasj to the reviewers list. Stas, please see me comments below and review SliceRenderer.cpp for OpenGL issues.

SliceRenderer.cpp comments: 

1. While it is certainly possible to create a separate texture to handling missing values this necessitates the use of multiple texture units, which introduces a good bit of complexity. Missing value support could be added to your data texture simply by changing the internal texture format use with glTexImage2D from GL_R32F to GL_RG32F, and then adding the missing values to the dataValues array. No need to mess with activating/deactivating texture units then.

2. There are several OpenGL handles that may be freed without being initialized. For example, the destructor unconditionally frees _VAO, _vertexVBO, etc. These should be initialized to 0 (or whatever value is appropriate ) in the constructor and then only freed in the destructor if they are non-zero. I.e. The destructor is only valid if  _initializeGL() is completes successfully.

3. Note that all of the renderers have a one-time OpenGL initialization phase (performed in _initializeGL()), and a "setup" phase performed every time _paintGL() is called. It is essential that proper cleanup is performed for  *both* of these through all code paths. The class destructor should free any resources allocated in _initializeGL() and any return path from paintGL should ensure that any state set by "setup" is restored. Presently there are multiple return paths from paintGL that do not restore state from setup (your _initializeState() method)

4. You are passing pointers to internal container data structures (e.g. taking the address of _vertexCoords[0] inside of SliceRenderer::_initVertexVBO()). I think this is a no-no. @stasj ?

5. you are calling glDeleteTextures  on handles that may have already been freed (e.g. inside of SliceRenderer::_createDataTexture). These handles should be initialized in the constructor, and a guard should be used whenever you free them. 

E.g. (if _dataValueTextureID != 0) {glDeleteTextures(1, &_dataValueTextureID);}

Zero is value OpenGL often uses for these purposes, but you'll need to check the doc.

clyne:
Comment (5) from https://github.com/NCAR/VAPOR/pull/1016#issuecomment-443290138 is only partially addressed (you're still potentially freeing resources multiple times in, e.g. SliceRenderer::_resetColormapCache()

I don't believe that you need to use multiple texture units at all because, although you are using multiple textures (colormap and data) they are different shape. I think you can simply use the default texture unit, 0, and never invoke glActiveTexture. It would be worth trying this to simplify the code slightly.

lib/render/SliceRenderer.cpp
```diff
@@ -0,0 +1,700 @@
...
+    if (_texCoordVBO != 0)
+        glDeleteBuffers(1, &_texCoordVBO);
+    
+    glGenBuffers(1, &_texCoordVBO);
+    glBindBuffer(GL_ARRAY_BUFFER, _texCoordVBO);
+    glVertexAttribPointer(1, 2, GL_FLOAT, GL_FALSE, 0, (void*)0);
+    glEnableVertexAttribArray(1);
+    glBufferData(
+        GL_ARRAY_BUFFER,
+        sizeof(float)*sizeof(_texCoords),
```
StasJ: This is incorrect

lib/render/SliceRenderer.cpp
```diff
@@ -0,0 +1,700 @@
...
+    );
+}
+
+void SliceRenderer::_initVertexVBO() {
+    if (_vertexVBO != 0) 
+        glDeleteBuffers(1, &_vertexVBO);
+    
+    glGenBuffers(1, &_vertexVBO);
+    glBindBuffer(GL_ARRAY_BUFFER, _vertexVBO);
+    glVertexAttribPointer(0, 3, GL_DOUBLE, GL_FALSE, 0, (void*)0);
```
StasJ: Why are you using GL_DOUBLE?

lib/render/SliceRenderer.cpp
```diff
@@ -0,0 +1,700 @@
...
+    glEnableVertexAttribArray(0);
+    glBufferData(
+        GL_ARRAY_BUFFER, 
+        6*3*sizeof(double), 
+        _vertexCoords.data(),
```
StasJ: I think the use of a std::vector here is excessive but its up to you if you want to change it

lib/render/SliceRenderer.cpp
sgpearse: Consistency, because all of our coordinate information returned by the DataMgr is of type double.

lib/render/SliceRenderer.cpp
```diff
@@ -0,0 +1,700 @@
...
+        texMaxX, texMinY,
+        texMaxX, texMaxY,
+        texMinX, texMaxY
+    };
+    
+    glBindBuffer(GL_ARRAY_BUFFER, _texCoordVBO);
+    glBufferSubData(
+        GL_ARRAY_BUFFER, 
+        0, 
+        sizeof(float)*sizeof(_texCoords),
```
StasJ: You previously declared _texCoordVBO as GL_STATIC_DRAW and you are not using sizeof correctly here

lib/render/SliceRenderer.cpp
```diff
@@ -0,0 +1,700 @@
...
+    glBindTexture(GL_TEXTURE_2D, _dataValueTextureID);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
+
+    glTexImage2D(
+        GL_TEXTURE_2D, 
+        0, 
+        GL_RG32F,
```
StasJ: A 64bit per pixel floating point texture is a little excessive here however this is not a big deal.

lib/render/SliceRenderer.cpp
```diff
@@ -0,0 +1,700 @@
...
+   
+    glBindVertexArray(0);
+    glBindBuffer(GL_ARRAY_BUFFER, 0);
+    
+    ShaderProgram* s = _glManager->shaderManager->GetShader("Slice");
+    s->UnBind();
+    
+    glDisable(GL_BLEND);
+    glDisable(GL_DEPTH_TEST);
+    glDepthMask(GL_FALSE);
```
StasJ: While we don't have an official state (I think we should add one?) I think for the renderers the depth buffer should be enabled by default.

lib/render/SliceRenderer.cpp
```diff
@@ -0,0 +1,700 @@
...
+        _setXZVertexPositions(min, max);
+    else if (orientation == YZ)
+        _setYZVertexPositions(min, max);
+
+    glBindBuffer(GL_ARRAY_BUFFER, _vertexVBO);
+    glBufferSubData(
+        GL_ARRAY_BUFFER, 
+        0, 
+        6*3*sizeof(double), 
+        _vertexCoords.data()
```
StasJ: This was declared as static

lib/render/SliceRenderer.cpp
```diff
@@ -0,0 +1,700 @@
...
+    glBindTexture(GL_TEXTURE_2D, 0);
+
+    glActiveTexture(GL_TEXTURE0);
+    glBindTexture(GL_TEXTURE_1D, 0);
+   
+    glBindVertexArray(0);
+    glBindBuffer(GL_ARRAY_BUFFER, 0);
+    
+    ShaderProgram* s = _glManager->shaderManager->GetShader("Slice");
+    s->UnBind();
```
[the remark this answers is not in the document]
sgpearse: Gotcha.  It looks like this is how the ContourRenderer is unbinding.  Is it in a different situation?

lib/render/SliceRenderer.cpp
sgpearse: We were talking about this earlier, and and at the time we decided that I should turn off everything that gets turned on during the rendering.  I do think we could sit down and talk about this and flesh out a consistent protocol though.

sgpearse:
@StasJ Hey can you review the changes please?  Thanks!

PR #1017, opened 2018-10-11
TFWidget support for IsoSurfaces

sgpearse:
The TFWidget now supports movable isolines within the MappingFrame, and has modes that support isosurfaces.


apps/vaporgui/ContourSubtabs.cpp
```diff
@@ -3,9 +3,9 @@
 ContourAppearanceSubtab::ContourAppearanceSubtab(QWidget* parent) {
 	setupUi(this);
 
-	_TFWidget->Reinit((TFFlags)(CONSTANT));
-	_TFWidget->mappingFrame->setIsolineSliders(true);
-	_TFWidget->mappingFrame->setOpacityMapping(false);
+	_TFWidget->Reinit((TFFlags)(CONSTANT_COLOR));
+	_TFWidget->_mappingFrame->setIsolineSliders(true);
+	_TFWidget->_mappingFrame->setOpacityMapping(false);
```
sgpearse: This should be handled with flags, within the TFWidget

apps/vaporgui/IsoSurfaceSubtabs.cpp
```diff
@@ -4,7 +4,10 @@
 IsoSurfaceAppearanceSubtab::IsoSurfaceAppearanceSubtab(QWidget* parent) 
 {
     setupUi(this);
-    _TFWidget->Reinit((TFFlags)(CONSTANT));
+    _TFWidget->Reinit((TFFlags)(CONSTANT_COLOR | COLORMAP_VAR_IS_IN_TF2));
```
sgpearse: This should be handled with flags, within the TFWidget

clyne:
Hmm. This approach violates the zero-one-infinity “rule” (and probably some other best practices) and introduces lots of conditionals into the TFWidget. What happens if we need 3, and then 4 transfer function editors? We have to rewrite the TFWidget each time. We should figure how to generalize the SubTabs (if that is what it takes) so that they can support a container of TFWidget classes, not just a single instance.

jc

> On Oct 22, 2018, at 10:22 AM, sgpearse <notifications@github.com> wrote:
> 
> Having a renderer configure two TFWidgets is a feasible approach. On one hand, it would make the TFWidget more lightweight.
> 
> On the other hand, it increases the coupling between TFWidgets and the SubTabs that they reside within. The SubTabs for Isosurfaces and Contours would need to remember which of their TFWidgets contain colormaps vs isolines. This distinguishment would be needed during Update, Reinit, and any method that make changes to RenderParams.
> 
> I think that a decent amount of information can be hidden if a SubTab implements one TFWidget. My thinking was to make distinguishments under the hood, and return the isoline values or colormaps when requested, through a single object.
> 
> —
> 

John Clyne
clyne@ucar.edu

PR #1043, opened 2018-11-30
Dvr2

shaomeng:
This PR updates the RayCaster class so that 1) it uses glm to manipulate matrices and vectors, and 2) it uses the newly created infrastructure that Stas has put together. This PR enables fixed step DVR and IsoSurface renderer on MacOS platforms.


lib/render/RayCaster.cpp
```diff
@@ -98,6 +112,16 @@ RayCaster::~RayCaster()
...
+        _xyCoordsTextureId = 0;
+    }
+    if( _zCoordsTextureId )
+    {
+        glDeleteTextures( 1, &_zCoordsTextureId );
+        _zCoordsTextureId = 0;
+    }
 
     // delete buffers
     if( _frameBufferId )
```
StasJ: You call glDeleteBuffers on a FrameBuffer object which causes a memory error in OpenGL.

You need to use glDeleteFramebuffers.

shaomeng: Fixed, thansk!

shaomeng:
This PR should NOT be merged. I'm closing it now.

PR #1052, opened 2018-12-05
DVR2 current

sgpearse:
Master, DVR2, and the Isolines (aka TFWidget for Isosurfaces) are now in agreement with each other.


include/vapor/RayCaster.h
```diff
@@ -54,17 +58,19 @@ class RENDER_API RayCaster : public Renderer
         //              | /
         //              |/
         //            0 --------X
-        float *frontFace, *backFace;     // user coordinates, size == bx * by * 3
```
StasJ: Variable names do not follow our coding convention

lib/render/RayCaster.cpp
```diff
@@ -232,27 +267,31 @@ RayCaster::UserCoordinates::GetCurrentGrid( const RayCasterParams* params,
...
+    {
+        *gridpp = grid;
+        return 0;
     }
-    return grid;
 }
 
-bool RayCaster::UserCoordinates::IsUpToDate( const RayCasterParams* params,  
-                                                   DataMgr*         dataMgr ) const
+int  RayCaster::UserCoordinates::IsMetadataUpToDate( const RayCasterParams* params,
```
StasJ: Since this function has multiple return codes, I would recommend using an enum instead of arbitrary integers

lib/render/RayCaster.cpp
```diff
@@ -232,27 +267,31 @@ RayCaster::UserCoordinates::GetCurrentGrid( const RayCasterParams* params,
...
     {
-        return false;
+        return 1;
     }
 
     // compare grid boundaries and dimensions
-    StructuredGrid* grid = this->GetCurrentGrid( params, dataMgr );
-    if( !grid )
+    StructuredGrid* grid  = nullptr;
+    if( this->GetCurrentGrid( params, dataMgr, &grid ) != 0 )
```
StasJ: I am not sure how expensive getting the grid is however you should be able to get the extends from the params directly without requesting a new grid.

lib/render/RayCaster.cpp
```diff
@@ -430,41 +480,88 @@ bool RayCaster::UserCoordinates::UpdateCoordinates( const RayCasterParams* param
...
+    }
+
+    return 0;   // Success
+}
+
 int RayCaster::_paintGL( bool fast ) 
 {
-#ifdef Darwin
-    return 0;
+#ifdef DEBUG
```
StasJ: I think you mean NDEBUG. DEBUG is not a standard macro. Also, this is automatically disabled on release builds.

lib/render/RayCaster.cpp
```diff
@@ -541,94 +687,111 @@ int RayCaster::_paintGL( bool fast )
...
-    _drawVolumeFaces( 2, insideACell );    // 2nd pass, render front facing polygons
+    // 2nd pass, render front facing polygons
+    _drawVolumeFaces( 2, castingMode, insideACell );    
 
     glBindFramebuffer( GL_FRAMEBUFFER, 0 );
     glViewport( 0, 0, _currentViewport[2], _currentViewport[3] );
 
-    _drawVolumeFaces( 3, insideACell, ModelView, InversedMV, fast );  // 3rd pass, perform ray casting
+    // 3rd pass, perform ray casting
+    if(  castingMode == 1 )
```
StasJ: Could you make castingMode an enum or at least #define some constants? It is compared in multiple places with unlabeled integers.

lib/render/RayCaster.cpp
```diff
@@ -749,12 +908,10 @@ void RayCaster::_drawVolumeFaces( int            whichPass,
     else if( whichPass == 2 )
     {
         glUseProgram( _2ndPassShaderId );
-        uniformLocation = glGetUniformLocation( _2ndPassShaderId, "MVP" );
-        glUniformMatrix4fv( uniformLocation, 1, GL_FALSE, MVP );
-        uniformLocation = glGetUniformLocation( _2ndPassShaderId, "boxMin" );
-        glUniform3fv( uniformLocation, 1, _userCoordinates.boxMin );
-        uniformLocation = glGetUniformLocation( _2ndPassShaderId, "boxMax" );
-        glUniform3fv( uniformLocation, 1, _userCoordinates.boxMax );
+        GLint uniformLocation = glGetUniformLocation( _1stPassShaderId, "MV" );
```
StasJ: You are using the first pass shader id to configure the second pass shader

lib/render/RayCaster.cpp
```diff
@@ -940,15 +1181,44 @@ void RayCaster::_renderTriangleStrips() const
...
+                attrib1Buffer[ attribIdx + 2 ] = 0;
+                attrib1Buffer[ attribIdx + 3 ] = 1;
+            }
+            glBufferData( GL_ARRAY_BUFFER,      bx * by * 4 * sizeof(int),
+                          attrib1Buffer,        GL_STREAM_READ );
+            glVertexAttribIPointer( 1, 4,       GL_INT, 0, (void*)0 );
+        }
         glBufferData( GL_ELEMENT_ARRAY_BUFFER,  numOfVertices * sizeof(unsigned int),
-                      indexBuffer,              GL_STREAM_DRAW );
+                      indexBuffer,              GL_STREAM_READ );
```
StasJ: GL_STREAM_READ specifies you are writing data to the buffer from the GPU and reading it back to the CPU. You want to use GL_STREAM_DRAW

lib/render/RayCaster.cpp
```diff
@@ -996,15 +1324,44 @@ void RayCaster::_renderTriangleStrips() const
...
         glDrawElements( GL_TRIANGLE_STRIP,      numOfVertices,
                         GL_UNSIGNED_INT,        (void*)0 );
     }
 
+    //
     // Render left face
+    //
```
StasJ: You are supposed to have a VertexArray for each of these faces vs reconfiguring one VAO for each draw.

include/vapor/IsoSurfaceRenderer.h
```diff
@@ -22,8 +22,8 @@ class RENDER_API IsoSurfaceRenderer : public RayCaster
     }
 
 protected:
-    void _loadShaders();
-    void _3rdPassSpecialHandling( bool );
+    int  _loadShaders();
+    void _3rdPassSpecialHandling( bool, long );
```
StasJ: Please give the arguments names

lib/render/RayCaster.cpp
```diff
@@ -42,23 +45,36 @@ RayCaster::RayCaster( const ParamsMgr*    pm,
...
+    _zCoordsBufferId             = 0;
 
     _1stPassShaderId             = 0;
     _2ndPassShaderId             = 0;
     _3rdPassShaderId             = 0;
+    _3rdPassMode1ShaderId        = 0;
+    _3rdPassMode2ShaderId        = 0;
 
     _drawBuffers[0]              = GL_COLOR_ATTACHMENT0;
     _drawBuffers[1]              = GL_COLOR_ATTACHMENT1;
```
clyne: OpenGL call should not be made in the constructor:

    GLint viewport[4];
     glGetIntegerv( GL_VIEWPORT, viewport );
     std::memcpy( _currentViewport, viewport, 4 * sizeof(GLint) );

There is no OpenGL context until after RayCaster::_initializeGL is called

lib/render/RayCaster.cpp
```diff
@@ -232,27 +267,31 @@ RayCaster::UserCoordinates::GetCurrentGrid( const RayCasterParams* params,
     {
         MyBase::SetErrMsg("UserCoordinates::GetCurrentGrid() isn't on a StructuredGrid; "
                           "the behavior is undefined in this case.");
+        return 1;
```
clyne: Per our coding conventions a function returning an integer should return a *negative* number on failure, and 0 on success. We follow the posix system call conventions in this regard

lib/render/RayCaster.cpp
```diff
@@ -385,29 +430,34 @@ bool RayCaster::UserCoordinates::UpdateCoordinates( const RayCasterParams* param
     // Save the data field values and missing values
     size_t numOfVertices = dims[0] * dims[1] * dims[2];
     if( dataField )
+    {
         delete[] dataField;
+        dataField = nullptr;
+    }
     dataField = new float[ numOfVertices ];
-    if( !dataField )
+    if( !dataField )    // Test if allocation successful for 3D buffers.
```
clyne: The new[] operator will thrown an exception unless you overload it. I.e. if new[] fails this code will terminate.

lib/render/RayCaster.cpp
```diff
@@ -264,16 +303,16 @@ bool RayCaster::UserCoordinates::IsUpToDate( const RayCasterParams* params,
...
     }
 
     // now we know it's up to date!
     delete grid;
-    return true;
+    return 0;
 }
         
-bool RayCaster::UserCoordinates::UpdateCoordinates( const RayCasterParams* params,
+int  RayCaster::UserCoordinates::UpdateFaceAndData( const RayCasterParams* params,
```
clyne: This is an awfully long function with a lot of repetitive operations. How about creating a helper method parameterized by the face to perform the loading of the face data?

lib/render/RayCaster.cpp
```diff
@@ -430,41 +480,88 @@ bool RayCaster::UserCoordinates::UpdateCoordinates( const RayCasterParams* param
...
+    if( _initializeFramebufferTextures() != 0 )
+    {
+        MyBase::SetErrMsg("Failed to Create Framebuffer and Textures!");
+        return -1;
+    }
+
+    return 0;   // Success
+}
+
 int RayCaster::_paintGL( bool fast )
```
clyne: The paingGL function is way to long. Furthermore, it doesn't restore opengl state on an error condition. Please break it up into smaller, well-defined pieces.

lib/render/RayCaster.cpp
```diff
@@ -765,156 +922,210 @@ void RayCaster::_drawVolumeFaces( int            whichPass,
...
     
-void RayCaster::_3rdPassSpecialHandling( bool fast )
+void RayCaster::_3rdPassSpecialHandling( bool fast, long castingMode )
 {
     // Left empty intentially.
     // Derived classes feel free to put stuff here.
 }
     
-void RayCaster::_renderTriangleStrips() const
+void RayCaster::_renderTriangleStrips( int whichPass, long castingMode ) const
```
clyne: This method is too long with lots of repetition. Please use helper functions and reduce the repetitive code.

share/shaders/DVR3rdPassMode1.frag
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    return false;
+}
+
+//
+// Input:  Location to be evaluated in texture coordinates
+// Output: Gradient at that location
+//
+vec3 CalculateGradient( const in vec3 tc )
```
clyne: CalculateGradient is used by Isosurface and DVR. Please put in a separate file: don't cut-n-paste code with this many lines.

include/vapor/RayCaster.h
shaomeng: which one specifically?

StasJ: All of them. Private variable names start with and underscore.

lib/render/RayCaster.cpp
shaomeng: I don't want to change the type of returned value from int to enum, so I instead used macros, see commit 03bed17f. Does this address your concern?

lib/render/RayCaster.cpp
shaomeng: The params keeps boundaries in ASCII, which is most likely a slightly different value from the intended floating point value. Thus, comparing against values returned by params will most likely result in `false`, even though they're supposed to be `true`. 
Note this only affects floating point values return by params.
@clyne  is this correct?

lib/render/RayCaster.cpp
shaomeng: I intended to disable that code on release build, and only enable it on debug build. Does it make sense?

lib/render/RayCaster.cpp
shaomeng: With only 2 casting modes, I feel either enum or macros are over killing. Thus I reject this change request.

lib/render/RayCaster.cpp
shaomeng: addressed by commit 03bed17f.

lib/render/RayCaster.cpp
shaomeng: What are the tradeoffs between these two approaches?

include/vapor/IsoSurfaceRenderer.h
shaomeng: The argument name is given in the .cpp file. Does this address your concern?

lib/render/RayCaster.cpp
shaomeng: addressed in commit 14a25750

lib/render/RayCaster.cpp
shaomeng: This function is long but it's also got clear structures, so a reader can get understand it very easily. 
Plus, within this function, though each block seems repetitive, they aren't, because the face indices are different and they have to be hard coded before we utilize some real fancy parameterization. 
What do you think?

share/shaders/DVR3rdPassMode1.frag
shaomeng: GLSL only supports `#include` in its extension profile, but not core profile. What to do in this case?

lib/render/RayCaster.cpp
clyne: Not correct, per our discussion. If the value cached in RayCaster.cpp is set from a value obtained from the Params subsequent comparisons will only be unequal if the value in the Params is explicitly changed.

StasJ: While it is true the mapping between floating point and real numbers is not one to one, the code that converts between the two is deterministic

lib/render/RayCaster.cpp
StasJ: Yes. The macro to do that is NDEBUG. Also, the ShaderManager class already implements that functionality.

lib/render/RayCaster.cpp
StasJ: The approach I mentioned is the standard practice for OpenGL 3

include/vapor/IsoSurfaceRenderer.h
StasJ: No because you reference the header file for function definitions.

lib/render/RayCaster.cpp
shaomeng: This method seems to contain repetitions, but each face has different indices, and very hard to parameterize and deploy helper functions. Further, this method contains many OpenGL function calls that have very tricky arguments. Currently this function is fairly optimized and works well, I'd suggest we leave it alone at this point. What do you think?

lib/render/RayCaster.cpp
clyne: You could replace the 6 nested loop pairs with 3 calls to functions that looked like the snippet below. This would reduce the repetitive code, and make UpdateFaceAndData shorter and easier to understand.

`float *updateLeftRight(
    const vector <size_t> &dims,
    bool zMinFlag
) const {

    float buf[3];
    size_t z = zMinFlag ? 0 : dims[2] - 1;

    float *face = new float[ dims[0] * dims[1] * 3 ];
    size_t idx = 0;
    for( size_t y = 0; y < dims[1]; y++ ) {
        for( size_t x = 0; x < dims[0]; x++ )
        {
            grid->GetUserCoordinates( x, y, z, buf[0], buf[1], buf[2] );
            frontFace[ idx++  ] = (float)buf[0];
            frontFace[ idx++  ] = (float)buf[1];
            frontFace[ idx++  ] = (float)buf[2];
        }
    }
    return(face);
}

float *updateFrontBack(
    const vector <size_t> &dims,
    bool zMinFlag
) const {
...
}

float *updateBottomTop(
    const vector <size_t> &dims,
    bool zMinFlag
) const {
...
}`

share/shaders/DVR3rdPassMode1.frag
clyne: Ask @StasJ

lib/render/RayCaster.cpp
StasJ: I have to disagree. I found this function very confusing and difficult to understand.

include/vapor/RayCaster.h
shaomeng: This comment is addressed by commit 1fb9c9fd.

lib/render/RayCaster.cpp
StasJ: I disagree. I don't think I should have to go to the header file of a different class to find out what rendering mode 1 vs 2 is.

lib/render/RayCaster.cpp
clyne: You can easily consolidate the 6 pairs of face rendering nested loops by taking advantage of the fact that very little changes between each pair (e.g. left/right, top/bottom, front/back). For example, the front and back triangle rendering loops differ by constant values. In the first loop you have:

`unsigned int   attribIdx       = ((y + 1) * bx + x) * 4;
attrib1Buffer[ attribIdx + 2 ] = int(bz) - 2;
attrib1Buffer[ attribIdx + 3 ] = 0;`

and in the second loop you have:

`unsigned int   attribIdx       = (y * bx + x) * 4;
attrib1Buffer[ attribIdx + 2 ] = 0;
attrib1Buffer[ attribIdx + 3 ] = 1;`

You could have a single function for both of these two loops parameterized by a front/back flag:

`if (frontFlag) {
yinc = 1
att2 = int(bz) - 2;
att3 = 0
} else {
yinc = 0
att2 = 0
att3 = 1
}`

Then your inner loop would look like:

`attribIdx       = ((y + yinc) * bx + x) * 4;
attrib1Buffer[ attribIdx + 2 ] = att2
attrib1Buffer[ attribIdx + 3 ] = att3
`

This would make _renderTriangleStrips() much shorter, and make the face rendering code more maintainable.

lib/render/RayCaster.cpp
shaomeng: This is addressed in 41b11793.

lib/render/RayCaster.cpp
shaomeng: this is addressed in 49d19d28.

lib/render/RayCaster.cpp
clyne: Needs to follow shader initialization:

`    SmartShaderProgram shader = _glManager->shaderManager->GetShader("DVRsomething");
    if (!shader.IsValid())
        return -1;`

...
followed by shader unbind at any return point from paintGL

lib/render/RayCaster.cpp
clyne: Please make an enum or a boolean flag. Something with a name indicating what casting mode means

lib/render/RayCaster.cpp
clyne: @StasJ will refactor to set uniforms using SHaderProgram class API

include/vapor/IsoSurfaceRenderer.h
clyne: Agreed

lib/render/RayCaster.cpp
clyne: suggested changes to _renderTriangleStrips()

1. Better document rendering loops
2. Clearer variable names
3. Move opengl setup code at beginning of each loop into a single function, parameterized as needed.

lib/render/RayCaster.cpp
clyne: Use asserts for progammatic errors (checking for min and max extent vector size())

lib/render/RayCaster.cpp
clyne: @stasj will address

lib/render/RayCaster.cpp
clyne: Make it from 6 loops to 3 functions

lib/render/RayCaster.cpp
shaomeng: Fixed in 4d41f667

include/vapor/IsoSurfaceRenderer.h
shaomeng: fixed in 94798d0f

lib/render/RayCaster.cpp
shaomeng: addressed in 283f503a

lib/render/RayCaster.cpp
shaomeng: Fixed in ddf07345

Issue #1074, opened 2018-12-12
Changing volume renderer extents squishes domain vs clipping


StasJ:
Add a new volume renderer
Change the extents
Dataset appears to be squished to the extents instead of clipped to them

PR #1075, opened 2018-12-12
resolve issue 1074

shaomeng:
This PR resolves issue #1074


shaomeng:
This PR resolves the issue reported in #1074 , while doesn't incur extra computation to query a new grid. I.e., a grid is queried exactly once during a paintGL() event. This is achieved by storing a queried grid and use it next time without querying.

PR #1091, opened 2018-12-14
DVR sets depth correctly

shaomeng:
This PR modifies DVR to have depth value correctly set, thus DVR plays well with other renderers. 
A demo is located at:
https://drive.google.com/open?id=15qfbzlKcneCA2a3OSdc_9qXFQ6H7MGDj&authuser=shaomeng@ucar.edu


clyne:
Why isn't the annotation (white bounding box) visible when it's behind the volume rendering?

shaomeng:
The white bounding box is invisible because it's drawn *after* every renderer, and it's got depth behind the DVR. Let me try if it's resolved by the rendering order PR this morning.

shaomeng:
After syncing with the master branch, the white bounding box is still not shown... Is it rendered before the DVR?

PR #1095, opened 2018-12-17
Issue 1085 tf widget performance

sgpearse:
Incorporated a stride into our histogram generation, as well as calls to DataMgr::GetVariableRange().  Also now changing the histogram color to grey when it's out of date.


clyne:
@shaomeng I don't know if you received a notification, but this PR needs your review.

thanks - jc

shaomeng:
Sorry, I saw myself became an "assignee" and wasn't sure what that is...
I'll review it soon!

sgpearse:
> Which part of the code calculates the histogram with full accuracy? Or we lost this capability permanently?

Correct, we are no longer using full accuracy.

shaomeng:
@sgpearse If we don't support full accuracy histograms, and always calculate it with a hard-coded stride number, then why do we create a params entry for stride?

shaomeng:
Think about it twice, should we have a more intelligent logic to choose the stride value, instead of hard-code a sixteen? This logic could be pretty simple, like when the total grid point number is greater than a threshold, we set `stride = 16`, otherwise no stride. We could even use `stride=32` for very large grids. This simple logic could make our approach much more applicable in various cases. What do you think @clyne ?

include/vapor/MapperFunction.h
```diff
@@ -136,6 +136,17 @@ class PARAMS_API MapperFunction : public ParamsBase
 	return(getMinMaxMapValue()[1]);
  };
 
+ int getHistogramStride() {
+    return GetValueDouble(_histogramStrideTag, 1.f);
+ }
```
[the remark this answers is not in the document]
sgpearse: Fair enough.

sgpearse:
@clyne and @shaomeng,

Sam, your change requests are in.  Note that there's a workaround in the RenderParams, which has duplicate code for calculating the stride (copied from TFWidget).  This is going to be fixed during a refactorization of RenderParams (and maybe ParamsBase?) so that their responsibilities are lessened, but we're just a band-aid for now.

apps/vaporgui/MappingFrame.cpp
```diff
@@ -316,9 +327,9 @@ void MappingFrame::populateSamplingHistogram() {
     coords[Y] = minExts[Y];
     coords[Z] = minExts[Z];
 
-    int iSamples = SAMPLE_RATE;
-    int jSamples = SAMPLE_RATE;
-    int kSamples = SAMPLE_RATE;
```
clyne: This code doesn't look correct. It appears to be doubling up on the stride, and not advancing the coordinates correctly.

More importantly, what is the purpose of populateSamplingHistogram()?

clyne: But why is populateSamplingHistogram  needed? How is it different from populateIteratingHistogram()? I.e why do we need both?

sgpearse: @clyne 

It's debatable whether we need it at this point.  Previously, we would compute our histograms based on the actual data volume within a renderer's given region.  This required sampling the grid for the SliceRenderer.  Using an iterator was not representative of the actual data volume of the slice.

Now that we're fudging the histogram with _stride, maybe it's reasonable to fudge it further by just gathering an iterator from a hyperslab that's adjacent to the current slice.  However this will only work if slices are axis aligned.  Leigh indicated that rotating slices would be an important thing to have for the slicer.

apps/vaporgui/TFWidget.cpp
```diff
@@ -268,6 +288,23 @@ void TFWidget::getVariableRange(
 	values[1] = tf->getMaxMapValue();
 }
 
+void TFWidget::calculateStride( string varName ) {
+    std::vector<size_t> dimsAtLevel;
+	int ref = _rParams->GetRefinementLevel();
+    int rc = _dataMgr->GetDimLensAtLevel(varName, ref, dimsAtLevel);
+    assert(rc>=0);
+
+    int size = 1;
```
shaomeng: An `int` can represent the biggest value around 1000x1000x2000 . What will this code do when the grid is larger than that?

lib/params/RenderParams.cpp
```diff
@@ -338,6 +342,21 @@ void RenderParams::SetColorbarPbase(ColorbarPbase* pb){
 	
 }
 
+void RenderParams::_calculateStride( string varName ) {
```
shaomeng: This `_calculateStride()` function seems to be duplicate to the one from `TFWidget.cpp`. Any specific reason to duplicate it in two classes?

shaomeng:
@sgpearse Scott, could you double check the delta calculation in `MappingFrame.cpp`?
The following lines, are they correct? 
``` C++
    double dx = (maxExts[X]-minExts[X])/(1+SAMPLE_RATE);
    double dy = (maxExts[Y]-minExts[Y])/(1+SAMPLE_RATE);
    double dz = (maxExts[Z]-minExts[Z])/(1+SAMPLE_RATE);
```

lib/params/RenderParams.cpp
sgpearse: Hey Sam, there is a reason.  From my comment in this thread on December 19:

> Sam, your change requests are in. Note that there's a workaround in the RenderParams, which has duplicate code for calculating the stride (copied from TFWidget). This is going to be fixed during a refactorization of RenderParams (and maybe ParamsBase?) so that their responsibilities are lessened, but we're just a band-aid for now.

sgpearse:
Yes, that is correct.

lib/params/RenderParams.cpp
shaomeng: Scott, I understand you have a work plan to resolve this duplication later on. But I don't understand *why* does it have to be like this at the first place.

shaomeng:
My understanding is that for stride values, when they're small (likely 1, 2, 3, 4, etc), it's a lot slower to have a statement `itr += stride` than `++itr`. It's only when the stride is large enough, the former use has performance advantage. Is it true? If so, have we taken this into consideration? @sgpearse

shaomeng:
Your delta calculation, `double dx = (maxExts[X]-minExts[X])/(1+SAMPLE_RATE);` still confuses me. I'll need some clarification from @sgpearse later today.

clyne:
> @sgpearse Scott, could you double check the delta calculation in `MappingFrame.cpp`?
> The following lines, are they correct?
> 
> ```c++
>     double dx = (maxExts[X]-minExts[X])/(1+SAMPLE_RATE);
>     double dy = (maxExts[Y]-minExts[Y])/(1+SAMPLE_RATE);
>     double dz = (maxExts[Z]-minExts[Z])/(1+SAMPLE_RATE);
> ```

Doesn't look correct to me either. If SAMPLE_RATE is the total number of samples you should divide by SAMPLE_RATE-1

sgpearse:
> Line 341 of MappingFrame.cpp is the following:
> for (int k=0; k<=kSamples; k++) {
> This means when you have a value in variable kSamples, you actually sample one more value along that axis. Not a good coding practice. Can you either change the name of the variable, or change the loop termination condition so that the variable name does reflect the number of samples?

Ok, I've changed the variable names to more accurately reflect count and incrementation.

> My understanding is that for stride values, when they're small (likely 1, 2, 3, 4, etc), it's a lot slower to have a statement itr += stride than ++itr. It's only when the stride is large enough, the former use has performance advantage. Is it true? If so, have we taken this into consideration? We will likely need to run some timing numbers to decide what stride number is big enough to give us a performance gain.

I think you're most likely correct, and that this would probably improve our performance.  I'm apprehensive on implementing it right now for a few reasons.  

1) There's a pretty high priority bug that I need to address.

2) Our performance improvements over the past few weeks seem to have given Vapor a satisfactory response time.  Without more evidence, I think there's more of a risk to introduce regressions with more modifications to the stride calculation than there are rewards.

3) I know that the iterator stride is used in other places in the code base, not just here.

I think we should make a feature request of this and revisit.

clyne:
For some reason won't let me provide an in-line response to Sam's comment: 

"My understanding is that for stride values, when they're small (likely 1, 2, 3, 4, etc), it's a lot slower to have a statement itr += stride than ++itr. It's only when the stride is large enough, the former use has performance advantage. Is it true? If so, have we taken this into consideration? We will likely need to run some timing numbers to decide what stride number is big enough to give us a performance gain."

@shaomeng is correct. For a stride of 1 the ++ operator is faster than +=. For larger strides the += is faster. Mileage will varying on where the break even point is. I don't think we should do an analysis now. Let's move on.

shaomeng:
@clyne @sgpearse  Mileage will varying on where the break even point is, but it's definitely not 1. In other words, we choose any reasonable value greater than one, could be 3, 4, 5, it'll be better than what is currently there right now, before we run a precise analysis. 

Regarding the time to revisit it, I think right now is the best time, since this feature isn't in the master yet. It'll be much more difficult to come back and address it once it got merged, both psychologically and according to how we used to deal with similar situations. I would agree that a precise analysis is optional, but choosing a number and implement the logic is necessary.

sgpearse:
> Doesn't look correct to me either. If SAMPLE_RATE is the total number of samples you should divide by SAMPLE_RATE-1

The division actually needs to just be by SAMPLE_RATE (not SAMPLE_RATE-1 as sam has emphasized).  We are nudging the first sample by delta/2 so that we are slightly within the bounds of the domain.  This avoids retrieving missing values at the edge.

clyne:
Then you should probably add some comments to that effect because as it stands the code looks wrong.

> On Dec 21, 2018, at 11:59 AM, sgpearse <notifications@github.com> wrote:
> 
> Doesn't look correct to me either. If SAMPLE_RATE is the total number of samples you should divide by SAMPLE_RATE-1
> 
> The division actually needs to just be by SAMPLE_RATE (not SAMPLE_RATE-1 as sam has emphasized). We are nudging the first sample by delta/2 so that we are slightly within the bounds of the domain. This avoids retrieving missing values at the edge.
> 
> —
> 

John Clyne
clyne@ucar.edu

clyne:
Sam, I would like to get this rather large branch merged before Scott and I both disappear for the holiday. The code is correct. There are only minor performance considerations. They are minor, because the way the stride is calculated unit stride will only result for very small data sets, in which case the performance penalty will be negligible. 


> On Dec 21, 2018, at 11:53 AM, Samuel Li <notifications@github.com> wrote:
> 
> @clyne @sgpearse Mileage will varying on where the break even point is, but it's definitely not 1. In other words, we choose any reasonable value greater than one, could be 3, 4, 5, it'll be better than what is currently there right now, before we run a precise analysis.
> 
> Regarding the time to revisit it, I think right now is the best time, since this feature isn't in the master yet. It'll be much more difficult to come back and address it once it got merged, both psychologically and according to how we used to deal with similar situations. I would agree that a precise analysis is optional, but choosing a number and implement the logic is necessary.
> 
> —
> 

John Clyne
clyne@ucar.edu

shaomeng:
@clyne The current code will result in a stride of 1 with 125^3 cubes, and stride of 2 with 158^3 cubes, and 3 with 181^3 cubes. They sure look small, but I remember we used to spend seconds to calculate the histogram with similar sizes. **If**  the break even point is bigger than 3, then we're degrading performance at 181^3 scale.

clyne:
@shaomeng By my calculations a non-unit strides results for grids over 1,000,000 grid points (100^3), which is what Scott and I agreed upon. This is ~32 times smaller than Leigh’s data (371x401x201) where I did my benchmarking before doing *any* optimizations and found the histogram calculation to take ~2.5 seconds. Moreover, I’ve already benchmarked this branch and found the performance to be reasonable. Let’s close this issue and move on.

Scott, if there are no further code correctness issues I’m prepared to accept this PR.

thanks -jc


> On Dec 21, 2018, at 12:18 PM, Samuel Li <notifications@github.com> wrote:
> 
> @clyne The current code will result in a stride of 1 with 125^3 cubes, and stride of 2 with 158^3 cubes, and 3 with 181^3 cubes. They sure look small, but I remember we used to spend seconds to calculate the histogram with similar sizes. If the break even point is bigger than 3, then we're degrading performance at 181^3 scale.
> 
> —
> 

John Clyne
clyne@ucar.edu

shaomeng:
@clyne I'm OK with moving on, but a quick correction, a grid size of 1,999,999 (~2Million) will result in a stride size one. 
(Line 329-330 in TFWidget.cpp)

PR #1106, opened 2018-12-19
Fix session related bugs and comply to session flow chart

StasJ:


clyne:
The modeling transforms don't appeared to be restored from a session file. If you load a data set, enable the 2D renderer, rotate the scene, and then save the session when the resulting session is reloaded everything is restored except the scene rotation. Note I'm exiting the application before loading the session file generated in the first step.

clyne:
@stasj Just making sure you get email re my comment.

StasJ:
I think I know why this regression was introduced. It shouldn't take long to fix.

Also in the right sidebar there is a list of participants. GitHub emails notifications to everyone on that list.

clyne:
Viewpoints work now. However, one other minor problem: if you load a session from the command line the renderer subtabs are not correct for whatever renderers are enabled. Disabling and re-enabling them fixes the problem. For example, load Katrina, enable a 2D renderer, and save the state. Then load the session file from the command line. The variables subtab looks like the attached. 
variablesbad

Issue #1123, opened 2018-12-20
Transfer function editor improvements


clyne:
Improvements to Transfer Function editor suggested by @leighorf:

1. Allow appearance tab to be detached from the main GUI and resized to make it larger

2. Finer control of opacity at low levels. When opacity is close to zero it is difficult to make minor adjustments. Another way of saying this is that small adjustments result in large changes of the displayed transparency.

PR #1143, opened 2018-12-28
Issue1123 TFWidget opacity slider maps to opacity quadratically for DVR

StasJ:
#1123 

This results in a linear opacity adjustment for volume renderings. Flat renderers still have a linear opacity slider.

out

### DIFF
--- a/apps/vaporgui/TFWidget.cpp
+++ b/apps/vaporgui/TFWidget.cpp
@@ -56,6 +56,7 @@ TFWidget::TFWidget(QWidget* parent)
 	_mainHistoRangeChanged		= false;
 	_secondaryHistoRangeChanged = false;
 	_discreteColormap 			= false;
+	_isOpacityIntegrated        = false;
 	_mainVarName	  			= "";
 	_secondaryVarName 			= "";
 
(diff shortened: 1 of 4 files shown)


apps/vaporgui/TFWidget.cpp
```diff
@@ -1039,6 +1040,22 @@ string TFWidget::getTFVariableName(bool mainTF=true) {
...
+{
+    if (_rParams->GetName() == DVRParams::GetClassType())
+        return 100 * sqrtf(opacity);
+    else
+        return 100 * opacity;
+}
+
+float TFWidget::convertSliderValueToOpacity(int value) const
+{
+    if (_rParams->GetName() == DVRParams::GetClassType())
```
clyne: Why does this only apply to the DVR renderer?

StasJ: The opacity of a volume rendering is proportional to the integrated opacity along the ray. This equals the opacity squared. The reason the old opacity slider felt unnatural (no change at higher slider values and too rapid change at lower slider values) is because we were essentially mapping the square root of the slider to the resulting opacity. If we apply the new mapping to other renderers which do not integrate opacity, we will be mapping the slider squared to opacity which would make every other renderer feel wrong.

The proper OOP solution would be for a renderer to have an `IntegratesOpacity` property or something similar however the DVR is the only renderer I could think of that would need it so I just opted for this.

On a side note, we may want to consider implementing this for the TF opacity control points too, although this will create the issue that we can no longer linearly interpolate them and will require much more effort than the single slider.

clyne: How about simply adding a property to the TFWidget class that can be set by the user of the class, in this case DVRSubtabs.h?

PR #1146, opened 2018-12-30
Load Session: 2D variable in 2D/3D renderer not loaded properly

StasJ:
Fixed a bug where if you saved a session with a 2D variable with a renderer that supports both 2D and 3D variables, the VariablesWidget would reset the variable to a 3D variable upon loading the session file.

This feels like a hack but I couldn't think of a better fix currently.

### DIFF
--- a/apps/vaporgui/VariablesWidget.cpp
+++ b/apps/vaporgui/VariablesWidget.cpp
@@ -227,6 +227,7 @@ void VariablesWidget::set2DOrientation(const QString& orientation) {
     cout << "2D orientation is currently a no-op" << endl;
 }
 
+// This takes the dropdown menu index, not the dimension
 void VariablesWidget::setVariableDims(int index){
 	assert(_rParams);
 	if (! ((_dimFlags & TWOD) && (_dimFlags & THREED)) ) return;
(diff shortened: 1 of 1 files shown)


apps/vaporgui/VariablesWidget.cpp
```diff
@@ -370,6 +371,10 @@ string VariablesWidget::updateVarCombo(
 void VariablesWidget::updateScalarCombo() {
 	if (_variableFlags & SCALAR) {
 		string setVarReq = _rParams->GetVariableName();
+        
+        if (_dataMgr->VariableExists(_rParams->GetCurrentTimestep(), setVarReq))
+            _activeDim = _dataMgr->GetNumDimensions(setVarReq);
+
```
sgpearse: I believe that these four lines will have the correct behavior we put them in VariablesWidget::updateDimCombo().  I don't think this is the right place.

sgpearse:
@StasJ, I think the logic for this belongs in VariablesWidget::UpdateDimCombo().  It's currently in a function that updates the combo boxes for scalar variables, and as I understand it, this bug regards our vector variable combo boxes, and the dimension box associated with the vector variables.  Hope that makes sense.

I've pushed changes that I believe exhibit the correct behavior.  Lets chat and make sure we're thinking about the same problem.

PR #1156, opened 2019-01-02
Auto load NCAR site local file when on NCAR machine

StasJ:


StasJ:
I realized there might be something I overlooked so please postpone this PR

PR #1162, opened 2019-01-03
Fix for issue 1148

sgpearse:


StasJ:
The original bug is fixed however if you resize the window to make it smaller, it still exists.

screen shot

PR #1264, opened 2019-01-10
IsoSurface now uses constant color by default. 

shaomeng:


apps/vaporgui/PlotParams.cpp
```diff
@@ -40,7 +40,9 @@ static RenParamsRegistrar<PlotParams> registrar( PlotParams::GetClassType() );
 
 PlotParams::PlotParams( DataMgr* dmgr, ParamsBase::StateSave *ssave) 
           : RenderParams( dmgr, ssave, PlotParams::GetClassType()) 
-{ }
+{
+    SetUseSingleColor(true);
```
clyne: How does this change to PlotParams fix a problem with the isosurface renderer?

shaomeng: My bad git manipulations... Now it's at the right place.

shaomeng:
@clyne I believe this PR is ready now.

Issue #1204, opened 2019-01-07
Undo with IsoSurface renderer causes error state which leads to crash


StasJ:
The following steps are approximate. I did not narrow down exactly what caused this bug, but I was able to reproduce it 3 times in a row.

1. Open a dataset. (I confirmed with `/little_glade/stasj/lee.nc` and `Katrina`)
2. Add/enable an IsoSurfaceRenderer
3. Change the main variable a couple times
4. Change the fidelity
5. Play with the min/max values and iso values
6. Move the view
7. Undo between 5-20 steps

(Upon further testing, it appears steps 3 and 5 are most important)

Vapor will show the following error:

```
Error performing paint event
Failed to initialize renderer IsoSurface
Failed to load shaders!
Resource does not exist and unable to load by name
Failed to link shader:
ERROR: Compiled vertex shader was corrupt.
ERROR: Compiled fragment shader was corrupt.

Undefined variable name : NULL
```

If you close the error and delete/add a new IsoSurfaceRenderer, vapor will crash.

I ran a sanity test and performed the a similar series of actions with the TwoDData renderer without any issues.

**EDIT**

This issue appears to be limited to the IsoSurfaceRenderer. I tested the volume renderer too without any issues.

shaomeng:
There are two issues in this report: 1) OpenGL failure upon undo, and 2) crash. 2) seems to be caused by 1), so we address 1) first. 

Upon looking at the code, `_initializeGL()` was called at an `undo` event. Should an undo call `_initializeGL()`?  @clyne

Issue #1275, opened 2019-01-11
One undo operation generates six _paintGL() events. 


shaomeng:
To reproduce:
1. Insert a print statement at the beginning of the `_paintGL()` function of `RayCaster.cpp`.
2. Compile, open vapor, create an isosurface renderer, enable it, change the isovalue a little bit.
3. There are a few printouts in the terminal window. Remember this number.
4. Issue one `undo` operation. Notice there are more printouts in the terminal. 
5. Count the number again, it's 6 more than the previous number.

Side effect of this behavior:
1. Worse performance
2. Among these 6 paint events, earlier ones are issued when OpenGL environment isn't ready. For example, viewport could have zero dimensions. Some renderers might fail upon these conditions.

PR #1276, opened 2019-01-11
Fix Issue #1204

shaomeng:
Note to John: this is a band aid fix so RayCaster won't try to create a texture of size zero. This PR won't fix a bigger problem of weird behavior of `undo`, which is reported in issue  #1275.

### DIFF
--- a/lib/render/RayCaster.cpp
+++ b/lib/render/RayCaster.cpp
@@ -549,6 +549,11 @@ int RayCaster::_initializeGL()
     GLint viewport[4];
     glGetIntegerv( GL_VIEWPORT, viewport );
     std::memcpy( _currentViewport, viewport, 4 * sizeof(GLint) );
+    // newViewport contains zero width and height sometimes. 
+    //   Need to filter out those instances.
+    for( int i = 2; i < 4; i++ )
+        if( _currentViewport[i] < 1 )
+            _currentViewport[i] = 8;
 
     // Create any textures, framebuffers, etc.
     if( _initializeFramebufferTextures() != 0 )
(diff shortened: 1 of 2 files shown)


shaomeng:
This PR is ready to merge.

lib/render/RayCaster.cpp
```diff
@@ -511,6 +511,11 @@ int RayCaster::_initializeGL()
     GLint viewport[4];
     glGetIntegerv( GL_VIEWPORT, viewport );
     std::memcpy( _currentViewport, viewport, 4 * sizeof(GLint) );
+    // newViewport contains zero width and height sometimes. 
+    //   Need to filter out those instances.
+    for( int i = 2; i < 4; i++ )
```
clyne: I think a better fix would be to NOT perform any viewport dependent initialization here. I.e. inside of _initializeFramebufferTextures don't do:

    glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA32F, _currentViewport[2], _currentViewport[3],
                 0, GL_RGBA, GL_FLOAT, nullptr);

glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA32F, _currentViewport[2], _currentViewport[3],
                 0, GL_RGBA, GL_FLOAT, nullptr);

Because you are using phony values and then correct them inside of _updateViewportWhenNecessary()

I.e. just call glTexImage2D inside of _updateViewportWhenNecessary

shaomeng: Does your suggestion mean that `glTexImage2d()` needs to be called every `_paintGL()` is called? 
I definitely try to avoid that from happening, since viewport change is a pretty rare situation, and viewport being zero is more rare...

clyne:
No. You can simply initialize _viewport to all zeros and _updateViewportWhenNecessary() will do the rest. 

> On Jan 11, 2019, at 2:48 PM, Samuel Li <notifications@github.com> wrote:
> 
> @shaomeng commented on this pull request.
> 
> In lib/render/RayCaster.cpp:
> 
> > @@ -511,6 +511,11 @@ int RayCaster::_initializeGL()
>      GLint viewport[4];
>      glGetIntegerv( GL_VIEWPORT, viewport );
>      std::memcpy( _currentViewport, viewport, 4 * sizeof(GLint) );
> +    // newViewport contains zero width and height sometimes. 
> +    //   Need to filter out those instances.
> +    for( int i = 2; i < 4; i++ )
> 
> Does your suggestion mean that glTexImage2d() needs to be called every _paintGL() is called?
> I definitely try to avoid that from happening, since viewport change is a pretty rare situation, and viewport being zero is more rare...
> 
> —
> 

John Clyne
clyne@ucar.edu

shaomeng:
This PR is now ready to merge @clyne

Issue #1214, opened 2019-01-07
Automatic domain stretching doesn't work when multiple data sets loaded


clyne:
If Settings/automatically-stretch-domain is enabled loading multiple data sets the scaling for the second data set is not computed correctly. See attached image.

To reproduce:

1. Make sure "auto scaling" is disabled
2. Import the WRF data set from Data/CoupledWrfRoms
3. Import the ROMS data set from Data/CoupledWrfRoms. When prompted, select "new session"
![untitled](https://user-images.githubusercontent.com/6777736/50798694-7a6f0000-1296-11e9-9180-9e59b37f9be7.png)

PR #1286, opened 2019-01-14
Issue 1214

sgpearse:
Fix for #1214, regarding stretch factors with multiple datasets loaded


sgpearse:
Strange.  I'm able to build on my OSX machine, but it looks like I'm seeing the problem on cisl-vapor.  I had to include glm/gtc/matrix_transforms.hpp, which seems to have fixed it for me on cisl-vapor.  Let me know if that fixed it for you.

StasJ:
@sgpearse you are correct. I never fully re-wrote that code, I just kept patching it as issues arose. I patched it once again and now everything should work as expected in master. I need to consolidate it into one location since currently this same code exists in 3 places (legacy). That requires more significant changes so I'll do a PR for this later.

lib/render/AnnotationRenderer.cpp
```diff
@@ -278,40 +279,264 @@ void AnnotationRenderer::applyTransform(Transform *t) {
...
+    const glm::mat4 translationMatrix,
+    const glm::mat4 translateOriginMatrix
+) const {
+    // transform our corner
+    glm::vec4 transformedCorner;
+    transformedCorner = glm::inverse(translateOriginMatrix) * dataMgrCorner;
+    transformedCorner = rotationMatrix * transformedCorner;
+    transformedCorner = translationMatrix * transformedCorner;
+    transformedCorner = scalingMatrix * transformedCorner;
+    transformedCorner = translateOriginMatrix * transformedCorner;
```
StasJ: Matrices can be concatenated. You are supposed to use them like this:

```
M = identity
M *= trans
M *= rot
M *= scale

Final vert = M * input vert 
```

include/vapor/AnnotationRenderer.h
```diff
@@ -108,9 +109,44 @@ class RENDER_API AnnotationRenderer : public MyBase {
...
+        glm::mat4 &scalingMatrix,
+        glm::mat4 &rotationMatrix,
+        glm::mat4 &translationMatrix,
+        glm::mat4 &translateOriginMatrix
+    ) const;
+
+    void _applyDataMgrCornerToDomain(
+        std::vector<double> &domainExtents,
+        const glm::vec4 dataMgrCorner,
+        const glm::mat4 scalingMatrix,
```
StasJ: See my other comment but you don't need to pass around all these matrices

include/vapor/AnnotationRenderer.h
```diff
@@ -108,9 +109,44 @@ class RENDER_API AnnotationRenderer : public MyBase {
 		double width,
 		std::vector<double> color);
 
-//! Render the domain fram
-	void drawDomainFrame(size_t ts) const;
-
+    void _makeTransformMatrices(
+        const Transform* transform,
+        glm::mat4 &scalingMatrix,
+        glm::mat4 &rotationMatrix,
```
StasJ: I am against using non-const pass by reference in the function argument list as it leads to ambiguous code. What I mean by this is if you are passing a list of parameters, one could be a surprise output. This is a common issue when trying to parse legacy code.

```
func(a, b, c) // b is modified. Surprise!
func(a, &b, c) // I know right away that b will be an output.
```

This is not one of our official coding practices (but I would like to propose it) so feel free to reject this review.

lib/render/AnnotationRenderer.cpp
sgpearse: Interesting.  That's cleaner and I'll adopt this for my vectors and matrices moving on.

sgpearse:
@stasj and @clyne,

I'm not thrilled about it, but I believe that the behavior we're seeing when rotating and scaling on the same axis is correct.  At the least, it's what we were doing before my changes.  We haven't specified new requirements for how this should work and I think we need to talk about it.

sgpearse:
@StasJ Thanks for the advice.  Should be good to go.

sgpearse:
Hmm I cannot reproduce this @StasJ.  Which file are you opening?

StasJ:
I'm using `Katrina-3.0.0.nc`

If you click on the video you'll be able to see more precisely what I'm doing.

out

StasJ:
Turns out this is caused by an outdated version of Katrina.

PR #1287, opened 2019-01-14
Fix for 1283

sgpearse:


shaomeng:
This change is applied on class `VariablesWidget` which is used across many different components of VAPOR, instead of on an IsoSurface specific class (`IsoSurfaceSubtabs` for example). Does it change the behavior of other GUI components outside of IsoSurface renderer then?

clyne:
That should be fine.

> On Jan 14, 2019, at 3:28 PM, sgpearse <notifications@github.com> wrote:
> 
> Yes, this means that no renderer GUI's will be able to select '0' as a color mapped variable.
> 
> This should really only be controlled in one place, within the Appearance tab, under the 'use constant color' checkbox.
> 
> —
> You are receiving this because you are subscribed to this thread.
> 

John Clyne
clyne@ucar.edu

Issue #1310, opened 2019-01-16
colorbar doesn't work with isosurface


clyne:
The colorbar exhibits strange behavior when used with the isosurface renderer: it has a multi-color checkered background and if the scene is rotated the colorbar becomes cropped in strange ways.

1. Load Duku
2. Enable an isosurface
3. Check "Colorbar Enabled" on the Annotation subtab. Note the checkerboard background of the colorbar
4. Rotate the scene and note how the colorbar becomes cropped.

<img width="398" alt="untitled" src="https://user-images.githubusercontent.com/6777736/51260467-972cc700-196b-11e9-8f0f-ff43492716af.png">

PR #1313, opened 2019-01-16
Fix for issue 1310

sgpearse:
#1310

### DIFF
--- a/lib/render/Renderer.cpp
+++ b/lib/render/Renderer.cpp
@@ -491,7 +491,8 @@ void Renderer::renderColorbar(){
 	if (makeColorbarTexture()) return;
 
     lgl->Color4fv(whitecolor);
-	
+
+    glActiveTexture( GL_TEXTURE0 );	
 	glTexParameterf( GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR );
 	glTexParameterf( GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR );
 	//Disable z-buffer compare, always overwrite:


lib/render/Renderer.cpp
```diff
@@ -491,7 +491,8 @@ void Renderer::renderColorbar(){
 	if (makeColorbarTexture()) return;
 
     lgl->Color4fv(whitecolor);
-	
+
+    glActiveTexture( GL_TEXTURE0 );
```
clyne: Is this need because some other renderer is not leaving OpenGL in the default state? I.e. is there a _paintGL() method that returns without restoring the active texture to GL_TEXTURE0. If that is the case better to fix the renderer or this problem will crop up again elsewhere

sgpearse: I believe this is the case.  Wouldn't it be prudent for all renderers to set their intended texture unit before drawing, and restore state after drawing?

StasJ: Yes that is the case, and we should strive to leave OpenGL in a default state, however I think that renderers should setup their own state within reason, and I think that setting the texture unit is reasonable.

clyne: Unfortunately because of the hundreds of state flags OpenGL supports it is not practical to require each renderer to check the value of every state flag that might impact its rendering. This places too much of a burden on the renderer developer. That is why the policy has always been that a renderer developer can assume OpenGL state flags are in the default state on entry to paintGL() (with a couple of exceptions), and a renderer developer is responsible for restoring any changes to state made within paintGL() to their default values before returning from paintGL(). The default active texture unit is GL_TEXTURE0.

Given that this problem is associated with the isorenderer my guess is that it's @shaomeng code that is not cleaning up state. You might want to kick this back over to Sam

sgpearse: Okey dokey.  @shaomeng, I'm kicking it over to you.

PR #1314, opened 2019-01-16
Fix for 1306

sgpearse:


shaomeng:
The "Scale" tab has the current data set name, while the "Translation", "Rotation", and "Origin" tabs do NOT have the data set name. This is an UI inconsistency for future improvement.

apps/vaporgui/VaporTable.cpp
```diff
@@ -381,6 +381,7 @@ void VaporTable::setVerticalHeader(std::vector<std::string> header) {
 	_table->setVerticalHeaderLabels(list);
 	_table->resizeRowsToContents();
 	_table->verticalHeader()->setResizeMode(QHeaderView::Stretch);
+    _table->verticalHeader()->setMaximumWidth(100);
```
clyne: You're hard-coding the header width (whatever that is) for all instances of the VaporTable. Isn't that likely to break something somewhere else? Wouldn't it be better to add a wrapper method to the VaporTable class and let the user of the class make this change if needed?

sgpearse: Maybe we can talk about this.  I think that in some situations it's reasonable to hard code values (IE TabManager).  I'm pretty certain that this isn't going to break anything.

apps/vaporgui/VaporTable.cpp
```diff
@@ -98,6 +99,14 @@ void VaporTable::Update(int rows, int cols,
 	resizeTableHeight();
 }
 
+void VaporTable::SetVerticalHeaderWidth(int width) {
+    _verticalHeaderWidth = width;
```
clyne: Seems like {Set,Get}VerticalHeaderWidth could be pure wrapper functions, no? I.e. call _table->{set,get}VerticalHeaderWidth directly?

sgpearse:
> Do we want or need to retrieve it?

@clyne I think so.  I can't think of a good reason to have a setter without a getter.

sgpearse:
@clyne Should be good to go now.

Issue #1300, opened 2019-01-15
Geometry subtab for image renderer has a Z manipulator


clyne:
The Geometry tab for the Image renderer has X, Y, and Z region selectors. Z should not exist. Moreover, if the model selector is changed from Navigation to Region a 3D bounding box is shown. This should be 2D.

Load any geo-referenced data set
Enable the Image renderer and select the Geometry subtab. The Z min and max selectors are shown
Now change the "mode" in the title bar from Navigation to Region. A 3D bounding box will be shown.

N.B. the using any of the Z controls results in inexplicable transforms.

PR #1316, opened 2019-01-16
Fix for 1300

sgpearse:
Fix for #1300


clyne:
> I think our code style says to delete code such as this instead of comment it out?

Correct. It should be deleted, or use 
#ifdef VAPOR_3_1_1 

if you are unsure that it should be deleted and may need to be resurrected.

sgpearse:
@shaomeng and @clyne 
Whoops.  That slipped past my radar.  Fixed now.

Issue #1301, opened 2019-01-15
Image Capture: GUI and Final file format not in agreement


shaomeng:
To reproduce: 
1. Load any data, create a 2D renderer
2. Select "capture single image." Also select file type to be ".tif, .tiff." (The screenshot shows this sequence.)
3. Click Save. The saved image is still .png format.

![peek 2019-01-15 15-33](https://user-images.githubusercontent.com/814885/51214730-fd6a0900-18db-11e9-8db3-04456e3b8ed6.gif)

sgpearse:
Notes:
https://bugreports.qt.io/browse/QTBUG-27186

PR #1330, opened 2019-01-18
Issue 1301 - Fix for broken jpeg/tiff on Ubuntu

sgpearse:
Fix for #1301


sgpearse:
@clyne Yes, I see that now.  It's now fixed.

Issue #1396, opened 2019-01-26
VAPOR crash upon creating new visualizer:


shaomeng:
To reproduce:
1. Open VAPOR: `./bin/vapor /little_glade/data/VDC/vapor-3.0.0/WRF/DUKU/DUKU-3.0.0.nc`
2. Select "Create new visualizer"
3. VAPOR crashes. 

Platform: Ubuntu only

shaomeng:
@sgpearse , please take a look at this bug, and also discussion happened at #1409 .

PR #1409, opened 2019-01-28
fix bug #1396

shaomeng:
The root of this issue is that `Throwing exceptions from an event handler is not supported in Qt.`

I changed it so that an error of our custom type is reported. Please review and determine if this is a good solution.

### DIFF
--- a/apps/vaporgui/VaporTable.h
+++ b/apps/vaporgui/VaporTable.h
@@ -6,6 +6,8 @@
 #include <QLineEdit>
 #include <sstream>
 
+#include <vapor/MyBase.h>
+
 struct Value;
 
 class CustomLineEdit : public QLineEdit {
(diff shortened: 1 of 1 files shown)


apps/vaporgui/VaporTable.h
```diff
@@ -165,17 +167,24 @@ public slots:
...
+    template<typename T>
+    operator T() const   //implicitly convert into T
+    {
+        std::stringstream ss(_value);
+        T convertedValue;
+        if ( ss >> convertedValue )
+            return convertedValue;
+        else
+        {
+            std::string msg = "conversion failed: " + _value;
```
clyne: This won't fly either. If <T> is a string the else clause will return a NULL. Moreover there is no error checking performed for by VaporTable::GetValue() - the only user of Value - so the error message recorded by SetErrMsg will never be reported.

Not sure what should be done here.

shaomeng:
I would think a more reasonable design is to have ”VaporTable" truthfully
return a string that's stored there, and each individual widget whoever
makes use of "VaporTable" to take the responsibility of converting a string
to the data type it prefers...

shaomeng:
Since the issue requires better fix than this band-aid, the issue is kicked back to @sgpearse and this PR is closed.

shaomeng:
This isn't even a band-aid. VAPOR still crashes occasionally with this fix. There's a deeper problem...

PR #1443, opened 2019-02-01
Mac DVR fix

shaomeng:
This PR replaces a few uses of `glBufferData` with `glBufferSubData`.

The benefits are:
1) it improves performance (by a small amount)
2) it behaves more friendly with mac. In fact, it stop crashing on Sam's mac. However, it still crashes on John's mac.

### DIFF
--- a/include/vapor/RayCaster.h
+++ b/include/vapor/RayCaster.h
@@ -207,7 +207,6 @@ protected:
     void _updateColormap( RayCasterParams* params );
     void _updateDataTextures();
     int  _updateVertCoordsTexture( const glm::mat4& MV );
-    void _enableVertexAttribute( const float* buf, size_t length, bool attrib1Enabled ) const;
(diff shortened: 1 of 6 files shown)


clyne:
Discard is supposed to behave that way, but there is a comment in the VAPOR2 code that says discard is broken on the mac.



> On Jan 31, 2019, at 6:19 PM, StasJ <notifications@github.com> wrote:
> 
> @StasJ approved this pull request.
> 
> Discard terminates the thread so there is no need to return after. Just FYI
> 
> —
> You are receiving this because you are subscribed to this thread.
> 

John Clyne
clyne@ucar.edu

shaomeng:
Discard certainly hasn't given me any surprises yet... Maybe they fixed the
VAPOR2 era bug? But I thought it doesn't hurt to have a return statement
there.

PR #1461, opened 2019-02-08
3 less textures for ray caster

shaomeng:
This PR eliminates the use of 3 textures in the ray caster.


shaomeng:
Not needed with the new DVR implementation.

PR #1463, opened 2019-02-13
better approximation of point in cell test

shaomeng:
The original point inside of a cell test shipped with VAPOR3.1RC0 will report false negatives (i.e., a point IS inside the cell is reported false). An accurate such test would involve tessellations. 

This PR makes it report false positives (i.e., a point is NOT inside the cell is reported true) instead of false negatives, so the ray casting algorithm could carry on.

This approximation is expected to be very close to the ground truth with not-very-concave cells.


shaomeng:
Not needed with the new DVR implementation.

PR #1483, opened 2019-03-08
Added default variable for Z axis w.r.t. Vector variables

sgpearse:


sgpearse:
You are correct.  Changes have been added per your note.

PR #1487, opened 2019-03-12
Added VaporWidgets

sgpearse:
Added new widgets that encapsulate labels+spacers+miscComponents for easier plug and play gui creation.


apps/vaporgui/VaporWidgets.h
```diff
@@ -0,0 +1,112 @@
...
+    VPushButton(
+        QWidget* parent, 
+        std::string labelText = "Label",
+        std::string buttonText = "Button"
+    );
+
+    void SetLabelText(  const std::string text );
+    void SetLabelText(  const QString text );
+    void SetButtonText( const std::string text );
+    void SetButtonText( const QString text );
```
shaomeng: Line 29 - 32 where a new string is passed in, a reference type is preferred. By using references, the application does not need to make a copy. For complex classes, i.e., strings, this copy could be expensive.

apps/vaporgui/VaporWidgets.h
```diff
@@ -0,0 +1,112 @@
...
+public:
+    VComboBox(
+        QWidget* parent,
+        std::string labelText = "Label"
+    );
+    void        SetLabelText(  const std::string text );
+    int         GetCurrentIndex() const;
+    std::string GetCurrentText() const;
+    void        AddOption( std::string option, int index=0 );
+    void        RemoveOption( int index );
```
shaomeng: Similar arguments here regarding reference type.

apps/vaporgui/VaporWidgets.cpp
```diff
@@ -0,0 +1,206 @@
...
+std::string VComboBox::GetCurrentText() const {
+    return _myCombo->currentText().toStdString();
+}
+
+void VComboBox::AddOption( std::string option, int index) {
+    _myCombo->insertItem( index, QString::fromStdString(option) );
+}
+
+void VComboBox::RemoveOption( int index=0 ) {
+    _myCombo->removeItem( index ) ;
```
shaomeng: what if `index` passed in is out of bound?

apps/vaporgui/VaporWidgets.cpp
```diff
@@ -0,0 +1,206 @@
...
+int VComboBox::GetCurrentIndex() const {
+    return _myCombo->currentIndex();
+}
+
+std::string VComboBox::GetCurrentText() const {
+    return _myCombo->currentText().toStdString();
+}
+
+void VComboBox::AddOption( std::string option, int index) {
+    _myCombo->insertItem( index, QString::fromStdString(option) );
```
shaomeng: what if `index` isn't a valid value?

apps/vaporgui/VaporWidgets.h
```diff
@@ -0,0 +1,112 @@
...
+class VCheckBox : public QWidget, public Ui_VCheckBox
+{
+    Q_OBJECT
+
+public:
+    VCheckBox(
+        QWidget* parent,
+        std::string labelText = "Label"
+    );
+    void SetLabelText( std::string text );
```
shaomeng: Same argument here regarding reference, e.g., `const std::string&`

apps/vaporgui/VaporWidgets.h
```diff
@@ -0,0 +1,112 @@
...
+
+signals:
+    void _pathChanged();
+
+private:
+    std::string _filePath;
+};
+
+
+#endif // VAPORWIDGETS_H
```
shaomeng: A few questions regarding a `PathSelector`:
1. Does/Should it support selecting both a folder and a file?
2. Does/Should it support selecting both a single file/folder and multiple files/folders?
3. Does/Should it support file type filters? 
4. Does/Should it check file permissions upon a selection is made?

apps/vaporgui/VaporWidgets.h
sgpearse: Got it.  You're right.  std::string isn't a built-in so pass by const reference makes sense.  Thanks for the catch.

apps/vaporgui/VaporWidgets.cpp
sgpearse: If index is out of bounds, nothing happens.

apps/vaporgui/VaporWidgets.cpp
sgpearse: If index is negative, the item gets prepended.  If it's greater than the count, it gets appended.

apps/vaporgui/VaporWidgets.cpp
shaomeng: Should this function return an error code to notify the caller if "nothing happens"?
(Brainstorming)

apps/vaporgui/VaporWidgets.h
sgpearse: 1) I've added file and directory selection options in a recent push
2) I don't think so.  This widget is too small for it.  I can implement another widget that has a QTextEdit so we can display a list of files, but a QLineEdit should really only hold one thing, not a list.
3) I've added this as well, thanks for catching it.
4) Yes, it does for reading files.  Not for writing though.  I'm not certain if we currently need it yet, but if you have a use-case I'll get it in there.

apps/vaporgui/VaporWidgets.cpp
sgpearse: I don't know.  The Qt guys didn't seem to think it's needed.  We can if you like though.

apps/vaporgui/VaporWidgets.h
sgpearse: Got it.

apps/vaporgui/VaporWidgets.cpp
shaomeng: @clyne  what do you think on this topic? I.e., should a composite widget of our own check if the indices passed in are valid?

apps/vaporgui/VaporWidgets.h
shaomeng: I can think of a use case of reading a file: the flow renderer will need to be able to read in a list of seeding points. Can we add this capability now?

apps/vaporgui/VaporWidgets.cpp
clyne: I'm not sure any purpose is served by informing the user of the class that they have tried to delete something that doesn't exist. I.e. I think silently ignoring an invalid request is reasonable in this case.

shaomeng: Let's not check out of bound then.

apps/vaporgui/VaporWidgets.h
sgpearse: Yup.

apps/vaporgui/VaporWidgets.h
sgpearse: Done.

apps/vaporgui/VaporWidgets.h
```diff
@@ -0,0 +1,135 @@
...
+class QLabel;
+class QComboBox;
+class QCheckBox;
+class QPushButton;
+class QLineEdit;
+class QDoubleValidator;
+class QSpacerItem;
+class QHBoxLayout;
+
+class VaporWidget : public QWidget
```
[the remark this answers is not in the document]
sgpearse: Destructors are called through the ownership tree, through the this*.  

The setLayout() gives ownership of layouts to the QWidget that the function is called on, but I get that this isn't obvious.  I've changed the layout to get a this* in its constructor, and added a comment at line 38 noting that the spacer's ownership is given to the VaporWidget.

apps/vaporgui/VaporWidgets.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+void VPathSelector::SetPath( const std::string& path ) {
+    _filePath = path;
+    _lineEdit->setText( QString::fromStdString(path) );
+    std::cout << "set path to " << _filePath << std::endl;
+}
+
+void VPathSelector::_openFileDialog() {
+    QString title = "Select file containing seed points";
+    std::cout << "GetPath returns " << GetPath() << std::endl;
```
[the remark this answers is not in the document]
sgpearse: Ok.

apps/vaporgui/VaporWidgets.h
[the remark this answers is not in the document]
shaomeng: Just curious, QT documentation says the following on `addItem()`:
 `This function is not usually called in application code. To add a widget to a layout, use the addWidget() function.` 

I do see that `addWidget()` is used for adding a label. Why is `addItem()` used for the spacer?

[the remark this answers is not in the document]
sgpearse: This is because addWidget() operates on QWidgets, which QSpacerItems are not.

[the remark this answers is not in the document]
shaomeng: Sounds good!

PR #1488, opened 2019-03-12
Flow gui

sgpearse:
Added a seeding tab to the Flow GUI.


apps/vaporgui/FlowEventRouter.cpp
```diff
@@ -62,6 +69,7 @@ FlowEventRouter::FlowEventRouter( QWidget *parent, ControlExec *ce)
 FlowEventRouter::~FlowEventRouter(){
 	if (_variables)  delete _variables;
 	if (_geometry)   delete _geometry;
+	if (_seeding)    delete _seeding;
 	if (_appearance) delete _appearance;
 	if (_annotation) delete _annotation;
 }
```
shaomeng: Who has the ownership of these widgets, (i.e., _variables, _geometry, _seeding) and who is responsible of cleaning them up?

sgpearse: These are all subtabs (not TFWidgets, GeometryWidgets etc) and are all owned by the EventRouter.  SubTabs do not share a base class.  They just hold GUI components.

EventRouter -> SubTab -> TabWidgets

apps/vaporgui/FlowEventRouter.cpp
```diff
@@ -44,6 +44,13 @@ FlowEventRouter::FlowEventRouter( QWidget *parent, ControlExec *ce)
 	qsapp->setWidgetResizable(true);
 	addTab(qsapp,"Appearance");
 
+	_seeding = new FlowSeedingSubtab(this);
+	QScrollArea* qsseed = new QScrollArea(this);
+	qsseed->setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
```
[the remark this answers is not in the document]
sgpearse: This is something that has persisted from legacy versions of our subtabs.  Hypothetically we should not need this because all of our widgets should have a size policy that does not push the tabs beyond the width of the control panel.  However this could still occur if somebody wanted a label with a really long name.

The true fix for this would be to have a control panel that can be dynamically sized, but it's currently hard coded.  I don't know why it was done this way.  Like a lot of our legacy code, I suggest keeping this as-is until the control panel can be reassessed at a higher level.

apps/vaporgui/FlowEventRouter.cpp
sgpearse: You're correct.  I've removed the redundant deletions.

Issue #1476, opened 2019-02-28
Dataset scale not preserved in session if automatic scaling enabled


StasJ:
1. Ensure automatic scaling is enabled
2. Open a dataset and scale it
3. Save the session
4. Re-open the session

The dataset scale will be reset

StasJ:
If the dataset scale is set to (1,1,1) it will be reset

StasJ:
Related to #1171

PR #1526, opened 2019-05-13
Fix for issue 1476

sgpearse:


sgpearse:
#1476

apps/vaporgui/NavigationEventRouter.cpp
```diff
@@ -280,6 +280,16 @@ void NavigationEventRouter::_performAutoStretching(string dataSetName) {
...
 	
 	for (int i=0; i<winNames.size(); i++) {
+		ViewpointParams *vpParams = paramsMgr->GetViewpointParams(winNames[i]);
+		Transform* transform = vpParams->GetTransform(dataSetName);
+        std::vector<double> scales = transform->GetScales();
+        int zDimension = 2;
+
+        // If the z-dimension scale is not 1.f, the user has saved a session with
+        // a non-default value.  Don't modify it.
+        if ( scales[ zDimension ] != 1.f )
```
clyne: The new code assumes that only scaling along the Z dimension is possible, while the code below is more general and supports scaling along any axis.

sgpearse: That is correct.  Good to go now.

Issue #1523, opened 2019-05-12
Renderer Controls disappear with multiple visualizers


shaomeng:
**Describe the bug**
The controls of multiple visualizers (bottom left of VAPOR) would display with multiple visualizers. 

**To Reproduce**
1. Open VAPOR, load a data set, create another visualizer. 
2. Create a renderer in one of the two visualizers. 
2. Click on the second visualizer.
3. The renderer controls disappear for both visualizers. 

**Screenshots**

![May-12-2019 14-05-55](https://user-images.githubusercontent.com/814885/57587238-22a00a00-74bf-11e9-97e7-ee3d88b751fd.gif)

sgpearse:
John fixed this bug with one of his own PR's.  

@clyne says:
"The fix I just merged for #1481 also fixes #1523. I think we can reject this PR and close 1523."



https://github.com/NCAR/VAPOR/pull/1536

Issue #1481, opened 2019-03-08
Deleting renderer after closing window causes crash


StasJ:
OS: OSX

1. Create/enable a renderer
2. Close the current visualizer window by pressing `Command-W` or the red X
3. Delete the renderer

Vapor will crash

PR #1536, opened 2019-05-15
Fix for issue_1523

sgpearse:
Fix for #1523


shaomeng:
Hi @clyne , I'm too not familiar with this part of the code, that I feel I'm not qualified to review. How about you give it a review instead?

clyne:
The fix I just merged for #1481 also fixes #1523. I think we can reject this PR and close 1523.

PR #1539, opened 2019-05-16
Fix for overlooked mistake in in statement

sgpearse:
The previous fix for #1476 had a mistake in a condition for performing auto scaling.  This is a fix for that error.


sgpearse:
Dang, this was correct in the first place.  Please disregard.

Issue #1535, opened 2019-05-15
New Visualizer resets scale factors


sgpearse:
Creating a new visualizer resets that visualizer's scale factors to 1.

To reproduce:
1) Open any dataset and adjust the Z-axis scale factor, if it hasn't been done so automatically
2) Create a new visualizer.  That visualizer's scale factor will be reset to 1.  The original visualizer's scale factor will be unchanged.  The GUI will reflect the scale factor of the new visualizer, which has been reset.

![Screen Shot 2019-05-15 at 10 02 03 AM](https://user-images.githubusercontent.com/9522770/57790704-83feed80-76f8-11e9-8789-814f8d0392a2.png)

PR #1540, opened 2019-05-16
Issue 1535

sgpearse:
Fix for issue #1535

### DIFF
--- a/apps/vaporgui/NavigationEventRouter.cpp
+++ b/apps/vaporgui/NavigationEventRouter.cpp
@@ -289,9 +289,11 @@ void NavigationEventRouter::_performAutoStretching(string dataSetName) {
 
         // If a dimension's scale is not 1.f, the user has saved a session with
         // a non-default value.  Don't modify it.
-        if ( scales[ xDimension ] != 1.f &&
-             scales[ yDimension ] != 1.f &&
-             scales[ zDimension ] != 1.f )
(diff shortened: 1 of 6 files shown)


apps/vaporgui/NavigationEventRouter.cpp
```diff
@@ -289,11 +289,9 @@ void NavigationEventRouter::_performAutoStretching(string dataSetName) {
...
         // If a dimension's scale is not 1.f, the user has saved a session with
         // a non-default value.  Don't modify it.
-        if ( scales[ xDimension ] != 1.f )
-            continue;
-        if ( scales[ yDimension ] != 1.f )
-            continue;
-        if ( scales[ zDimension ] != 1.f )
+        if ( scales[ xDimension ] != 1.f &&
+             scales[ yDimension ] != 1.f &&
+             scales[ zDimension ] != 1.f )
```
clyne: Same logic issue as in PR #1539. Not sure why these same changes are appearing in two PRs....

sgpearse: Right.  This is because I needed to merge this branch with #1539 for testing.

Both this branch and #1539 contain the same commit that applies these changes, so there shouldn't be any conflicts.

sgpearse: It turns out that this was correct the first time.  I'll remove it from this branch.

PR #1568, opened 2019-05-30
Warnings

StasJ:


include/vapor/VolumeRenderer.h
```diff
@@ -58,10 +58,10 @@ class RENDER_API VolumeRenderer : public Renderer
...
     bool _needToSetDefaultAlgorithm() const;
     
-    unsigned int _VAO = NULL;
-    unsigned int _VBO = NULL;
-    unsigned int _VAOChunked = NULL;
-    unsigned int _VBOChunked = NULL;
+    unsigned int _VAO = (int)NULL;
+    unsigned int _VBO = (int)NULL;
+    unsigned int _VAOChunked = (int)NULL;
+    unsigned int _VBOChunked = (int)NULL;
```
shaomeng: 1. I think we better give these `unsigned int` value of 0?

include/vapor/VolumeRenderer.h
```diff
@@ -58,10 +58,10 @@ class RENDER_API VolumeRenderer : public Renderer
...
     VolumeAlgorithm *_algorithm = NULL;
```
shaomeng: We decided to adtop c++11 convention, so NULL is discouraged, and nullptr is encouraged.

lib/render/ShaderManager.cpp
```diff
@@ -106,7 +106,7 @@ SmartShaderProgram ShaderManager::GetSmartShader(const std::string &name)
 int ShaderManager::LoadResourceByKey(const std::string &key)
 {
 //    printf("Begin Compile %s\n", key.c_str());
-    void *t = GLManager::BeginTimer();
+//    void *t = GLManager::BeginTimer();
```
shaomeng: Remove commented code? The same applies to a few other comments.

StasJ: Those are there for debugging in the future

include/vapor/VolumeRenderer.h
StasJ: These are handles that happen to be represented as integers. NULL indicates that they are not just numbers.

include/vapor/VolumeRenderer.h
StasJ: Fixed in #b2e2b92

PR #1572, opened 2019-05-31
Reset locale to C

StasJ:
I found some documentation referencing this issue:

https://doc.qt.io/qt-5/qcoreapplication.html#locale-settings

I reset the LC_ALL to "C" which is different from what the documentation states because we do not support any localization.


StasJ:
Fix #1574

Issue #1620, opened 2019-06-04
Import ROMS Crash


sgpearse:
To reproduce:

1) Launch Vapor on Casper.  Import the a few files in /glade/p/cisl/vast/vapor/data/Source/ROMS/jsmall.  **DO NOT INCLUDE THE GRID FILE**
2) Vapor crashes

clyne:
Kicking this back to @sgpearse. The code in AnnotationRenderer::_calculateDomainExtents assumes a data set will contain 2D or 3D variables. It may not.

PR #1665, opened 2019-06-11
Fix for 1620

sgpearse:
Fix for #1620 which crashed during the importation of ROMS data without grid files.  Specifically, crashes would occur when the dimensionality of a dataset was less than 2.


sgpearse:
Ok. We should be set now.

Issue #1656, opened 2019-06-07
Navigating the scene resets GUI tab


StasJ:
1. Add a renderer
2. Click on the `Navigation` or `Settings` tab
3. Navigate the scene

The GUI will reset back to the `Renderers` tab

PR #1675, opened 2019-06-12
Fix for 1656

sgpearse:
Fix for #1656


apps/vaporgui/TabManager.cpp
```diff
@@ -94,6 +94,7 @@ void TabManager::SetActiveRenderer(string activeViz, string renderClass, string
 
 void TabManager::ShowRenderWidget(string subTabName)
 {
+    cout << "MoveToFront " << subTabName << endl;
```
shaomeng: Shall we leave this print output to Master branch?

sgpearse: My bad!

PR #1680, opened 2019-06-13
Show warning about curvilinear DVR on Intel systems

StasJ:


apps/vaporgui/RenderHolder.cpp
```diff
@@ -224,6 +227,45 @@ void RenderHolder::_initializeNewRendererDialog(vector<string> datasetNames) {
...
+    if (rendererType != VolumeRenderer::GetClassType() &&
+        rendererType != VolumeIsoRenderer::GetClassType())
+        return;
+    
+    ParamsMgr *paramsMgr = _controlExec->GetParamsMgr();
+    SettingsParams *sp = (SettingsParams *)paramsMgr->GetParams(SettingsParams::GetClassType());
+    if (sp->GetDontShowIntelDriverWarning())
+        return;
+    
+    // Qt will automatically delete this for us apparently
```
shaomeng: I don't think this is the case. Scott has spent some time researching into QT memory management; you might want to ask him.

apps/vaporgui/RenderHolder.cpp
```diff
@@ -224,6 +227,45 @@ void RenderHolder::_initializeNewRendererDialog(vector<string> datasetNames) {
 	}
 }
 
+void RenderHolder::_showIntelDriverWarning(const string &rendererType)
```
shaomeng: What's the rational to display this warning message from RenderHolder, instead of from VolumeRenderer?

apps/vaporgui/RenderHolder.h
```diff
@@ -19,6 +19,16 @@ namespace VAPoR {
 	class ParamsMgr;
 }
 
+class QCheckBoxTest : public QCheckBox {
+    Q_OBJECT;
+    
+public:
+    QCheckBoxTest(QString s):QCheckBox(s){}
+    ~QCheckBoxTest() {
+        printf("Test::Destructor() +++++++++\n");
```
shaomeng: Please remove this print statement.

apps/vaporgui/RenderHolder.cpp
StasJ: The render library cannot display a Qt popup and we no not have a mechanism for propagating warnings

apps/vaporgui/RenderHolder.h
StasJ: 1. That is correct but this was the simplest solution.
2. Well some datasets are just slower I guess. I am currently rendering a 549*349*32 dataset which is 4 million cell larger in under a second. I don't see this as a reason to remove the option to render curvilinear datasets entirely.
3. This print statement was to confirm that Qt does in fact delete the object, as per your other question. I'll do that.

shaomeng: Interesting, I didn't know the performance would be that much different. 128^3 is probably the smallest resolution that's still useful to an user. My 128^3 data set is that sphere data set, essentially regular. Does the regular nature hurt performance of the curvi algorithm?

StasJ: No it doesn't make a difference if the grid is actually curvilinear or not. I don't know why you are experiencing worse performance. I can take a look at it though.

sgpearse:
Not sure if I'm splitting hairs here.  Do we want the user to be able to turn warnings back on?  I'm concerned that a user may not read the warning, silence it, then move on and never see it again.

If we had the 'do not show warnings' checkbox in the RenderHolder, they would be able to turn it off and on each time they make a new renderer.  We'd also be able to warn the user if they're making a 3D renderer when only 2D variables are available, as Sam has pointed out to me.

If this is something that can wait, I'm fine accepting the PR.

PR #1752, opened 2019-07-01
Issue1750: Curvilinear Raycaster crashes on AMD (Mesa drivers) + Ubuntu

### LINKED ISSUE #1750: Curvilinear Raycaster crashes on AMD (Mesa drivers) + Ubuntu
To reproduce:

1) On CISL-VAPOR2, start up vapor, import the WRF files in ~/Downloads/wrfout*
2) Create a volume renderer and enable it

StasJ:
Fix #1750 

I've added some comments to the issue


StasJ:
The code will work with any tree depth, it can only affect performance. My code skips the first few levels since they generally don't give any performance improvements and they provide the compiler with less complex code. In the case of the Mesa driver, it would attempt to compile the largest search trees even if they weren't being used while other compilers only compile the ones being used.

PR #1766, opened 2019-07-17
Changes to VWidgets

shaomeng:
@sgpearse  as we discussed, this PR 
1) comment out a dangerous practice which could lead to dangling pointers
2) catch a proper signal and rename the functions.

As I mentioned, I think it's actually acceptable to offload the validation work to whoever owns a VLineEdit widget, so VLineEdit is cleaner.

### DIFF
--- a/apps/vaporgui/VaporWidgets.cpp
+++ b/apps/vaporgui/VaporWidgets.cpp
@@ -145,13 +145,16 @@ double VDoubleSpinBox::GetValue() const {
     return _value;
 }
 
+//
+// ====================================
+//
+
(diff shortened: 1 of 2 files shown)


shaomeng:
@sgpearse  I think it's probably even better to not provide any validation at all here, since QValidator classes does **not** provide a `fixup()` implementation: https://doc.qt.io/archives/qt-4.8/qdoublevalidator.html

In other words, this widget won't know what to do if it receives an invalid input. It's best to leave this part of work to users who owns this widget.

PR #1770, opened 2019-07-23
introduce widget VSlider

shaomeng:


apps/vaporgui/VaporWidgets.h
```diff
@@ -138,6 +139,47 @@ private slots:
     std::string _text;
 };
 
+//
+// ====================================
+// VSlider augments QSlider by providing a text box displaying the value.
+//
+// Note: this widget is re-implemented by Sam instead of using the existing
+//   QSliderEdit class because:
+//   1) this class does NOT use a UI file following the VWidget convension;
```
sgpearse: Yes, this class does not use a UI file.  However the VWidget convention does not use UI files either.

apps/vaporgui/VaporWidgets.cpp
```diff
@@ -145,6 +146,127 @@ double VDoubleSpinBox::GetValue() const {
...
+{
+    return _currentVal;
+}
+
+void
+VSlider::_respondQSliderReleased()
+{
+    /* QSlider is always giving a valid value, so no need to validate range */
+    int newvalI = _qslider->value();
+    float perc  = (float)newvalI / 100.0f;
```
sgpearse: Could we change this variable name to "precision"?

shaomeng: I'll fix it in my next PR.

PR #1838, opened 2019-08-23
Circle ci

sgpearse:
Bring Circle test configuration into Master branch.


shaomeng:
The CircleCI dashboard shows compilation failures and successes. Does it
also show warnings if they do exist?

PR #1850, opened 2019-09-10
Fix for broken windows build

sgpearse:
I guess MSVC cannot compile dynamically sized arrays onto the stack...

### DIFF
--- a/lib/common/utils.cpp
+++ b/lib/common/utils.cpp
@@ -3,6 +3,8 @@
 #include <algorithm>
 #include <vapor/utils.h>
 
+#define MAXCOORDS 4
+
 using namespace std;
 using namespace Wasp;
 
@@ -50,8 +52,9 @@ size_t Wasp::LinearizeCoords(
 size_t Wasp::LinearizeCoords(
 	const size_t *coords, const size_t *dims, int n
 ) {
-	size_t* min = new size_t[n];
-	size_t* max = new size_t[n];
+	VAssert(n <= MAXCOORDS);
+	size_t* min = new size_t[MAXCOORDS];
+	size_t* max = new size_t[MAXCOORDS];
 	
 	for (int i=0; i<n; i++) {
 		min[i] = 0;
@@ -123,9 +126,9 @@ void Wasp::VectorizeCoords(
     const size_t *dims,
     size_t *coords, int n
 ) {
-	
-	size_t* min = new size_t[n];
-	size_t* max = new size_t[n];
+	VAssert(n <= MAXCOORDS);
+	size_t* min = new size_t[MAXCOORDS];
+	size_t* max = new size_t[MAXCOORDS];
 	for (int i=0; i<n; i++) {
 		min[i] = 0;
 		max[i] = dims[i]-1;
--- a/lib/common/utils.cpp
+++ b/lib/common/utils.cpp
@@ -53,8 +53,8 @@ size_t Wasp::LinearizeCoords(
 	const size_t *coords, const size_t *dims, int n
 ) {
 	VAssert(n <= MAXCOORDS);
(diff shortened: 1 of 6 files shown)


shaomeng:
>  If the Windoze C++ compiler can't handle allocating arrays on the stack when the dimension isn't known until run time than I'd suggest a constant allocation size. E.g.

John, I like your stylish spelling of Windows. 

Btw, I think VC is doing the right thing here... GCC is providing too much sugar coating by allowing it to compile at the first place...

sgpearse:
@clyne, your suggested fix has been added.  There are several other things that have been recently changed that MSVC does not like which needed fixing too, FYI.

lib/common/utils.cpp
```diff
@@ -50,8 +52,9 @@ size_t Wasp::LinearizeCoords(
 size_t Wasp::LinearizeCoords(
 	const size_t *coords, const size_t *dims, int n
 ) {
-	size_t* min = new size_t[n];
-	size_t* max = new size_t[n];
+	VAssert(n <= MAXCOORDS);
+	size_t* min = new size_t[MAXCOORDS];
```
clyne: @sgpearse I think you misunderstood my comment. These quantities should not be allocated off the heap because it's expensive and LinearizeCoords is often called inside large loops. They should be statically allocated from the stack:

size_t min[MAXCOORDS]
size_t max[MAXCOORDS]

lib/vdc/DataMgr.cpp
```diff
@@ -547,11 +547,18 @@ bool is_int(std::string str) {
...
 	}
 }
 
+// MSVC has a bug where isnan() is not overloaded for integral types.
+// This function specializes the template to bypass this bug.
+template <>
+void _sanitizeFloats(int *buffer, size_t n) {
+	return;
+}
+
```
shaomeng: I don't quite understand --- this function is called `_sanitizeFloats `, why does it have to handle integral types?
What's exactly the error Visual Studio throws if there's no this addition?

lib/vdc/DataMgr.cpp
```diff
@@ -547,11 +547,18 @@ bool is_int(std::string str) {
 
 template <class T>
 void _sanitizeFloats(T *buffer, size_t n) {
-	for (size_t i=0; i<n; i++) {
+	for (size_t i = 0; i < n; i++) {
 		if (isnan(buffer[i])) buffer[i] = std::numeric_limits<T>::infinity();
 	}
```
shaomeng: Can we try to specify the namespace, i.e., `std::isnan()`, and also include the required header, i.e., `#include <cmath>`, and then check again if it behaves as expected?

lib/vdc/DataMgr.cpp
sgpearse: It's a templated function, so it configures its specializations at compile time.

Since it can technically be passed an integral type, this specialization gets expanded and fails to compile.

MSVC not-so-helpfully reports the following error:
C2668 'fpclassify': ambiguous call to overloaded function

Strangely, this is at line 414 of math.h.  I'm not sure where it's being included.  The only files in our code base that include math.h are LegacyVectorMath.h and kdtree.c.

lib/vdc/DataMgr.cpp
sgpearse: We are currently including cmath.  Adding the namespace qualifier results in the same error.

lib/common/utils.cpp
sgpearse: Got it.  Fixed.

lib/vdc/DataMgr.cpp
shaomeng: Let's look at together on a windows machine. I'm not convinced that you create an empty function is a good fix to a potential programming error.

Issue #855, opened 2018-05-09
wrf2vdc: Can't convert infinity values


clyne:
wrf2vdc fails on data sets that have variables containing NaNs. 

To reproduce:
1. Run .wrfvdccreate Data/WRF/headwaters/wrfout_d01_2053-03-20_00:00:00 headMaster.nc
2. Run wrf2vdc Data/WRF/headwaters/wrfout_d01_2053-03-20_00:00:00 headMaster.nc

Generates the following message:

Error accessing netCDF file "/tmp/headMaster_data/data/UDROFF/UDROFF.0000.nc", nc_put_vara_double : NetCDF: Numeric conversion not representable -- file (/Users/clyne/src/vapor3/lib/wasp/NetCDFCpp.cpp), line(877)
Error accessing netCDF file "/tmp/headMaster_data/data/UDROFF/UDROFF.0000.nc", nc_put_vara_double : NetCDF: Numeric conversion not representable -- file (/Users/clyne/src/vapor3/lib/wasp/NetCDFCpp.cpp), line(877)

Note: it would be good to have a command line feature to the *vdccreate and *2vdc command line utilities to exclude variables by name

PR #1852, opened 2019-09-12
Issue 855

clyne:
This PR works around #855. The problematic file has a variable that contains NaNs. This PR simply provides a command line option that allows named variables to be excluded from copying.

### DIFF
--- a/lib/vdc/DataMgr.cpp
+++ b/lib/vdc/DataMgr.cpp
@@ -545,6 +545,13 @@ bool is_int(std::string str) {
 	return str.find_first_not_of("0123456789") == std::string::npos;
 }
 
+template <class T>
+void _sanitizeFloats(T *buffer, size_t n) {
+	for (size_t i=0; i<n; i++) {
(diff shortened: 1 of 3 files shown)


lib/vdc/DataMgr.cpp
```diff
@@ -545,6 +545,13 @@ bool is_int(std::string str) {
 	return str.find_first_not_of("0123456789") == std::string::npos;
 }
 
+template <class T>
+void _sanitizeFloats(T *buffer, size_t n) {
+	for (size_t i=0; i<n; i++) {
+		if (isnan(buffer[i])) buffer[i] = INFINITY;
```
shaomeng: This line has the following warning message, which doesn't seem very benign

````
/home/shaomeng/Git/VAPOR-flow-src/lib/vdc/DataMgr.cpp:551:35: warning: overflow in implicit constant conversion [-Woverflow]
   if (isnan(buffer[i])) buffer[i] = INFINITY;
````

PR #1861, opened 2019-10-03
Flow

shaomeng:
Hello guys, the flow functionality is ready for you to review! 

Note that 
1. The GUI still needs more refinement, so let's review the functionality first.
2. There are compiler warnings present right now. I'll eliminate them when I feel the functionality codes are more stable. 
2. There might be some serious bugs still present, so please don't spend too much time on this PR. If bugs or unexpected behavior keeps coming up, file them and then stop! 

Happy testing!


shaomeng:
Good widget! A few minor areas for improvement: 

1. With the following menu configuration, the "diverging" option doesn't seem to change anything when it's checked/unchecked. 
Screenshot from 2019-10-07 15-38-33

2. Sometimes it's hard to point to and click on the transfer function control points. It would improvement usability if the internal calculation considers a bigger area than the control points.

3. It seems I could input an arbitrary number in the box without having any impact on the rendering? 
Screenshot from 2019-10-07 15-57-31

lib/flow/UnsteadyVAPORScalar.cpp
```diff
@@ -0,0 +1,79 @@
+#include "vapor/UnsteadyVAPORScalar.h"
```
clyne: Dead code?

lib/flow/UnsteadyVAPORVelocity.cpp
```diff
@@ -0,0 +1,249 @@
+#include "vapor/UnsteadyVAPORVelocity.h"
```
clyne: Dead code?

lib/flow/VaporField.cpp
```diff
@@ -0,0 +1,424 @@
...
+                                       extMin, extMax );
+    if( _recentGrids.contains( key ) )
+    {
+        *gridpp = _recentGrids.find( key );
+        return 0;
+    }
+
+    // There's no such grid in our cache! Let's ask for it from the data manager,
+    // and then keep it in our cache!
+    VAPoR::Grid* grid = _datamgr->GetVariable( timestep, varName, refLevel, compLevel,
```
clyne: You are locking the grid data into memory. It needs to be unlocked with DataMgr::UnlockGrid()

clyne: GetVariable() will create a new Grid instance. This will also need to be freed. How is it being deleted?

lib/flow/VaporField.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+    return true;
+}
+
+
+bool
+VaporField::InsideVolumeScalar( float time, const glm::vec3& pos )
+{
+    if( ScalarName.empty() )
+        return NO_FIELD_YET;
```
clyne: Won't this result in a true value (the method returns a boolean)? Is that what you want? Why not just return false?

lib/flow/VaporField.cpp
```diff
@@ -0,0 +1,424 @@
...
+    return true;
+}
+
+void
+VaporField::GetFirstStepVelocityIntersection( glm::vec3& minxyz, glm::vec3& maxxyz )
+{
+    const VAPoR::Grid* grid = nullptr;
+    std::vector<double> min[3], max[3];
+
+    for( int i = 0; i < 3; i++ )
```
clyne: hard wired loop limits aren't so great. Better to loop over VelocitNames.size(). What happens with a 2D field?

lib/flow/VaporField.cpp
```diff
@@ -0,0 +1,424 @@
...
+            rv = _getAGrid( floorTS + 1, scalarname, &grid );
+            assert( rv == 0 );
+            float ceilScalar = grid->GetValue( coords );
+            if( ceilScalar  == grid->GetMissingValue() )
+            {
+                scalar = 0.0f;
+                return 0;
+            }
+            float weight = (time - _timestamps[floorTS]) /
+                           (_timestamps[floorTS+1] - _timestamps[floorTS]);
```
clyne: Check for divide by zero would be good.

lib/render/FlowRenderer.cpp
```diff
@@ -0,0 +1,1053 @@
...
+                      std::string&        instName,
+                      DataMgr*            dataMgr )
+          : Renderer( pm,
+                      winName,
+                      dataSetName,
+                      FlowParams::GetClassType(),
+                      FlowRenderer::GetClassType(),
+                      instName,
+                      dataMgr ),
+            _velocityField     ( 9 ),
```
clyne: Is this setting the cache size to 9 fields, and does that mean that up to 9 Grid instances might be kept in memory? Could it be reduced to 6 (3 velocity components * 2 time steps)? For big data this could be a problem.

lib/params/FlowParams.cpp
```diff
@@ -0,0 +1,273 @@
...
+            bools[i] = true;
+
+    return bools;
+}
+
+void              
+FlowParams::SetPeriodic( const std::vector<bool>& bools )
+{
+    std::vector<long> longs( 3, 0 );
+    for( int i = 0; i < 3; i++ )
```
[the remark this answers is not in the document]
shaomeng: In this particular case, my GUI code that calls `SetPeriodic` guarantees to pass in here a 3-valued vector.
But anyway, I added another layer of sanity check here.

include/vapor/Advection.h
```diff
@@ -0,0 +1,134 @@
...
+    int  OutputStreamsGnuplotMaxTime( const std::string& filename, 
+                                      float time, 
+                                      bool  append = false ) const;
+    int  InputStreamsGnuplot(  const std::string& filename );
+
+    // Query properties (most are properties of the velocity field)
+    int  CheckReady() const;
+
+    // Specify periodicity on each dimension
+    void  SetXPeriodicity( bool, float min = 0.0f, float max = 0.0f );
```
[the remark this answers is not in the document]
shaomeng: The boundary for periodic calculations. I've added a comment to it.

include/vapor/Field.h
```diff
@@ -0,0 +1,54 @@
...
+{
+public:
+    // Constructor and destructor
+    Field();
+    virtual ~Field();
+
+    // 
+    // If a given position at a given time is inside of this field
+    //
+    virtual bool InsideVolumeVelocity( float time, const glm::vec3& pos ) = 0;
```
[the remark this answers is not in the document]
shaomeng: Not in this case. I have a local cache that stores recently used grids, and this query would modify that cache.

include/vapor/Field.h
```diff
@@ -0,0 +1,54 @@
...
+    virtual bool InsideVolumeScalar(   float time, const glm::vec3& pos ) = 0;
+
+    //
+    // Retrieve the number of time steps in this field
+    //
+    virtual int GetNumberOfTimesteps() = 0;
```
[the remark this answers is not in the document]
shaomeng: This is oen getter that could be made const, and now the change is already made. however, the others cannot be const simply because they're modifying the internal cache state.

include/vapor/UnsteadyVAPORVelocity.h
```diff
@@ -0,0 +1,61 @@
+/*
```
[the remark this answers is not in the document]
shaomeng: Yeah, it's dead, and it's deleted from this PR.

lib/flow/SteadyVAPORScalar.cpp
```diff
@@ -0,0 +1,75 @@
+#include "vapor/SteadyVAPORScalar.h"
```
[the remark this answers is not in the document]
shaomeng: This is also deleted from this PR.

lib/flow/SteadyVAPORVelocity.cpp
```diff
@@ -0,0 +1,104 @@
+#include "vapor/SteadyVAPORVelocity.h"
```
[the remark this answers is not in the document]
shaomeng: This is also deleted from this PR.

lib/flow/UnsteadyVAPORScalar.cpp
shaomeng: This is also deleted from this PR.

lib/flow/UnsteadyVAPORVelocity.cpp
shaomeng: This is also deleted from this PR.

lib/flow/VaporField.cpp
shaomeng: The `GridWrapper` class was supposed to do the proper cleanup, but `GridWrapper` wasn't used. It is used now and makes sure grids are properly cleaned.

shaomeng: It is guaranteed to be unlocked and cleaned by the `GridWrapper` class, which itself is guaranteed to be cleaned up by the cache.

lib/flow/VaporField.cpp
shaomeng: Good point, now it just returns false.

lib/flow/VaporField.cpp
shaomeng: Hmm, the variable `VelocityNames` is declared as an array of 3, but not an std container. I believe this is a case where we could use hard wired loop limits.

lib/flow/VaporField.cpp
shaomeng: It's really an abnormal case to have zero here, so I added a `VAssert()` a few lines above.

lib/render/FlowRenderer.cpp
shaomeng: It is correct. Sure, I can change it to 6.

lib/params/FlowParams.cpp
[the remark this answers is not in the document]
clyne: Good idea. It just makes the code more robust :-)

include/vapor/Field.h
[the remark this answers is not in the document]
clyne: Understood. FYI the 'mutable' keyword can be used in cases like these, though not suggesting that you need to make use of it:-)

See https://arne-mertz.de/2017/10/mutable/

lib/flow/VaporField.cpp
clyne: How about adding a comment stating this?

lib/flow/VaporField.cpp
clyne: Nevertheless, the application crashes when I change one of the variables in a 3D data set to "no variable". This may be part of the problem.

clyne:
> Hey John, I've fixed or commented on all of the items you pointed out. Please see if they're good now!

Cool. I've resolved most of the comments, and there are just a couple more that are unresolved.

lib/flow/VaporField.cpp
shaomeng: with the latest changes, there is a comment stating this.

lib/flow/VaporField.cpp
shaomeng: Let's mark that crash as a bug, and I'm pretty sure it isn't caused by this loop. 
(Given how this variable is declared: `std::string VelocityNames[3];` it'll always have three elements.)

include/vapor/Field.h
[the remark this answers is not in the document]
shaomeng: Actually, that's a pretty good keyword to use! I'll add it in!

include/vapor/FlowRenderer.h
```diff
@@ -0,0 +1,145 @@
...
+    long                _cache_steadyNumOfSteps;
+    size_t              _cache_currentTS;
+    bool                _cache_periodic[3];
+    float               _cache_rake[6];
+    long                _cache_rakeNumOfSeeds[4];
+    std::string         _cache_rakeBiasVariable;
+    float               _cache_rakeBiasStrength;
+    int                 _cache_seedInjInterval;
+
+    // A few different modes to generate advection seeds:
```
StasJ: This should be an enum

include/vapor/FlowRenderer.h
```diff
@@ -0,0 +1,145 @@
...
+    //   0 - uniformly generate 
+    //   1 - randomly generate
+    //   2 - randomly generate with bias
+    //   3 - reading a list of seeds
+    long                _cache_seedGenMode;
+    std::string         _cache_seedInputFilename;
+
+    // A few different ways to integrate a flow line in steady mode:
```
StasJ: This should be an enum

include/vapor/Particle.h
```diff
@@ -0,0 +1,59 @@
...
+    float       value    = 0.0f;
+
+    // Constructors.
+    // This class complies with rule of zero.
+    Particle() = default;
+    Particle( const glm::vec3& loc, float t, float val = 0.0f );
+    Particle( const float* loc, float t, float val = 0.0f );
+    Particle( float x, float y, float z, float t, float val = 0.0f );
+
+    void  AttachProperty  (  float v );
```
StasJ: What is a "Property"?

include/vapor/Advection.h
```diff
@@ -0,0 +1,135 @@
...
+
+namespace flow
+{
+class Advection final
+{
+public:
+    enum class ADVECTION_METHOD
+    {
+        EULER =  0,
+        RK4   =  1      // Runge-Kutta 4th order
```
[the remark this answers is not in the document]
shaomeng: RK4 is a pretty common acronym for this method, so I don't think it needs spelling out everytime.

include/vapor/Particle.h
shaomeng: A property is a value that has a certain meaning associated with.

VDC-required-pkgs
```diff
@@ -0,0 +1,4 @@
+These packages are required to build VAPOR with only VDC enabled:
```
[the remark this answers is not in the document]
shaomeng: It's now fixed by adding a comment in the CMakeList file.

include/vapor/FlowRenderer.h
shaomeng: fixed.

include/vapor/FlowRenderer.h
shaomeng: fixed.

include/vapor/Particle.h
StasJ: Yes but what is it in this case? What is it used for?

shaomeng: Oh, I see why you're confused. It's a data field that could be used to hold any information that the owner of particles want the particle object to hold. This data field is optional and could be anything that in addition to the fixed data fields such as time/position, etc.

StasJ: Yes I understand that but my question is what is it used for.

shaomeng: Ok, the usage is better documented.

shaomeng:
@sgpearse  the updated code is ready for you to review.

include/vapor/FlowParams.h
```diff
@@ -0,0 +1,157 @@
...
+
+    std::vector<bool>   GetPeriodic() const;
+    void                SetPeriodic( const std::vector<bool>& );
+
+    /* 
+     * 6 values to represent a rake in this particular order: 
+     * xmin, xmax, ymin, ymax, zmin, zmax 
+     * If the rake wasn't set by users, it returns a vector containing nans.
+     */
+    std::vector<float>  GetRake() const;
```
StasJ: Everything else in Vapor uses the Params Box class to store regions. Please use that instead of the proprietary method here.

shaomeng: The old Params Box is hard wired to the params entry of the user-defined Geometry region. Rake needs to be wired to a different params entry, thus not be able to use ParamsBox.

PR #1867, opened 2019-10-10
Fixed #1842 - vapor unnecessarily reads full data domain when an ROI is selected

### LINKED ISSUE #1842: vapor unnecessarily reads full data domain when an ROI is selected
When the user selected a ROI with the geometry tab vapor still reads the entire data domain to calculate the variables range of values. For blocked data (e.g. VDC) vapor should only be reading the blocks that intersect the ROI.

To reproduce:
1. set the environment variable VAPOR_DEBUG and start vapor from the command line
2. Load the VDC version of the 1024^3 Taylor Green (TG) turbulence data
3. Create a volume renderer
4. Change the ROI to something small, less than 1/4 the domain size.
5. Change the currently selected variable to force the variable to be read from disk. Observe that the overloaded DataMgr::GetVariable() method is called right after GetRange(), and that the GetVariable() method used is NOT the version that specifies an ROI. I.e. the entire data volume is read

clyne:
Vapor was unnecessarily reading the entire data domain in two instances, despite a region of interest (ROI) having been selected. The first instance was whenever DataMgr::GetDataRange() was invoked to get a variable's minimum and maximum value. GetDataRange() is used in a number of places throughout the application, e.g. computing histograms.  The second instance was inside the DVR and isosurface volume renderer. Both were calling the overloaded DataMgr::GetVariable() method *without* specifying an ROI. 

The fix for the first instance was to overload DataMgr::GetDataRangeI() method with a version that accepts an ROI (min and max extents) as an argument. The volume rendering problem simply required call the correct version of DataMgr::GetVariable()

### DIFF
--- a/lib/render/VolumeIsoRenderer.cpp
+++ b/lib/render/VolumeIsoRenderer.cpp
@@ -37,7 +37,10 @@ VolumeIsoRenderer::VolumeIsoRenderer( const ParamsMgr*    pm,
     // An ugly fix but I don't think we have a mechanism for this
     if (_needToSetDefaultAlgorithm()) {
         VolumeParams *vp = (VolumeParams*)GetActiveParams();
-        Grid *grid = _dataMgr->GetVariable(vp->GetCurrentTimestep(), vp->GetVariableName(), vp->GetRefinementLevel(), vp->GetCompressionLevel());
+		vector <double> minExt, maxExt;
+		vp->GetBox()->GetExtents(minExt, maxExt);
+
+        Grid *grid = _dataMgr->GetVariable(vp->GetCurrentTimestep(), vp->GetVariableName(), vp->GetRefinementLevel(), vp->GetCompressionLevel(), minExt, maxExt);
         if (grid) {
             string algorithmName = _getDefaultAlgorithmForGrid(grid);
             vp->SetAlgorithm(algorithmName);
--- a/lib/render/VolumeRenderer.cpp
+++ b/lib/render/VolumeRenderer.cpp
@@ -67,7 +67,10 @@ VolumeRenderer::VolumeRenderer(
     
     if (_needToSetDefaultAlgorithm()) {
         VolumeParams *vp = (VolumeParams*)GetActiveParams();
-        Grid *grid = _dataMgr->GetVariable(vp->GetCurrentTimestep(), vp->GetVariableName(), vp->GetRefinementLevel(), vp->GetCompressionLevel());
+        vector <double> minExt, maxExt;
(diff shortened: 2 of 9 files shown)


lib/render/VolumeIsoRenderer.cpp
```diff
@@ -37,7 +37,10 @@ VolumeIsoRenderer::VolumeIsoRenderer( const ParamsMgr*    pm,
     // An ugly fix but I don't think we have a mechanism for this
     if (_needToSetDefaultAlgorithm()) {
         VolumeParams *vp = (VolumeParams*)GetActiveParams();
-        Grid *grid = _dataMgr->GetVariable(vp->GetCurrentTimestep(), vp->GetVariableName(), vp->GetRefinementLevel(), vp->GetCompressionLevel());
+		vector <double> minExt, maxExt;
+		vp->GetBox()->GetExtents(minExt, maxExt);
+
+        Grid *grid = _dataMgr->GetVariable(vp->GetCurrentTimestep(), vp->GetVariableName(), vp->GetRefinementLevel(), vp->GetCompressionLevel(), minExt, maxExt);
```
StasJ: The sole purpose of this grid is to get the grid type (Regular vs Curvilinear etc.). Is it possible to get this information without creating a new grid?

lib/render/VolumeRenderer.cpp
```diff
@@ -391,21 +394,31 @@ int VolumeRenderer::_initializeAlgorithm()
...
     VolumeParams *RP = (VolumeParams *)GetActiveParams();
+    vector <double> minExt, maxExt;
+    RP->GetBox()->GetExtents(minExt, maxExt);
+
     CheckCache(_cache.var, RP->GetVariableName());
     CheckCache(_cache.ts, RP->GetCurrentTimestep());
     CheckCache(_cache.refinement, RP->GetRefinementLevel());
     CheckCache(_cache.compression, RP->GetCompressionLevel());
+    CheckCache(_cache.minExt, minExt);
+    CheckCache(_cache.maxExt, maxExt);
```
StasJ: This will cause the data to be reloaded every time that the user changes the extents. Could we check to see if the new extents are still within the previously loaded grid and only reload the data if needed?

lib/render/VolumeIsoRenderer.cpp
clyne: There is not a method on the DataMgr to returns the grid type. One could of course be added if we thought it was worth the effort. In this particular instance i'm not sure it would buy us much as the data will have already been read and cached.

PR #1881, opened 2019-10-16
Rake region selector

StasJ:
Region selector edits the rake when the seed tab is open.


clyne:
I just wanted to add some comments to clarify this PR. I asked Stas to look at refactoring the manipulator to facilitate using it to support the Flow rake. Stas evaluated the code and concluded that the effort to do it right would be fairly substantial. However, as a proof of concept he was able to integrate the existing manipulator into the Flow GUI without much effort, albeit with a fairly ugly hack. I suggested that he submit a PR with this quick-and-dirty approach, and we'd table the refactoring that is need for a later time.

StasJ:
> I was about to review this PR, and then realize the flow branch is already merged into master. How about we modify this PR to be merging into master?

It turns out we have to merge it into `flow`

PR #1885, opened 2019-10-18
detect empty variable names

### LINKED ISSUE #1866: Flow crashes when selecting "no variable"
In the flow branch, if a user selects "no variable" in one of the velocity fields, VAPOR would crash.

shaomeng:
This PR makes FlowRenderer detect user selections of variable names, and return an error code in the case of empty variable name. 
It would fix #1866 .


lib/render/FlowRenderer.cpp
```diff
@@ -450,13 +454,26 @@ FlowRenderer::_updateFlowCacheAndStates( const FlowParams* params )
     // If names not the same, entire stream is out of date
     // Note: variable names are kept in VaporFields.
     std::vector<std::string> varnames = params->GetFieldVariableNames();
+    for( const auto& e : varnames )
```
clyne: I'm not sure that this code does what we want. It should not be an error for a user to select fewer than three variables. The missing variables should be treated as the "zero" variable: a variable that contains all zeros. A common use case is to do flow advection using only the horizontal components of velocity and ignore the vertical.

shaomeng: Oh really? That's interesting, and I I'll change the behavior of it.

shaomeng:
@clyne , this PR is modified so that the behavior is now the same as what you described. It's time to review the changes.

include/vapor/ConstantGrid.h
```diff
@@ -0,0 +1,85 @@
...
+#include "vapor/Grid.h"
+
+namespace VAPoR
+{
+
+class VDF_API ConstantGrid : public Grid
+{
+public:
+
+    // The constant value is specified via the constructor
```
clyne: The constant grid is also a dimensionless grid. I wonder if that is what we want, or if it would be better to specify dimensions, and coordinates? Maybe instead of being derived from Grid it should be derived from StructuredGrid (or RegularGrid). Something to consider for the future if we use ConstantGrid elsewhere.

lib/flow/VaporField.cpp
```diff
@@ -156,18 +174,28 @@ VaporField::GetVelocity( float time, const glm::vec3& pos, glm::vec3& velocity,
...
-            grid = _getAGrid( currentTS, varname );
-            VAssert( grid );
+            const auto& varname = VelocityNames[i];
+            if( varname.empty() )
+            {
+                grid = &cgrid;
+            }
+            else
+            {
+                grid = _getAGrid( currentTS, varname );
```
clyne: How about having _getAGrid return a constant grid as needed and eliminating all of these conditionals?

lib/flow/VaporField.cpp
```diff
@@ -124,12 +133,21 @@ VaporField::GetFirstStepVelocityIntersection( glm::vec3& minxyz, glm::vec3& maxx
...
-        grid->GetUserExtents( min[i], max[i] );
+        const auto& varname = VelocityNames[i];
+        if( varname.empty() )
+        {
+            min[i].resize( 3, std::numeric_limits<double>::min() );
+            max[i].resize( 3, std::numeric_limits<double>::max() );
+        }
+        else
+        {
+            grid = _getAGrid( 0, varname );
```
clyne: Here is an instance where if ConstantGrid were an actual grid with dimensions and coordinates this code wouldn't require special handling (conditionals)

lib/vdc/ConstantGrid.cpp
```diff
@@ -0,0 +1,104 @@
...
+{
+    return 0;
+}
+    
+size_t ConstantGrid::GetMaxVertexPerCell() const 
+{
+    return 0;
+}
+    
+VAPoR::Grid::ConstCoordItr ConstantGrid::ConstCoordBegin() const
```
clyne: Probably should throw an assert here. What will happen if someone tries to use the returned iterator?

lib/vdc/ConstantGrid.cpp
```diff
@@ -0,0 +1,104 @@
...
+}
+    
+VAPoR::Grid::ConstCoordItr ConstantGrid::ConstCoordEnd() const
+{
+    return VAPoR::Grid::ConstCoordItr();
+}
+    
+float ConstantGrid::GetValueNearestNeighbor( const std::vector <double> &coords) const
+{
+    return 0.0f;
```
clyne: Should this return _value instead of 0.0?

lib/vdc/ConstantGrid.cpp
```diff
@@ -0,0 +1,104 @@
...
+}
+    
+float ConstantGrid::GetValueLinear( const std::vector <double> &coords) const 
+{
+    return 0.0f;
```
clyne: Return _value?

include/vapor/ConstantGrid.h
shaomeng: I guess we can have create derivatives of `ConstantGrid` that represent `ConstantStructuredGrid` and `ConstantRegularGrid`, and each one of them will have more specific information held by them?

lib/vdc/ConstantGrid.cpp
shaomeng: Well, I moved these functions that are not supposed to be called by a user into a private block. In that way, it's guaranteed that nobody would call them.

lib/vdc/ConstantGrid.cpp
shaomeng: fixed.

lib/vdc/ConstantGrid.cpp
shaomeng: fixed.

lib/vdc/ConstantGrid.cpp
clyne: ok

lib/flow/VaporField.cpp
shaomeng: fixed.

lib/flow/VaporField.cpp
shaomeng: I changed the behavior of `GetUserExtents()` to return floating point mins and maxes. I guess if we decided to further implement things like `ConstantStructuredGrid`, then it would make sense for those derived classes to keep an actual user extent.

shaomeng:
Hey @clyne , I've updated the behavior of `ConstantGrid` to address some of your comments. You can see the updated header file to have an idea on what I've changed. 

If you also think we should go ahead and implement things like `ConstantRegularGrid`, then please let me know!

include/vapor/ConstantGrid.h
clyne: At some point we want to make ConstantGrid more like a real Grid class: have all of the public methods do reasonable things so that a ConstantGrid instance can easily be used anywhere we are using a Grid class. For now I'd say let's just move on since the current implementation is meeting our needs for flow visualization.

PR #1887, opened 2019-10-21
Issue1886 - Session files do not work if dataset path has spaces

### LINKED ISSUE #1886: Session files do not work if dataset path has spaces
### Works as expected:
1. Create a new session
2. Import a NetCDF file `/test/file.nc`
3. Save the session to `/session1.vs3`
4. Restart Vapor and open the session `/session1.vs3`

### Broken:
1. Create a new session
2. Import a NetCDF file `/test 2/file.nc`
3. Save the session to `/session2.vs3`
4. Restart Vapor and open the session `/session2.vs3`

Nothing will happen. The session will not load and there will be no error message.

StasJ:
The problem is that string vectors are space delimited in the params database/xml files. Since I'm guessing this is not something we can change due to backwards compatibility and potentially breaking things, I added escaping for spaces for string vectors in the params database.

Fix #1886

### DIFF
--- a/lib/params/XmlNode.cpp
+++ b/lib/params/XmlNode.cpp
@@ -27,6 +27,7 @@
 #include <algorithm>
 #include <expat.h>
 #include <vapor/XmlNode.h>
+#include <vapor/STLUtils.h>
 
 using namespace VAPoR;
 using namespace Wasp;
@@ -361,7 +362,7 @@ void XmlNode::SetElementStringVec(
 
 	string s;
 	for (int i=0; i<strvec.size(); i++) {
-		s.append(strvec[i]);
+		s.append(STLUtils::ReplaceAll(strvec[i], " ", "\\ "));
 		if (i<strvec.size()-1) s.append(" ");
 	}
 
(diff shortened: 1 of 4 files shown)


lib/params/XmlNode.cpp
```diff
@@ -412,8 +413,9 @@ const string &XmlNode::GetElementString(const string &tag) const {
 void XmlNode::GetElementStringVec(const string &tag, vector <string> &vec) const {
 	
 	string s = XmlNode::GetElementString(tag);
-
 	StrToWordVec(s, vec);
+    for (string &e : vec)
```
clyne: Does this code do what we want? If I have a two element string vector:

a[0] = "string"
a[1] = "string with spaces"

that I encode with SetElementStringVec

will I get be able to get it back as a *two* element vector with GetElementStringVec()? It looks like I'll get four element vector back

StasJ:
The previous behavior of StrToWordVec was to what you describe: you would
get a four element vector back. I changed StrToWordVec to only split at
non-escaped strings and when saving to params, spaces inside strings are
escaped then they are joined with non-escaped strings.

lib/params/XmlNode.cpp
StasJ: The previous behavior of `StrToWordVec` was to what you describe: you would
get a four element vector back. I changed `StrToWordVec` to only split at
non-escaped strings and when saving to params, spaces inside strings are
escaped then they are joined with non-escaped strings.

clyne: OK. I overlooked the changes to StrToWordVec

PR #1898, opened 2019-10-25
Fixed problem with hidden files

sgpearse:
TFColorWidget is throwing an error condition into MyBase when it tries to read hidden files in our share/palettes directory.


StasJ:
I think the error being logged is caused by a non-.tf3 file being read, for example a `.DS_Store`, not by a hidden file per se.

PR #1910, opened 2019-11-07
Fixed Issue 1891

clyne:
Added support to vdccreate for specifying stretched grids

### DIFF
--- a/apps/vdccreate/vdccreate.cpp
+++ b/apps/vdccreate/vdccreate.cpp
@@ -19,6 +19,10 @@ struct opt_t {
     std::vector <size_t> cratios;
 	string wname;
 	string xtype;
+	string xcoords;
+	string ycoords;
+	string zcoords;
+	string tcoords;
 	int numts;
 	int nthreads;
     std::vector <string> vars3d;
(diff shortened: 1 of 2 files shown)


apps/vdccreate/vdccreate.cpp
```diff
@@ -162,7 +186,15 @@ void	set_coord(
 	if (rc<0) exit(1);
 }
 
-void	set_coords(
+void	set_coord_stretched(
+	VDCNetCDF	&vdc,
+	string dimname, const vector <float> &coords
+) {
+	int rc = vdc.PutVar(0, dimname, -1, coords.data());
```
sgpearse: I see that VDC::PutVar has two forms, one that applies to a defined timestep (0 in this case) and another that applies to all timesteps in the VDC.  Is there's a rule that coordinate variables only need to be written to timestep 0?

clyne: If the coordinate variables are static (i.e. don't vary with time) they do not have a time index. vdccreate does not support time-varying coordinate variables.

PR #1921, opened 2019-11-15
Windows compatibility

StasJ:


shaomeng:
Can we explain what the bug is, and why this PR fixes the bug? This info is gonna make the review easier.

sgpearse:
> Can we explain what the bug is, and why this PR fixes the bug? This info is gonna make the review easier.

This is for fixing the windows build.  Stas may want to elaborate, but the main issues here are VS2015 not accepting dynamic arrays on the heap, a problem with std::pair, and tying up some errors and warnings that were postponed during the initial PR submissions of the Model Renderer and TFEditor.

shaomeng:
@sgpearse Thanks for the explanation. With this amount of changes, it does look like touching more than std::pare.

PR #1951, opened 2019-11-26
Fix #1832: Add color interpolation in RGB, LAB, and LCH spaces

### LINKED ISSUE #1832: Colormap adds extra colors when performing linear interpolation
Below is a colormap with three control points: red, green, and blue. In Vapor with linear interpolation, there will be yellow added between red and green, and cyan between green and blue. If I add a yellow control point between green and blue, it creates another green section between the yellow and blue.

StasJ:
Fix #1832

### DIFF
--- a/lib/params/ColorMap.cpp
+++ b/lib/params/ColorMap.cpp
@@ -425,6 +425,47 @@ ColorMap::Color ColorMap::color(float value) const
     return colorNormalized(nv);
 }
 
+static void lab2lch(const float lab[3], float lch[3])
+{
+    const float l = lab[0];
+    const float a = lab[1];
+    const float b = lab[2];
+    
+    const float c = sqrtf(a * a + b * b);
+    const float h = fmod((atan2f(b, a) / (2*M_PI) * 360 + 360), 360.f);
+    
+    lch[0] = l;
+    lch[1] = c;
+    lch[2] = h;
+}
+
+static void lch2lab(const float lch[3], float lab[3])
+{
+    const float l = lch[0];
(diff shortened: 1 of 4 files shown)


lib/params/ColorMap.cpp
```diff
@@ -425,6 +425,47 @@ ColorMap::Color ColorMap::color(float value) const
     return colorNormalized(nv);
 }
 
+static void lab2lch(const float lab[3], float lch[3])
```
clyne: Better to put these in an anoymous name space rather than make them static.

StasJ: Fixed

clyne:
@sgpearse the option to include or not include white space in a diverging color map seems like a hack, and adds complexity to the GUI. What is the motivation behind this option? Would it not be better addressed by simply providing color maps with different choices for the middle color?

sgpearse:
> @sgpearse the option to include or not include white space in a diverging color map seems like a hack, and adds complexity to the GUI. What is the motivation behind this option? Would it not be better addressed by simply providing color maps with different choices for the middle color?

Adding white space is a step taken in [Ken Moreland's ](http://www.kennethmoreland.com/color-maps/ColorMapsExpanded.pdf) diverging algorithm.  If the two colors at the endpoints of a diverging map are saturated enough, white space is usually desired.  Since it isn't desired 100% of the time, we have the option to turn it off.  I agree that adding the white control point to the color map would reduce complexity in the code base and the gui.

Pages 14 and 15 of Ken's paper go into details on this step.

PR #1958, opened 2019-12-03
Only allow the creation of 2D renderers if there is 2D data present and same for 3D

### LINKED ISSUE #1199: Creating TwoDData Renderers
Creating TwoDData Renderers is currently an option for our users when their data has no 2-D variables.  This is confusing them, and we should disable the option to make 2-D exclusive renderers when it isn't possible.

The opposite should also take place when their data is exclusively 2D, and has no 3D variables.

Related to #1195

### LINKED ISSUE #1619: Error thrown on volume enabled without 3D variable
1. Open a 2D only dataset
2. Add/enable a volume renderer

We get the following error:
```
Error performing paint event
Failed to read variable "" at time step (0), and
refinement level (-1) and level-of-detail (0)
Invalid variable reference : 
```

We should have a cleaner error message or just grey out the option to select a 3D renderer if there are no 3D variables present.

### LINKED ISSUE #1650: Mapping Frame is glitched in 2D renderers if there are no 2D variables
1. Open a dataset without a 2D variable
2. Create a 2D renderer (contour, 2DData)

tf

StasJ:
Fix #1199 
Fix #1619 
Fix #1650

### DIFF
--- a/apps/vaporgui/RenderHolder.cpp
+++ b/apps/vaporgui/RenderHolder.cpp
@@ -314,13 +314,6 @@ void RenderHolder::_showIntelDriverWarning(const string &rendererType)
 void RenderHolder::_showNewRendererDialog() {
 	ParamsMgr *paramsMgr = _controlExec->GetParamsMgr();
 	vector <string> dataSetNames = paramsMgr->GetDataMgrNames();
-    
-    DataMgr *dm = _controlExec->GetDataStatus()->GetDataMgr(dataSetNames[0]);
-    auto vars = dm->GetDataVarNames();
-    int d2 = dm->GetDataVarNames(2).size();
-    int d3 = dm->GetDataVarNames(3).size();
-    printf("3D Vars: %i\n", d3);
-    printf("2D Vars: %i\n", d2);
 	
 	_initializeNewRendererDialog(dataSetNames);
 	if (_newRendererDialog->exec() != QDialog::Accepted) {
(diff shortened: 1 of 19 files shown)


apps/vaporgui/RenderHolder.cpp
```diff
@@ -270,6 +314,13 @@ void RenderHolder::_showIntelDriverWarning(const string &rendererType)
 void RenderHolder::_showNewRendererDialog() {
 	ParamsMgr *paramsMgr = _controlExec->GetParamsMgr();
 	vector <string> dataSetNames = paramsMgr->GetDataMgrNames();
+    
+    DataMgr *dm = _controlExec->GetDataStatus()->GetDataMgr(dataSetNames[0]);
+    auto vars = dm->GetDataVarNames();
+    int d2 = dm->GetDataVarNames(2).size();
+    int d3 = dm->GetDataVarNames(3).size();
+    printf("3D Vars: %i\n", d3);
```
clyne: Remove diagnostics

clyne:
Does this PR also fix #1650

StasJ:
@clyne Yes, it does fix #1650 and I've added that to the PR description

I see what you mean. I fixed the title.

Since `DataMgr::GetDataVarNames(int dim)` includes derived variables then it will work as expected. A dataset would need to have a 2D grid to output the variable to, however, which I assume a 3D-only dataset would not have.

apps/vaporgui/RenderHolder.cpp
StasJ: Fixed

PR #1963, opened 2019-12-04
Fix to reduce windows 3rd party file size

sgpearse:
My 3rd party library archive contained redundant .dll files.  This just points CMake to a new location for .dll's.


sgpearse:
I'll get copies up onto glade tomorrow.  I know there are concerns with storing these kinds of files on Google Drive, but right now it's the only way for us to distribute the files to users.

PR #1965, opened 2019-12-05
Fix #1748: Transforms: First loaded dataset gets ignored

### LINKED ISSUE #1748: Transforms: First loaded dataset gets ignored
To reproduce:

1) Load any WRF VDC dataset
2) Import one if its WRF source files
3) In the Navigation->Viewpoint tab, scale the VDC dataset by 30 on Z
4) Move the scene, and the scale factor will revert to its previous value
5) Scale the imported dataset, and the transform will be honored

StasJ:
Fix #1748 

There was some code in the annotation renderer that was breaking the dataset scaling

### DIFF
--- a/lib/render/AnnotationRenderer.cpp
+++ b/lib/render/AnnotationRenderer.cpp
@@ -514,14 +514,6 @@ void AnnotationRenderer::InScenePaint(size_t ts)
 
 	vector<string> names = m_paramsMgr->GetDataMgrNames();
     Transform* t = vpParams->GetTransform(names[0]);
-    if (names.size()) {
-        Transform* tmp;
-        for (int i=0; i<names.size(); i++) {
-            tmp = vpParams->GetTransform(names[i]);
-            if ( tmp->GetScales()[Z] < t->GetScales()[Z] )
-                *t = *tmp;
-        }
-    }
 	applyTransform(t);
 
 	double mvMatrix[16];
--- a/lib/params/CMakeLists.txt
+++ b/lib/params/CMakeLists.txt
@@ -61,6 +61,7 @@ set (HEADERS
 	${PROJECT_SOURCE_DIR}/include/vapor/VolumeParams.h
 	${PROJECT_SOURCE_DIR}/include/vapor/VolumeIsoParams.h
 	${PROJECT_SOURCE_DIR}/include/vapor/ModelParams.h
+	${PROJECT_SOURCE_DIR}/include/vapor/Transform.h
 )
 
 add_library (params SHARED ${SRC} ${HEADERS})


lib/render/AnnotationRenderer.cpp
```diff
@@ -514,14 +514,6 @@ void AnnotationRenderer::InScenePaint(size_t ts)
...
 	vector<string> names = m_paramsMgr->GetDataMgrNames();
     Transform* t = vpParams->GetTransform(names[0]);
-    if (names.size()) {
-        Transform* tmp;
-        for (int i=0; i<names.size(); i++) {
-            tmp = vpParams->GetTransform(names[i]);
-            if ( tmp->GetScales()[Z] < t->GetScales()[Z] )
-                *t = *tmp;
-        }
-    }
```
sgpearse: I believe this will reintroduce issue #1214 

In retrospect, I should have commented this code though.  It's taking the smallest z scale factor from all datasets, and using that is the default.  If a ROMS default scale factor is 1000, this will blow up the scene if it's loaded alongside of a WRF dataset.

StasJ: I don't quite understand that bug. It says "[if auto scaling is enabled ... step 1: make sure auto scaling is disabled]"

How would this fix that bug? The annotation renderer shouldn't fix the renderers' scaling factors?

sgpearse: What this code is doing is 

1) Gather the default scale factors for all of the data managers
2) Select the smallest scale factor, and apply that to the ViewpointParams

Say we loaded an ocean model with a default scale factor of 1000, and an atmospheric model with a scale factor of 50. 

When we randomly take the scale factor associated with names[0], we could be applying the ocean model 1000x scale.  This makes the atmospheric domain too big, and unusable.  

Therefore we look at all scale factors, and use the smallest one for the ViewpointParams.

clyne:
@sgpearse you approved the PR, but the conversation has not been marked "resolved". Is it resolved?

lib/render/AnnotationRenderer.cpp
clyne: @sgpearse you have approved this PR, but it's not clear that your concern regarding a regression has been addressed: "I believe this will reintroduce issue #1214". Has it been addressed?

sgpearse: I went ahead tested this branch to see if #1214 had regressed, but I wasn't able to reproduce it.  Since there's no regression, I approved the PR.

sgpearse:
@clyne I went ahead tested this branch to see if #1214 had regressed, but I wasn't able to reproduce it. Since there's no regression, I approved the PR.

PR #1977, opened 2019-12-10
TF editor allow users to specify range bounds outside of current time step

### LINKED ISSUE #1728: TF editor can't specify range bounds outside of current time step
- [ ] Did VAPOR crash?
- [x] Did you get wrong results?

**Impact**

### To Reproduce
1. Load all of the MPAS data in vapor/data/Source/MPAS/FalkoJudt/30km
2. Enable 2D renderer for the current time step
3. Try to set the max value of the TF editor to something larger than ~79. It's not possible
4. Change the time step to the last time step. Note that the range of values is much larger for the later time steps in the data set. Now it's possible to change the max value. It should ALWAYS be possible to change the min/max value.

### LINKED ISSUE #1923: allow range sliders to operate outside of limits where reasonable
The range sliders used to set a min and max value should, in some instances, allow the user to specify values outside of the range. Per @leighorf:

Please make it possible to set the max / min values for data values go beyond the max / min in the data for the frame that is loaded (I'm talking about the Appearance tab/transfer function). This allows me to set colormaps at "standard values" even for data that does not span the full range. I know VAPOR can already do this, as if you have a sequence of times, go to one where the data really does span all the values, and then go back to the earlier data set, it displays as intended.

### LINKED ISSUE #1964: Allow for user-defined colormap range
I am trying to fix the colormap range of a volume renderer to values that are the min and max of a 4D dataset. However, I seem to only be able to set the range to the min and max of the **current** frame. When I put in values outside of that range, it gets overwritten by the minimum of the current frame (e.g. if the min of the dataset is -4.0 m/s but the min of the current frame is -2.0 m/s, if I type in -4.0 to the transfer function minimum value box will revert to -2.0).

This also makes it difficult if you have two slices (e.g. XY and YZ) that you want to set to a common color range.

It would be nice to simply be able to set the min and max value to any number and then have a button to set the range to the range of the current frame if desired.

**Impact**
- [ ] High - User productivity significantly improved
- [x] Medium - User productivity partially improved
- [ ] Low - User productivity trivially improved

### LINKED ISSUE #1976: Floating sidebar broken in Qt5
out

StasJ:
Fix #1964
Fix #1728
Fix #1923

Also
Fix #1976

### DIFF
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -301,6 +301,7 @@ MainForm::MainForm(
     _tabDockWindow = new QDockWidget(this );
     addDockWidget(Qt::LeftDockWidgetArea, _tabDockWindow );
     _tabDockWindow->setAllowedAreas(Qt::LeftDockWidgetArea | Qt::RightDockWidgetArea);
+    _tabDockWindow->setFeatures(QDockWidget::NoDockWidgetFeatures);
 
 	// Register additional params with the ParamsMgr
 	//
--- a/apps/vaporgui/QRangeSlider.h
+++ b/apps/vaporgui/QRangeSlider.h
@@ -35,6 +35,8 @@ private:
     
     int _position[2];
     int _value[2];
+    bool _isOutOfBounds[2] = {false};
+    float _outOfBoundValue[2];
     int _grabbedControl = -1;
     int _lastSelectedControl = 0;
     
@@ -49,4 +51,5 @@ private:
     bool doesGrooveContainPixel(const QPoint &pixel) const;
     bool isSliderDown(int i) const;
     void swapSliders();
+    void emitValueChanged(bool intermediate=false);
 };
--- a/apps/vaporgui/QRangeSliderTextCombo.cpp
+++ b/apps/vaporgui/QRangeSliderTextCombo.cpp
@@ -67,7 +67,7 @@ void QRangeSliderTextCombo::AllowCustomRange()
     
     QAction *newMinAction     = new QAction("Set as new min", this);
     QAction *newMaxAction     = new QAction("Set as new max", this);
-    QAction *resetRangeAction = new QAction("Reset to default", this);
+    QAction *resetRangeAction = new QAction("Reset range to default", this);
     QObject::connect(newMinAction,     &QAction::triggered, this, &QRangeSliderTextCombo::makeLeftValueNewMin);
     QObject::connect(newMaxAction,     &QAction::triggered, this, &QRangeSliderTextCombo::makeRightValueNewMax);
     QObject::connect(resetRangeAction, &QAction::triggered, this, &QRangeSliderTextCombo::RangeDefaultRequested);
--- a/apps/vaporgui/QRangeSliderTextCombo.cpp
+++ b/apps/vaporgui/QRangeSliderTextCombo.cpp
@@ -59,7 +59,6 @@ void QRangeSliderTextCombo::SetValue(float left, float right)
 
 void QRangeSliderTextCombo::AllowCustomRange()
 {
-    assert(!_allowCustomRange);
     if (_allowCustomRange)
         return;
     
--- a/apps/vaporgui/QRangeSliderTextCombo.cpp
+++ b/apps/vaporgui/QRangeSliderTextCombo.cpp
@@ -3,6 +3,7 @@
 #include <QDoubleValidator>
 #include <vapor/VAssert.h>
 #include <cfloat>
+#include <QAction>
 
 QRangeSliderTextCombo::QRangeSliderTextCombo()
 {
@@ -31,6 +32,12 @@ void QRangeSliderTextCombo::SetRange(float min, float max)
     VAssert(_max >= _min);
     _min = min;
     _max = max;
+    
+    if (_allowCustomRange) {
+        min = -FLT_MAX;
+        max = FLT_MAX;
+    }
+    
     setValidator(_leftText, new QDoubleValidator(min, max, 100));
     setValidator(_rightText, new QDoubleValidator(min, max, 100));
     SetValue(_left, _right);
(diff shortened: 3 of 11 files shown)


apps/vaporgui/MainForm.cpp
```diff
@@ -301,6 +301,7 @@ MainForm::MainForm(
     _tabDockWindow = new QDockWidget(this );
     addDockWidget(Qt::LeftDockWidgetArea, _tabDockWindow );
     _tabDockWindow->setAllowedAreas(Qt::LeftDockWidgetArea | Qt::RightDockWidgetArea);
+    _tabDockWindow->setFeatures(QDockWidget::NoDockWidgetFeatures);
```
clyne: What does do?

apps/vaporgui/QRangeSlider.h
```diff
@@ -35,6 +35,8 @@ class QRangeSlider : public QSlider {
     
     int _position[2];
     int _value[2];
+    bool _isOutOfBounds[2] = {false};
```
clyne: Can you initialize an array of size two with a single elements?

apps/vaporgui/QRangeSliderTextCombo.cpp
```diff
@@ -31,21 +32,54 @@ void QRangeSliderTextCombo::SetRange(float min, float max)
...
+    }
     _left = left;
     _right = right;
     setTextboxes(left, right);
     _slider->SetValue((left-_min)/(_max-_min), (right-_min)/(_max-_min));
 }
 
+void QRangeSliderTextCombo::AllowCustomRange()
+{
+    assert(!_allowCustomRange);
```
clyne: Do you really want this assert? It would appear that calling AllowCustomRange() more than once will fail the assert. Is that what you want?

If an assert is needed use VAssert

apps/vaporgui/QRangeSliderTextCombo.h
```diff
@@ -3,25 +3,28 @@
...
     float _min, _max;
     float _left, _right;
+    bool _allowCustomRange = false;
     
 public:
     QRangeSliderTextCombo();
     
     void SetRange(float min, float max);
     void SetValue(float left, float right);
+    void AllowCustomRange();
```
clyne: Doxygen documentation please.

apps/vaporgui/TFHistogramWidget.cpp
```diff
@@ -83,6 +83,9 @@ void TFHistogramMap::paintEvent(QPainter &p)
         stride *= 2;
     startBin -= startBin % stride;
     
+    if (_histo.getRange() < FLT_EPSILON || (mapRange[1]-mapRange[0])/_histo.getRange() > 10000)
```
clyne: Comment what this does please

include/vapor/MapperFunction.h
```diff
@@ -153,6 +153,12 @@ class PARAMS_API MapperFunction : public ParamsBase
  //! Obtain min and max mapping (histo) values
  //! \return minimum and maximum as a 2-vector of doubles
  vector<double> getMinMaxMapValue() const;
+
```
clyne: Needs doxygen documentation

lib/params/MapperFunction.cpp
```diff
@@ -372,6 +374,30 @@ vector<double> MapperFunction::getMinMaxMapValue() const {
 }
 
 
+vector<double> MapperFunction::GetCustomMappingSliderRange() const
+{
+    vector<double> defaultv(2, 0);
+    assert(IsUsingCustomMappingSliderRange());
```
clyne: Should this really be an assert? Use VAssert if so.

lib/params/MapperFunction.cpp
```diff
@@ -372,6 +374,30 @@ vector<double> MapperFunction::getMinMaxMapValue() const {
...
+}
+
+bool MapperFunction::IsUsingCustomMappingSliderRange() const
+{
+    return GetValueLong(IsUsingCustomMappingSliderRangeTag, false);
+}
+
+void MapperFunction::SetCustomMappingSliderRange(const vector<double> &range)
+{
+    assert(IsUsingCustomMappingSliderRange());
```
clyne: VAssert, not assert

clyne:
@stasj does the implementation allow the user to reset the slide to the new range as discussed?

StasJ:
Yes, by right clicking the text input.

apps/vaporgui/MainForm.cpp
StasJ: It fixes  #1976 by disallowing floating the left bar.

apps/vaporgui/QRangeSlider.h
StasJ: Yes

apps/vaporgui/QRangeSliderTextCombo.cpp
StasJ: Yes. If it is done, nothing will break so we don't want to crash the release application but generally if it is called multiple times there is a mistake someone made so it lets them know.

StasJ: Fixed

apps/vaporgui/QRangeSliderTextCombo.h
StasJ: Done

apps/vaporgui/TFHistogramWidget.cpp
StasJ: Done

include/vapor/MapperFunction.h
StasJ: Done

lib/params/MapperFunction.cpp
StasJ: Done

lib/params/MapperFunction.cpp
StasJ: Done

PR #2006, opened 2019-12-17
Issue1993

### LINKED ISSUE #1993: Flow GUI inconsistency
The newly introduced rake selector is able to stay in pretty inconsistent states by typing in an obvious out-of-bound value.
    -- Note 1: the inconsistency isn't corrected even after hitting the "Enter" key
    -- Note 2: the old flow GUI doesn't exhibit the inconsistency.

Screenshot from 2019-12-12 16-11-52

StasJ:
Fix #1993

### DIFF
--- a/apps/vaporgui/VDoubleValidator.h
+++ b/apps/vaporgui/VDoubleValidator.h
@@ -7,6 +7,10 @@
 //! They behave the same except QDoubleValidator does not accept values outside
 //! the range while VDoubleValidator does not allow the user to input
 //! values outside of the range. This solves some quirks with QLineEdit
+//! With QDoubleValidator, the user can type a value outside the range in the text
+//! box but the text box will say the value is invalid and it will stop working
+//! VDoubleValidator, will not allow the user to type in a value outside the range
+//! omitting the above problem.
 
 class VDoubleValidator : public QDoubleValidator {
 public:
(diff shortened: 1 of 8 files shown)


apps/vaporgui/VDoubleValidator.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <QDoubleValidator>
+
+//! \class VDoubleValidator
+//! Use as a direct replacement of QDoubleValidator.
```
clyne: Please document how VDoubleValidator differs from QDoubleValidator

StasJ: done

clyne: Is there a typo in here:

QDoubleValidator does not accept values outside
//! the range while VDoubleValidator does not allow the user to input
//! values outside of the range

It appears that they do the same thing

StasJ:
> Unfortunately this does not fix #1993. I'm ok with accepting this PR, but the issue remains.
> 
> The QDoubleValidator prohibits its parent from emitting signals when an invalid value is entered. This just leaves the invalid value unchanged until our next round of Update() calls.
> 
> I recommend performing our own validation within the leftTextChanged() and rightTextChanged() functions, and doing away with QValidators. Thoughts?

@pearse, have you tried reproducing #1993?

apps/vaporgui/VDoubleValidator.h
StasJ: QDoubleValidator allows users to input values outside the range.

sgpearse:
Turns out I had a local branch named issue1993, so my test was failing on that branch, not this one.  Seems to work now.

PR #2032, opened 2019-12-24
Use TF lookup for contour and wireframe renderers

### LINKED ISSUE #2025: Wireframe and contour renderer regenerates primitives for colormap change
This causes poor performance when using the transfer function editor. Performance would be greatly improved if we just stored the data value and did a TF lookup in the shader like the other renderers.

out

### LINKED ISSUE #2027: Contour renderer TF shows opacity despite the renderer not supporting opacity

StasJ:
Fix #2025 
Fix #2027

### DIFF
--- a/include/vapor/ContourRenderer.h
+++ b/include/vapor/ContourRenderer.h
@@ -34,6 +34,7 @@
 #include <vapor/Renderer.h>
 #include <vapor/ContourParams.h>
 #include <vapor/ShaderProgram.h>
+#include <vapor/Texture.h>
 
 namespace VAPoR {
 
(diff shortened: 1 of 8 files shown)


StasJ:
@clyne you can just delete one of the color control points

clyne:
> @clyne you can just delete one of the color control points

Hmm. I find that less intuitive and a little bit cumbersome. I'd like to hear what @shaomeng thinks.

shaomeng:
I'm actually in favor of keep it simple, i.e., remove a "Constant Color" knob. However, I'd advocate not allowing deleting the last control point, so that it imprints in the user's mind that control points control the colors! (Where as currently you still get a white color without control points.)

StasJ:
@shaomeng I agree that users should not be able to delete the last control point (although it does not break anything) and I've fixed it so you no longer can.

Issue #2040, opened 2019-12-27
Params buggy code


StasJ:
Some params classes have the following code, some don't:

```
// If node isn't tagged correctly we correct the tag and reinitialize
// from scratch;
//
if (node->GetTag() != BarbParams::GetClassType()) {
	node->SetTag(BarbParams::GetClassType());
	_init();
}
```

It creates bugs if you try to inherit from a class that has this. What is it for? Can we delete this.

PR #2041, opened 2019-12-27
Issue2038 IsoSurface "Color by Second Variable" undo broken

### LINKED ISSUE #2038: IsoSurface "Color by Second Variable" undo broken
1. Enable "Color by Second Variable" for an isosurface
2. Do some action or some number of actions
3. Undo the last action
4. "Color by Second Variable" will be disabled

StasJ:
Fix #2038 

See #2040

### DIFF
--- a/lib/params/VolumeIsoParams.cpp
+++ b/lib/params/VolumeIsoParams.cpp
@@ -22,17 +22,7 @@ VolumeIsoParams::VolumeIsoParams(
 
 VolumeIsoParams::VolumeIsoParams(
 	DataMgr *dataMgr, ParamsBase::StateSave *ssave, XmlNode *node
-) : VolumeParams(dataMgr, ssave, node) {
-	SetDiagMsg("VolumeIsoParams::VolumeIsoParams() this=%p", this);
-
-	// If node isn't tagged correctly we correct the tag and reinitialize
-	// from scratch;
-	//
-	if (node->GetTag() != VolumeIsoParams::GetClassType()) {
-		node->SetTag(VolumeIsoParams::GetClassType());
-		_init();
-	}
-}
+) : VolumeParams(dataMgr, ssave, node) {}
 
 
 VolumeIsoParams::~VolumeIsoParams() {
--- a/lib/params/AnnotationParams.cpp
+++ b/lib/params/AnnotationParams.cpp
@@ -87,14 +87,6 @@ AnnotationParams::AnnotationParams(
 	ParamsBase::StateSave *ssave, XmlNode *node
 ) : ParamsBase(ssave, node) 
 {
-	// If node isn't tagged correctly we correct the tag and reinitialize
-	// from scratch;
-	//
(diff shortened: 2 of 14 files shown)


lib/params/VolumeIsoParams.cpp
```diff
@@ -22,17 +22,7 @@ VolumeIsoParams::VolumeIsoParams(
 
 VolumeIsoParams::VolumeIsoParams(
 	DataMgr *dataMgr, ParamsBase::StateSave *ssave, XmlNode *node
-) : VolumeParams(dataMgr, ssave, node) {
-	SetDiagMsg("VolumeIsoParams::VolumeIsoParams() this=%p", this);
-
-	// If node isn't tagged correctly we correct the tag and reinitialize
-	// from scratch;
-	//
-	if (node->GetTag() != VolumeIsoParams::GetClassType()) {
```
clyne: if this branch is evaluating to true something deeper is probably wrong. What is the the value node->GetTab()?

StasJ: The problem is that this “failsafe” code that exists in some Params classes breaks if you use inheritance and VolumeIsoParams inherits from VolumeParams. It calls the parent constructor which says I’m not a “VolumeParams“ so let’s delete everything and reinitialize as VolumeParams. Then VolumeIsoParams says I’m not a “VolumeIsoParams“ so let’s delete everything and reinitialize as VolumeIsoParams.

StasJ: The problem is that this “failsafe” code that exists in some Params classes breaks if you use inheritance and VolumeIsoParams inherits from VolumeParams. It calls the parent constructor which says I’m not a “VolumeParams“ so let’s delete everything and reinitialize as VolumeParams. Then VolumeIsoParams says I’m not a “VolumeIsoParams“ so let’s delete everything and reinitialize as VolumeIsoParams.

clyne: The failsafe code exists for backwards compatibility with old session files. If a new renderer is added to vapor and a user tries to use a session file generated from a previous version what will the behavior be? I don't think the application will work as expected. Is there another way to work around this problem while preserving backwards compatibility with older session files?

StasJ: In theory, yes I see what you mean, but first, we don't have backwards compatibility with old session files, and second, this code would not improve the situation in any way: This code only checks that a node's tag is the same as the class name, however the node would never be created from a session file if it does not have the same tag, Vapor would just assume it is a different renderer.

StasJ: However if you prefer, I can keep it and work around it.

clyne: OK. Agreed. Let's remove the failsafe from all of the code since it's liable to problematic elsewhere in the future, shall we?

StasJ: Sounds good. Done. (Fix #2040)

PR #2045, opened 2019-12-30
2d flow

sgpearse:
Support for 2D flow


shaomeng:
> I opened the VDC file WRF/2D and vapor crashed when I created a flow renderer

@sgpearse  could you take a look at `VSliderEdit.cpp` ?
It appears the the GUI wants to set a range to the slider where min is greater than max... Is it a data specific issue?

sgpearse:
@shaomeng I'm seeing the problem and I think this requires some discussion about our pathline length parameter (aka unsteady flow length).

The maximum length is currently the number of timesteps-1 (numTS-1).  The minimum value was changed to be 1 in the commit noted below.

In the case that there is only one time step, the max (numTS-1) is 0, and the minimum is 1.  When max is less than min, we see the this assertion.

I believe that the renderer should accept a value of 0 for pathline length, and just display a point instead of a line.  We can also special-case this widget for when the data only has one time step.  We should probably have a discussion about the tradeoffs between having a minimum value of 0 vs 1.

[3e9e398be2196e22ba0818c51631aa1876cd0036](https://github.com/NCAR/VAPOR/pull/2045/commits/3e9e398be2196e22ba0818c51631aa1876cd0036)

shaomeng:
> I opened the VDC file WRF/2D and vapor crashed when I created a flow renderer

I believe this crash is resolved. Can you give it another try?

StasJ:
> > I opened the VDC file WRF/2D and vapor crashed when I created a flow renderer
> 
> I believe this crash is resolved. Can you give it another try?

Vapor no longer crashes but it will freeze permanently.

out

include/vapor/GrownGrid.h
```diff
@@ -0,0 +1,84 @@
...
+class VDF_API GrownGrid final: public Grid 
+{
+public:
+
+    // Constructor: pass in the 2D grid to be managed, the data manager that will
+    // later be used to destroy the grid, and the default Z value to be used.
+    // Note: the passed in 2D grid should be queried with the "lock" option on,
+    //       and GrownGrid will take ownership of the 2D grid afterwards.
+    //       I.e., GrownGrid will be responsible deleting the underlying 2D grid.
+    GrownGrid( const VAPoR::Grid* gp, VAPoR::DataMgr* mp, float z );
```
clyne: Since this is a stopgap fix until we change the behavior of the Grid class how about moving the source files to the vapor application directory to prevent its use from propagating? Also, it would be good to document the problem that it is addressing.

lib/flow/VaporField.cpp
```diff
@@ -48,26 +46,22 @@ VaporField::InsideVolumeVelocity( float time, const glm::vec3& pos ) const
...
         {
-            if( !v.empty() )
-            {
-                grid = _getAGrid( floor, v );
-                VAssert( grid );
-                if( !grid->InsideGrid( coords ) )
-                    return false;
-            }
+            grid = _getAGrid( floor, v );
+            VAssert( grid );
```
clyne: Should you be using VAssert(grid) here? It looks like _getAGrid can return NULL under conditions that are NOT programming errors. E.g. the variable is not available at the requested timestep

lib/flow/VaporField.cpp
```diff
@@ -48,26 +46,22 @@ VaporField::InsideVolumeVelocity( float time, const glm::vec3& pos ) const
...
-                if( !v.empty() )
-                {
-                    grid = _getAGrid( floor + 1, v );
-                    VAssert( grid );
-                    if( !grid->InsideGrid( coords ) )
-                        return false;
-                }
+            {
+                grid = _getAGrid( floor + 1, v );
+                VAssert( grid );
```
clyne: Ditto VAssert(grid) here and elsewhere

lib/render/Renderer.cpp
```diff
@@ -715,25 +715,26 @@ void Renderer::GetClippingPlanes( float planes[24] ) const
...
+    x1Plane[3] = float(maxExts[0]);
+    y0Plane[3] = float(-minExts[1]);
+    y1Plane[3] = float(maxExts[1]);
+    if (minExts.size() == 3 )   // Fill normal Z extents
     {
         z0Plane[3] = float(-minExts[2]);
         z1Plane[3] = float(maxExts[2]);
     }
+    else    // Fill a thin layer around DefaultZ
+    {
```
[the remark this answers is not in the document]
shaomeng: The function here in the Renderer class returns 3D region anyway. Before my change, it returns a range between -0 and 0 in Z. After my change, it returns a thin layer around DefaultZ. This change has no impact on if individual renderers want to disable clipping along Z. Does it make sense?

apps/vaporgui/FlowSubtabs.cpp
```diff
@@ -23,18 +24,21 @@ namespace {
     const std::string GRIDDED_STRING   = "Gridded";
     const std::string LIST_STRING      = "List of seeds";
     const std::string RANDOM_STRING    = "Random";
+    const std::string RANDOM_BIAS_STR  = "Random w/ Bias";
```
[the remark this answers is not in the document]
shaomeng: There is a usability issue with that direction. We can talk about that f2f if needed.

lib/params/FlowParams.cpp
```diff
@@ -86,7 +87,8 @@ FlowParams::GetVelocityMultiplier() const
 void 
 FlowParams::SetVelocityMultiplier( double coeff )
 {
-    SetValueDouble( _velocityMultiplierTag, "Field Scale Factor", coeff );
+    if( float(coeff) != float( GetVelocityMultiplier() ) )
```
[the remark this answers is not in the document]
shaomeng: Fixed by commit 7a821cf

include/vapor/GrownGrid.h
shaomeng: Fixed in commit b6525ce

lib/flow/VaporField.cpp
shaomeng: Did you mean this place should not use an assert?

lib/render/Renderer.cpp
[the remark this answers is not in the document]
clyne: This still seems like a hack to me. I think we need a more robust way to address this issue. Why aren't you simply making use of the 'halo' argument to Renderer::EnableClipToBox(). I believe that is it's intended purpose. We should probably discuss f2f with @StasJ

lib/flow/VaporField.cpp
clyne: asserts should only be used for programing errors. It would appear that you are using an assert here to address something that is not a programming error: a data read failure, which should have proper error handing.

lib/render/Renderer.cpp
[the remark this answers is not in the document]
clyne: @shaomeng you should be using the Renderer::EnableClipToBox() method for clipping. See other renderers for an example.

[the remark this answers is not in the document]
shaomeng: Committed the requested changes in e7240ac. Will need to test it when the GeometryWidget is fixed. Don't mark it resolved yet.

lib/flow/VaporField.cpp
shaomeng: Fixed in commit c1ec340

lib/flow/VaporField.cpp
shaomeng: Fixed in commit c1ec340

lib/render/Renderer.cpp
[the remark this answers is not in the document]
shaomeng: Finally fixed via commit 2f6b4ed

shaomeng:
@clyne @StasJ  This PR is ready to be reviewed again!

lib/flow/VaporField.cpp
```diff
@@ -130,7 +136,6 @@ VaporField::GetFirstStepVelocityIntersection( glm::vec3& minxyz, glm::vec3& maxx
     for( int i = 0; i < 3; i++ )
     {
         grid = _getAGrid( 0, VelocityNames[i] );
-        VAssert( grid );
```
clyne: It looks like we'll get a seg. fault here if a variable doesn't exist at time step 0.

share/shaders/FlowLine.frag
```diff
@@ -8,21 +8,21 @@ in vec3             vertModel;
...
 uniform bool        singleColor;
-uniform vec4        clipPlanes[6];     // clipping planes in model coordinates
+//uniform vec4        clipPlanes[6];     // clipping planes in model coordinates
 
 
 void main(void)
 {
     // First, let's decide if we want to discard this fragment
-    vec4 positionModel = vec4( vertModel, 1.0 );
+    /*vec4 positionModel = vec4( vertModel, 1.0 );
```
clyne: It's really hard to see that this code is commented out with the old style C comments (/* */). I believe you can use #ifdefs. Check with @stasj. In any event can you add a comment as to why the code is commented out (and not simply deleted)?

shaomeng: It was old code, and has already been removed. I'm not sure #ifdef stuff works in the shader.

lib/flow/VaporField.cpp
shaomeng: Fixed in commit e58098c .

shaomeng:
@clyne @StasJ This PR is ready to be reviewed again!

clyne:
Will do. Sam, just an FYI that if you click the rotating arrows next to the reviewer’s name GitHub will notify them that the PR is ready to be reviewed again.

> On Jan 3, 2020, at 3:01 PM, Samuel Li <notifications@github.com> wrote:
> 
> @clyne @StasJ This PR is ready to be reviewed again!
> 
> —
> 

John Clyne
clyne@ucar.edu

shaomeng:
> I cannot get anything to render with WRF/2D. I tried messing with every setting, increasing the number of seeds, different seeding methods and I made sure all of the referenced variables were valid.

@StasJ if you increase the velocity multiplier from 1 to 100K, then you'll see something. This particular data set uses different units.

StasJ:
> > I cannot get anything to render with WRF/2D. I tried messing with every setting, increasing the number of seeds, different seeding methods and I made sure all of the referenced variables were valid.
> 
> @StasJ if you increase the velocity multiplier from 1 to 100K, then you'll see something. This particular data set uses different units.

Do we expect the user to have to scale by 100,000 to see anything at all? I mean thats 100 times what the limit was before you just increased it. Also, I tried re-testing but I cannot because vapor freezes when creating a flow renderer again.

shaomeng:
> > > I cannot get anything to render with WRF/2D. I tried messing with every setting, increasing the number of seeds, different seeding methods and I made sure all of the referenced variables were valid.
> > 
> > 
> > @StasJ if you increase the velocity multiplier from 1 to 100K, then you'll see something. This particular data set uses different units.
> 
> Do we expect the user to have to scale by 100,000 to see anything at all? I mean thats 100 times what the limit was before you just increased it. Also, I tried re-testing but I cannot because vapor freezes when creating a flow renderer again.

Regarding your first question, I don't have a good answer too. The full domain is in the order of 1e+6 and the particles were moving at a scale of 1e-2 before the adjustment. Or in an extreme case, the velocity field that the user chooses happen to have all zeros, and the particles don't move anywhere. I don't have a good solution for either of these situations. 

Regarding your 2nd question, I personally cannot reproduce it on cisl-vapor. Maybe 1) you provide a more detailed steps, and 2) @sgpearse can try it on a macbook?

clyne:
> > > > I cannot get anything to render with WRF/2D. I tried messing with every setting, increasing the number of seeds, different seeding methods and I made sure all of the referenced variables were valid.
> > > 
> > > 
> > > @StasJ if you increase the velocity multiplier from 1 to 100K, then you'll see something. This particular data set uses different units.
> > 
> > 
> > Do we expect the user to have to scale by 100,000 to see anything at all? I mean thats 100 times what the limit was before you just increased it. Also, I tried re-testing but I cannot because vapor freezes when creating a flow renderer again.
> 
> Regarding your first question, I don't have a good answer too. The full domain is in the order of 1e+6 and the particles were moving at a scale of 1e-2 before the adjustment. Or in an extreme case, the velocity field that the user chooses happen to have all zeros, and the particles don't move anywhere. I don't have a good solution for either of these situations.
> 
> Regarding your 2nd question, I personally cannot reproduce it on cisl-vapor. Maybe 1) you provide a more detailed steps, and 2) @sgpearse can try it on a macbook?

Regarding mismatched units: I don't believe that there is any way that we can automatically perform a unit conversion (i.e. pick an appropriate scaling factor). We could calculate something that would result in pathlines progressing, but unless the scaling factor chosen is correct the resulting pathlines will be wrong. Bottom line: users have to know their data, and if they're mixing units they will have to address this themselves by either selecting an appropriate scaling factor or deriving new variables. That said, we should not place limits on the scaling factor. 

One thing that we might consider to reduce confusion is to render glyphs (e.g. spheres, diamonds) at the seed locations. Let's discuss f2f.

shaomeng:
Alright, the multiplier upper limit is removed in commit 
0e1fd05 . Happy testing everyone!

clyne:
> Alright, the multiplier upper limit is removed in commit
> [0e1fd05](https://github.com/NCAR/VAPOR/commit/0e1fd0538a1dea05302014d0e0226fb3470e7569) . Happy testing everyone!

Terrific. Thanks. Just need @StasJ to approve

PR #2062, opened 2020-01-07
Fix for #2058

### LINKED ISSUE #2058: Flow GUI Infinity Loop
To reproduce:
1. Open VAPOR, import CF compliant netcdf: `/glade/p/cisl/vast/vapor/Bugs/1950/rh_barry02l.2019071118.hwrfprs.storm.0p015.f003.nc`
2. Create a Flow renderer.
3. GUI enters infinity loop.

sgpearse:
Fix #2058

### DIFF
--- a/apps/vaporgui/VariablesWidget.cpp
+++ b/apps/vaporgui/VariablesWidget.cpp
@@ -314,6 +314,20 @@ void VariablesWidget::setDefaultVectorVar(
         }
 		else
             defaultVars.push_back("");
+
+        // If we have no variables that resemble U, V, or W, default to the
+        // first variable in the currently active dimension for GUI initialization
+        // purposes ( TFWidget, GeometryWidget, etc )
+        if ( 
+            defaultVars[0] == "" &&
+            defaultVars[1] == "" &&
+            defaultVars[2] == ""
+        ) {
+            std::vector<std::string> vars = _dataMgr->GetDataVarNames( _activeDim );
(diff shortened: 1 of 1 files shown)


sgpearse:
@StasJ I agree that it's a stop-gap.   Did you happen to see the comment in lines 318-320 though?

StasJ:
@clyne yes, but that comment does not explain why this code is necessary and how it fixes the problem, or even that there is a problem to be fixed. For example, why does setting the first var to something prevent an infinite loop in the GUI?

clyne:
> @clyne yes, but that comment does not explain why this code is necessary and how it fixes the problem, or even that there is a problem to be fixed. For example, why does setting the first var to something prevent an infinite loop in the GUI?

I assume this comment should have been directed to @sgpearse not clyne

StasJ:
> > @clyne yes, but that comment does not explain why this code is necessary and how it fixes the problem, or even that there is a problem to be fixed. For example, why does setting the first var to something prevent an infinite loop in the GUI?
> 
> I assume this comment should have been directed to @sgpearse not clyne

Ah yes. I read this from an email thread that was initiated when you requested my review so it said "from clyne" in gmail

clyne:
@sgpearse why was this PR closed? It hasn't been approved. It has unresolved comments that need to be approved by @StasJ

sgpearse:
I've discarded my changes for this branch, and opened a new PR, #2065

Issue #1375, opened 2019-01-25
3D variables with a dimension of length one are not handled correctly


StasJ:
In `Katrina`, the 3D variable `TSLB` causes some weird behavior:
- In the Slice renderer, it renderers but only for the first few refinement levels
- The wireframe renderer shows nothing
- The volume renderer freaks out:

![out](https://user-images.githubusercontent.com/2772687/51726663-b095d900-2025-11e9-9957-03b8ff2b3b9a.gif)

clyne:
Works for me. @stasj can you be more specific about what you are doing? FYI the default Z coordinate is top of the domain, which results in nothing being shown as might be expected. You'll need to move it.

StasJ:
1. Open Katrina from `/little_glade/data/VDC/vapor-3.2.0.alpha/WRF/Katrina`
2. Create a wireframe renderer
3. Select the variable `TSLB`
4. Enable the renderer

Nothing will be shown. I just checked and I can reproduce it following the above steps. I'm not sure what you mean by the Z coordinate being at the top of the domain. The geometry widget shows that the Z min and Z max of this variable are both 2.

<img width="872" alt="Screen Shot 2019-12-20 at 2 08 24 PM" src="https://user-images.githubusercontent.com/2772687/71293538-f3922680-2332-11ea-94b3-c1b85d1af2c2.png">

clyne:
Stalled pending resolution of #2048

clyne:
I am handing this over @sgpearse . Scott, note that there is an open branch with a partial fix, named "issue_1375". There are two problems here:

1. The StructuredGrid class was not handling degenerate grids where one or more dimensions were of length 1. This has been fixed in branch "issue_1375".

[...]

PR #2063, opened 2020-01-08
Issue 1375 geometry fix

sgpearse:


apps/vaporgui/GeometryWidget.cpp
```diff
@@ -462,6 +472,41 @@ void GeometryWidget::Update(ParamsMgr *paramsMgr,
...
+    if ( boxMin.size() != minFullExt.size() ||
+         boxMax.size() != maxFullExt.size() 
+    ) {
+        _box->SetExtents( minFullExt, maxFullExt );
+        return;
+    }
+
+    // Now check to see if the box is entirely outside of the variable's extents 
+    //
+    for ( int i=0; i<minFullExt.size(); i++ ) {
```
clyne: @sgpearse This algorithm is not quite right. It will reset the box if *any* of the axis fail to overlap. I believe we want to reset only when *all* of the axis fail to overlap. See https://stackoverflow.com/questions/20925818/algorithm-to-check-if-two-boxes-overlap

How about we add a BoxIntersect() method to the Box class? This would be a useful general utility.

sgpearse: @clyne Hmm, I do believe that the current algorithm is correct.  

In this bug, only the Z axis of TSLB and the other 3D variables fail to overlap.  However the X and Y axes do indeed overlap.  If we require all 3 to fail in overlapping, we miss the case reported in #1375.

clyne: You are correct, but the code is confusing. Checking boxMin < minFullExt and boxMax > maxFullExt are irrelevant. min is always less than or equal to max. So you simply need:

```
if ( boxMax[i] < minFullExt[i] II boxMin[i] > minFullExt[i]) {
  _box->SetExtents( minFullExt, maxFullExt );
  return;
}
```

clyne:
Closed and reopening with branch issue_1375

PR #2064, opened 2020-01-08
Fix  #1375 - 3D variables with a dimension of length one are handled correctly

### LINKED ISSUE #1375: 3D variables with a dimension of length one are not handled correctly

clyne:


clyne:
@stasj I think the case can be made that the behavior with the slice and barb renderer is correct, though perhaps unexpected. With both the slice and barb renderer we have a surface that that is close to planar, but not quite (WRF grids are essentially terrain following), and we're trying to sample it along a plane. The sampling plane simply does not intersect the surface. For both the slice and barb renderer a future enhancement would be to support slicing with logical (i-j-k) coordinates, and drawing barbs at grid nodes, respectively.

StasJ:
@clyne I see your point.

PR #2065, opened 2020-01-09
Updates to rake initialization sequence to fix #2058

### LINKED ISSUE #2058: Flow GUI Infinity Loop

sgpearse:
#2058 was the result of our rake being set with NaN values, when there are no valid field variables in the FlowRenderer.

For reasons I do not yet understand, the call to ParamsMgr::EndSaveStateGroup() (linked below) set Vapor into an infinite loop after issuing the call to ParamsBase::SetValueDoubleVec() with NaN.

https://github.com/NCAR/VAPOR/blob/310188472cc7fd6010cc12ad4701933b86724132/apps/vaporgui/FlowSubtabs.cpp#L532

The current fix is to test wheter we have valid variables.  We currently use dummy values of 0 and 1 for the rake extents until valid variables are assigned, and FlowSeedingSubtab::Update() is called.

Questions to consider:
1) Should SetValueDoubleVec() filter NaN values?
2) Should we be initializing parameters with NaN values?
3) Should the GUI be performing parameter initialization?

### DIFF
--- a/apps/vaporgui/FlowSubtabs.cpp
+++ b/apps/vaporgui/FlowSubtabs.cpp
@@ -72,8 +72,6 @@ FlowVariablesSubtab::Update( VAPoR::DataMgr      *dataMgr,
     assert(_params);
     
     _paramsMgr = paramsMgr;
-    
-    _variablesWidget->Update(dataMgr, paramsMgr, rParams);
 
     GUIStateParams *gp = dynamic_cast<GUIStateParams*>(_paramsMgr->GetParams(GUIStateParams::GetClassType()));
     int nDims = gp->GetFlowDimensionality();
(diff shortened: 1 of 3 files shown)


apps/vaporgui/FlowSubtabs.cpp
```diff
@@ -440,26 +442,32 @@ void FlowSeedingSubtab::_updateRake( VAPoR::DataMgr* dataMgr ) {
                                      minExt, 
                                      maxExt, 
                                      axes  );
-    // If there are no valid extents to set the rake with, just return
```
clyne: There is a lot of complexity in FlowSeedingSubtab::_updateRake(). Some of logic I don't completely understand. But it seems that much the complexity is due to the way FlowParams attempts to handle initialization of the rake via GetRake(). Two thoughts:

1. Can we replace the current rake implementation with a Box params class? This is the right thing to do, and it would  eliminate all of the special handling rake code we're currently struggling with.

2. If (1) is too much of an effort this late in the game, how about simply initializing the rake inside of the FlowParams constructor using the Box params class? This would get away from all of the nans, and complex logic inside of _updateRake.

clyne:
Rejected. @shaomeng and @sgpearse will refactor so rake uses Box class.

Issue #2059, opened 2020-01-07
Flow crash


sgpearse:
To reproduce:
1) Load the WRF/2D VDC, create a flow renderer
2) Delete the flow renderer
3) Create a new flow renderer.  Vapor will crash.

shaomeng:
Cannot reproduce with a Leigh Orf Tornado data set. Need more info.

StasJ:
I just reproduced this with the WRF/2D VDC dataset

shaomeng:
Investigation update:
Upon creation of the 2nd flow renderer, the variable `VariableWidget::_rParams` becomes a dangling pointer. Question: what's the correct behavior should be? A new Params object being created? The old Params object not being deleted?

shaomeng:
Investigation update:
This dangling pointer only occurs with data sets that have only 2D variables. I.e., the pointer is properly updated when 3D variables are present. I think there's some logic went wrong when `VariableWidget` handles 2D variables.

PR #2067, opened 2020-01-09
Fix for Issue #2059

sgpearse:
Note: I clobbered my previous changes in lieu of the simplest fix possible.  There will be other problems that still exist with the acceptance of this simpler fix.


shaomeng:
Alright, I'll merge it so I can start building installers.

Issue #2074, opened 2020-01-10
Flow GUI wrong Seed Distribution Settings


StasJ:
1. Create 2 flow renderers
2. Change the seed distribution type to Random for one of them

Both will now show the Random GUI.

sgpearse:
@clyne @shaomeng @StasJ 

Hey guys, we missed classifying this one yesterday.  Sam's kicking it over to me.  I think this one might be a High.  If you guys agree or disagree, let's mark it so.

shaomeng:
Scott, I think it's a high, and I tested my original fix, it was working.
More specifically, you reported a scenario in which it wasn't working, can
you confirm?

PR #2107, opened 2020-01-14
fix issue #2074

shaomeng:


sgpearse:
@shaomeng I do not believe this is quite fixed yet.  When I change one of the FlowRenderer's seed distribution to "Random", the other renderer will display "Gridded".  However the sliders for the gridded seed distribution do not appear, and we still have a "Seed count" slider that pertains to the Random seed distribution.

shaomeng:
> @shaomeng I do not believe this is quite fixed yet. When I change one of the FlowRenderer's seed distribution to "Random", the other renderer will display "Gridded". However the sliders for the gridded seed distribution do not appear, and we still have a "Seed count" slider that pertains to the Random seed distribution.

Sorry Scott, I thought it's a simple update problem, but apparently it's not... Will you take on this one since it's in the GUI realm?

sgpearse:
@shaomeng Sure, no problem.

PR #2110, opened 2020-01-15
Issue 2050

### LINKED ISSUE #1950: Loading a dataset into Flow caused crash
### Describe the bug
I imported a .nc storm file into VAPOR. Storm files are larger than 360+ megabytes. The storm file contained relative humidity data, and nothing else. In order to import large storm files, I had to strip out all of the other datasets, such as rainfall. I opened the file with a 3d volume renderer, where everything was fine. Then I closed the layers box and and opened the same file, except with a flow renderer. I received no error message about the file, VAPOR just immediately shut down. I re-opened VAPOR and tried to import the file I just used. The file successfully imported. Next, I tried to open the file with a flow renderer. Once more, VAPOR simply closed on me.

I tried attaching the file I used here, but it is too large for github, even compressed it still is 20 mb over the limit.

- [x] Did VAPOR crash?

**Impact**

### To Reproduce

I am not sure if this was a file size issue or not.

### Expected behavior
VAPOR closes upon selecting flow renderer.

### Attachments
Too big

### Desktop
 - Linux Redhat Enterprise 7
 - Vapor 3.2.0 alpha

sgpearse:
Updates for Box initialization in Flow Renderer.  Fixes #1950.

### DIFF
--- a/lib/params/FlowParams.cpp
+++ b/lib/params/FlowParams.cpp
@@ -32,6 +32,19 @@ FlowParams::FlowParams(   DataMgr*                 dataManager,
 {
 	SetVariableName("");
     SetDiagMsg("FlowParams::FlowParams() this=%p", this);
+
+    // At this point the base class is initialized, and the _Box is properly initialized
+    // to be the extents of the domain. Let's use that information to initialize the rake! 
+    std::vector<double> minext, maxext;
+    auto box = RenderParams::GetBox();
+    box->GetExtents( minext, maxext );
+    std::vector<float> floats( minext.size() * 2 );
+    for( int i = 0; i < minext.size(); i++ )
+    {
+        floats[i*2]   = minext[i];
+        floats[i*2+1] = maxext[i];
+    }
+    this->SetRake( floats );
 }
 
 FlowParams::FlowParams(   DataMgr*                dataManager, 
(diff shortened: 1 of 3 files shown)


shaomeng:
You mentioned yesterday that there're probably more changed required than simply removing NAN detectino. Are those changes in too?

sgpearse:
@shaomeng The changes I was referring to regard downstream bugs that we haven't been able to test for yet, due to this issue.  Next, we should hit the flow code hard with more tests and this will turn out new bugs.  This PR only regards #1950.

lib/params/FlowParams.cpp
```diff
@@ -206,11 +206,21 @@ std::vector<float> FlowParams::GetRake() const
...
     auto rakesize = doubles.size();
 
+    // If the rakesize isn't correct, then it isn't initialized yet.
+    // Let's use the values stored in _Box to initialize the rake.
     if ( rakesize != 6 && rakesize != 4 ) 
     {
-        // Six NANs represent the initial state
-        std::vector<float> tmp(6, std::nan("1") );
-        return tmp;
+        std::vector<double> minext, maxext;
```
clyne: Rather than initializing this in the getter, why not do the initialization in the constructor as is convention? It would also be consistent with the way Box class is initialized in the Renderer class.

shaomeng: Hi John, the rake initialization has been moved to the constructor in commit 86a4725, and this branch is ready to be reviewed again.

PR #2116, opened 2020-01-15
Re-add use single variable button to barbs

### LINKED ISSUE #2096: TFEditor: Cannot color barbs by secondary variable
I'm unable to find the option to color our barbs by a secondary variable.

To reproduce:
1) Load any dataset, create a barb renderer
2) Ensure that a "Color mapped variable" is selected in the Variables tab
3) Go to the Appearance tab.  The option to use a constant-color, or color by the secondary variable has gone missing.

StasJ:
Fix #2096

### DIFF
--- a/apps/vaporgui/BarbSubtabs.cpp
+++ b/apps/vaporgui/BarbSubtabs.cpp
@@ -56,7 +56,7 @@ void BarbGeometrySubtab::Update(
 
 BarbAppearanceSubtab::BarbAppearanceSubtab(QWidget* parent) {
 	setupUi(this);
-    verticalLayout->insertWidget(0, _tfe = new TFEditor(true));
+    verticalLayout->insertWidget(1, _tfe = new TFEditor(true));
     _tfe->SetShowOpacityMap(false);
 
 	_xDimCombo = new Combo(xDimEdit, xDimSlider, true);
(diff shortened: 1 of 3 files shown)


StasJ:
@sgpearse I agree but this was a simpler solution since every renderer params has a different mechanism for "use single color" and the TF editor is supposed to be somewhat universal, so the proper solution would involve refactoring multiple renderer params and renderers.

On hiding vs showing the TF editor, I think it is better to hide it when it is not applicable but I don't have a strong opinion so I am ok changing the behavior.

sgpearse:
@stasj 

> I agree but this was a simpler solution since every renderer params has a different mechanism for "use single color" and the TF editor is supposed to be somewhat universal, so the proper solution would involve refactoring multiple renderer params and renderers.

Interesting.  They should be using RenderParams::SetUseSingleColor(bool val).  I think that the Barb and Flow renderers are the only two that make use of this parameter.  It looks like they are using the aforementioned function, and the others defined in the RenderParams base class.  If there's a place that's not using the base class methods, we need to fix it.  I think that whether we do now, or later depends on how egregious the problem is.  If there's only one line in a renderer using the alternative const-color parameter, I think it's benign enough to fix.

> On hiding vs showing the TF editor, I think it is better to hide it when it is not applicable but I don't have a strong opinion so I am ok changing the behavior.

Thank you.

One more thing: I see the option for using a single color in the Barb renderer, but is there a way to pick what the color is?  I don't think I see a const-color picker for the flow renderer as well.

StasJ:
I started refactoring the renderers to use more appropriate params flags for using a color mapping variable but based on how “UseSingleColor” is used in certain places such as the color bar, I think it makes the most sense to leave it because in these cases, my replacement made even less sense than the old parameters. So I just addressed Scott’s concerns.

PR #2144, opened 2020-01-22
Fix for #2142

### LINKED ISSUE #2142: Too many open files
Vapor seems to be opening thousands of files without closing them.  Because of this, Vapor eventually gets into a state where it can no longer operate.  There seems to be several ways to reproduce this.

To reproduce:
1) Load Katrina-3.2.0.alpha.nc (located in /little_glade/data/VDC/vapor-3.2.0.alpha/WRF/Katrina)
2) Create a TwoDData renderer.  No need to enable.
3) Adjust the bounds of the Transfer Function editor ~20 times.
4) Load any of the pre-installed transfer functions.  Vapor will not be able to open any more files.

sgpearse:
Fixes #2142, which prevented vapor from opening any new files.

### DIFF
--- a/apps/vaporgui/hide_std_error_util.cpp
+++ b/apps/vaporgui/hide_std_error_util.cpp
@@ -11,10 +11,17 @@ static fpos_t pos;
 void HideSTDERR()
 {
 #ifndef WIN32
-    fflush( stderr );
-    fgetpos( stderr, &pos );
+    _savedSTDERR = -1;
+    if ( fflush( stderr ) != 0 )
+        return;
+    if ( fgetpos( stderr, &pos ) != 0 )
+        return;
 
-    _savedSTDERR = dup(STDERR_FILENO);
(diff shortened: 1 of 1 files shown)


apps/vaporgui/hide_std_error_util.cpp
```diff
@@ -6,17 +6,28 @@
 #endif
 
 static int _savedSTDERR;
+static fpos_t pos;
 
 void HideSTDERR()
 {
 #ifndef WIN32
+    fflush( stderr );
+    fgetpos( stderr, &pos );
```
clyne: Error checking is needed for any system call that returns an error status. It's not critical that we return an error message in the event of failure. So my recommendation would be to simply return from the function upon error, and set _savedSTDERR to an invalid value (e.g. -1). RestoreSTDERR() could then check _savedSTDERR before proceeding.

Issue #2157, opened 2020-01-24
Slice renderer restored from session bug


StasJ:
1. Open `/little_glade/stasj/Lee_Tornado`
2. Create/enable 4 slice renderers

Rotate the scene and notice a high FPS.

3. Save the session 
4. Restart vapor
5. Open the session

The visualizer is unusably slow, i.e. if I try rotating the scene I get about 1 FPS. Deleting the renderers solves this.

sgpearse:
@stasj What kind of file is that?  Session?  NetCDF-CF?  VDC?

StasJ:
It’s a directory

shaomeng:
Cannot verify because of #2184

PR #2165, opened 2020-01-28
Fix for #2157

sgpearse:
The slow rendering in #2157 was due to the SliceRenderer::CachedBox values not matching the extents of the SliceParams::Box.

The SliceRenderer has logic to accommodate moving domains: 

If the SliceParams::Box has extents outside of the data domain, we set the SliceRenderer::CachedBox to be within the data domain.  However the SliceParams::Box would still remain outside of the domain.  When we would subsequently compare SliceParams::Box to SliceRenderer::CachedBox, we would get a mis-match and resample.

The fix is to set SliceParams::Box whenever we update SliceRenderer::CachedBox.

### DIFF
--- a/lib/render/SliceRenderer.cpp
+++ b/lib/render/SliceRenderer.cpp
@@ -231,11 +231,15 @@ int SliceRenderer::_resetBoxCache() {
         return rc;
     }
     
-    // Moving domain allows area outside of data to be selected
+    // Moving domains do not update Box classes as they change their extents,
+    // so our Box becomes invalid.  We need to store valid extents for efficient
+    // sampling - IE we don't want to sample outside of the domain.
     for (int i = 0; i < _cacheParams.boxMax.size(); i++) {
         _cacheParams.boxMin[i] = max(_cacheParams.boxMin[i], _cacheParams.domainMin[i]);
         _cacheParams.boxMax[i] = min(_cacheParams.boxMax[i], _cacheParams.domainMax[i]);
     }
+    // Now set our box to have valid extents, within the data domain
+    p->GetBox()->SetExtents(_cacheParams.boxMin, _cacheParams.boxMax);
     
     _setVertexPositions();
     _resetTextureCoordinates();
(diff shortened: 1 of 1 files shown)


lib/render/SliceRenderer.cpp
```diff
@@ -231,11 +231,15 @@ int SliceRenderer::_resetBoxCache() {
         return rc;
     }
     
-    // Moving domain allows area outside of data to be selected
+    // Moving domains do not update Box classes as they change their extents,
+    // so our Box becomes invalid.  We need to store valid extents for efficient
+    // sampling - IE we don't want to sample outside of the domain.
```
clyne: I'm not sure that this is the right fix. It seems like the cached box class is being modified internally to keep the box inside the domain. To me it seems like this violates the notion of a cache. Secondly, according to the comment in the code it appears that the reason _cacheParams is being modified is to improve efficiency. I don't think performance is an issue here. Bottom line is I think we should consider removing the for loop that changes _cacheParams rather than modifying the user's selected region. The latter may lead to unintended consequences when the domain moves back to whence the region was selected.

sgpearse: I see what you mean.  I considered removing the domain/Box check, but chose this fix because I'm concerned about regressions.  

I'm not sure if this Box/Domain validation was a premature optimization, or if it was needed for another reason.  I don't think that this bug (#2157) was an easy one to find, and removing this code will require careful testing of some fringe edge cases.  

I think that setting the Box is less likely to break anything until we better support moving domains.  Just my 2c.

Let me know which you prefer and I'll move forward with your decision.

clyne: I appreciate concerns over regressions. However, in this case any regressions would be strictly limited to the slice renderer. I say we fix it the "right" way, and test it well

StasJ: Just chiming in since my review was requested, I would be against an "update cache" function modifying the thing it is caching.

sgpearse: Thanks guys.  Please re-review when ready.

PR #2167, opened 2020-01-28
Fix #2162 Barb renderer set correct colorbar variable

### LINKED ISSUE #2162: Barb Renderer: Colorbar not matching current TF
To reproduce:
1) Load any dataset, create a barb renderer and enable it
2) Make sure a color mapped variable is selected, and disable the "Use single color" checkbox
3) Load a built-in colormap, different from blue/white/red
4) Enable the colorbar annotation.  It will remain as blue/white/red

StasJ:
Fix #2162

### DIFF
--- a/include/vapor/BarbRenderer.h
+++ b/include/vapor/BarbRenderer.h
@@ -47,6 +47,9 @@ namespace VAPoR {
 	static string GetClassType() {
 		return("Barb");
 	}
+      
+  protected:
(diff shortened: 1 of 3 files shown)


StasJ:
I didn’t notice that but I’ll look into it

StasJ:
That was an unrelated bug but it was a simple fix so I added it to this PR.

PR #2168, opened 2020-01-28
Minor isosurface improvements

StasJ:


StasJ:
Before and after gif (notice ridges in bottom right flat part):

3nhl8h

PR #2187, opened 2020-01-30
Fix #2153 - Added size check before resizing Flow's vectors

### LINKED ISSUE #2153: Flow: Z Rake extents not preserved in session file
To reproduce:

1) Load any dataset, create a flow renderer
2) Reduce the Z Rake Maximum
3) Save Session, close Vapor
4) Load Vapor with the saved session file.  The Z Rake extents will be reset to the minimum/maximum.

sgpearse:
Fixes #2153.

The FlowSubtabs have a private variable called _numDims, which keeps track of what dimensionality is currently being used.  The default uninitialized value is -1.

When we opened a session, GUIStateParams would inform a dimensionality of 3.  This mismatch between _numDims and GUIStateParams triggers the Subtabs to resize the vectors used in FlowParams (seeds, periodicity, and rake).

This fix guards against the case where our vectors are already of correct size, and the GUI is uninitialized - in this case we want to set _numDims to whatever is in GUIStateParams, and not  resize anything.

### DIFF
--- a/apps/vaporgui/FlowSubtabs.cpp
+++ b/apps/vaporgui/FlowSubtabs.cpp
@@ -496,19 +496,27 @@ void FlowSeedingSubtab::_resizeFlowParamsVectors() {
 
     // Going from 3d vectors to 2d vectors.
     // Save the values we remove for restoration later on.
+    int seedSize = rakeSeeds.size();
+    int periodicitySize = periodicity.size();
+    int regionSize = rakeRegion.size();
+    VAssert( seedSize == 3 || seedSize == 2 );
+    VAssert( periodicitySize == 3 || periodicitySize == 2 );
+    VAssert( regionSize == 6 || regionSize == 4 );
+
     if ( _numDims == 2 ) {  
-        if ( rakeSeeds.size() != 2 ) {
+
+        if ( seedSize == 3 ) {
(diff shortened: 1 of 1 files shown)


apps/vaporgui/FlowSubtabs.cpp
```diff
@@ -497,29 +497,40 @@ void FlowSeedingSubtab::_resizeFlowParamsVectors() {
...
-        _oldZRakeNumSeeds = rakeSeeds[Z];
-        rakeSeeds.resize(2);
-
-        _oldZRakeMin = rakeRegion[Z_RAKE_MIN];
-        _oldZRakeMax = rakeRegion[Z_RAKE_MAX];
-        rakeRegion.resize(4);
-
-        _oldZPeriodicity = periodicity[Z];
-        periodicity.resize(2);
+        if ( rakeSeeds.size() != 2 ) {
```
clyne: This logic, as well as in the 'if' clauses below, assumes that rakeSeeds is either of size 2 or 3. Otherwise it will generate an invalid memory access. A more robust solution would avoid that assumption (consider the possibility of a 1D rake). At the very least asserts should be included to ensure this assumption is true.

sgpearse: Please let me know if you think the current changes are sufficient.

PR #2189, opened 2020-01-30
default deltaT is a little bigger

shaomeng:
This PR increases the default deltaT by a factor of 3.

### DIFF
--- a/lib/flow/VaporField.cpp
+++ b/lib/flow/VaporField.cpp
@@ -438,7 +438,7 @@ int VaporField::CalcDeltaTFromCurrentTimeStep( float& delT ) const
     // Let's find the maximum velocity on these sampled locations
     // Note that we want the raw velocity, which will be the value
     // returned by GetVelocity() divided by the velocity multiplier.
-    const float mult1o = 1.0f / _params->GetVelocityMultiplier();
+    const float mult1o = 1.0f / float(_params->GetVelocityMultiplier());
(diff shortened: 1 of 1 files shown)


shaomeng:
Well, there's unexpected behavior... I'll need to do more work here...

shaomeng:
@clyne  This PR is ready for merge now.

lib/flow/VaporField.cpp
```diff
@@ -436,21 +436,25 @@ int VaporField::CalcDeltaTFromCurrentTimeStep( float& delT ) const
             }
 
     // Let's find the maximum velocity on these sampled locations
+    // Note that we want the raw velocity, which will be the value
+    // returned by GetVelocity() divided by the velocity multiplier.
+    const float mult1o = 1.0f / float(_params->GetVelocityMultiplier());
```
clyne: Need to  avoid divide by zero

shaomeng: It cannot be zero. FlowParams defaults it to be 1.0, and FlowSubtabs rejects user input that is zero. 
That said, I've added an extra layer of guard in commit cb8e8b6

clyne: You mean it *should not* be zero :-). What happens if our future python or web interface sets it to zero? It just good coding practice to not rely on what is supposed to happen somewhere else in the code.

PR #2222, opened 2020-02-07
Ubuntu18 warnings

sgpearse:
Fixes warnings on Ubuntu 18.

These changes mostly involve unguarded conditionals, which regard inconsistent tabs after if statements.  We may want to consider fixing tabs throughout our codebase ('    ' instead of '\t')

I've also added a pragma to ignore an unused variable in hide_std_error_util.cpp


sgpearse:
Ok ok, changed to cast-to-void method!

PR #2227, opened 2020-02-11
PWidgets

StasJ:
Demo is under Menu Bar > Developer > Show Demo


StasJ:
@clyne, I took the concerns you mention into consideration when designing this so I have answers but I think it would be easiest to discuss face to face as you suggested. As a quick summary:

**Consistency**: ParamsWidgets guarantee to display the internal state. If the renderers are receiving a modified value from the params database that is where the consistency issues exist.

**Robustness**: The removal of thousands of lines of bug-prone GUI code that is spread out amongst multiple locations and files for every GUI element that can get out of sync will go much further towards creating robustness rather than maintaining the use of one off getters and setters which sometimes do much more than they are supposed to.

clyne:
Is there a reason we simply can pass pointers to the setter getter member functions into the PWidgets?

StasJ:
There is no technical issue preventing that but it goes against what ParamsWidgets were designed to be which is simple, efficient (design-wise), building blocks. Adding the ability for them to use function pointers will allow users of ParamsWidgets to circumvent the limitations specifically put in place for these design goals. If you need widgets that have more general functionality, you can use VaporWidgets. I can go into more detail but I think it would be easiest if we discuss in person. If you insist, however, I can add that functionality.

clyne:
Let’s discuss tomorrow at our meeting.

> On Feb 17, 2020, at 7:49 PM, StasJ <notifications@github.com> wrote:
> 
> There is no technical issue preventing that but it goes against what ParamsWidgets were designed to be which is simple, efficient (design-wise), building blocks. Adding the ability for them to use function pointers will allow users of ParamsWidgets to circumvent the limitations specifically put in place for these design goals. If you need widgets that have more general functionality, you can use VaporWidgets. I can go into more detail but I think it would be easiest if we discuss in person. If you insist, however, I can add that functionality.
> 
> —
> 

John Clyne
clyne@ucar.edu

apps/vaporgui/PTFEditor.h
```diff
@@ -0,0 +1,84 @@
...
+        Opacity,
+        Histogram,
+        Colormap,
+        IsoValues,
+        RegularIsoArray,
+        Default
+    };
+    
+    PTFEditor(const std::string &tag, const std::set<Element> elements = {Default}, const std::string &label = "");
+    PTFEditor *ShowOpacityBasedOnParam(const std::string &tag, int value);
```
clyne: Document

apps/vaporgui/PTFEditor.h
```diff
@@ -0,0 +1,84 @@
...
+    PTFEditor *ShowColormapBasedOnParam(const std::string &tag, int value);
```
clyne: Document public methods

apps/vaporgui/PVariableSelector.cpp
```diff
@@ -0,0 +1,26 @@
...
+PVariableSelector::PVariableSelector(const std::string &tag, const std::string &label)
+: PStringDropdown(tag, {}, label)
+{}
+
+void PVariableSelector::updateGUI() const
+{
+    RenderParams *rp = dynamic_cast<RenderParams*>(getParams());
+    assert(rp && "Params must be RenderParams");
+    
+    int nDims = rp->GetBox()->GetOrientation() == Box::XY ? 2 : 3;
```
clyne: This is not going to fly for variable dimension selection. There may be instances where a renderer mixes 2D and 3D variables for one thing. Perhaps the constructor should require a pointer to the method used to get the currently selected variable from the params data base, and then use the DataMgr to determine that variables dimensionality.  Note that most renderers strictly support either 2D or 3D variables, not both.

apps/vaporgui/PWidget.h
```diff
@@ -17,17 +28,62 @@ class PWidget : public QWidget {
...
     PWidget(const std::string &tag, QWidget *widget);
     void Update(VAPoR::ParamsBase *params, VAPoR::ParamsMgr *paramsMgr = nullptr, VAPoR::DataMgr *dataMgr = nullptr);
-    const std::string &GetTag() const;
+    
+    //! tag must be a key referencing a long value in the Params Database. If the associated value is equal
+    //! to whenEqualTo, the current widget will be shown/enabled, and hidden/disabled otherwise.
+    PWidget *ShowBasedOnParam(const std::string &tag, int whenEqualTo = true);
+    //! @copydoc PWidget::ShowBasedOnParam()
+    PWidget *EnableBasedOnParam(const std::string &tag, int whenEqualTo = true);
+    PWidget *SetTooltip(const std::string &text);
```
clyne: Document public methods

apps/vaporgui/PWidget.h
```diff
@@ -17,17 +28,62 @@ class PWidget : public QWidget {
...
 public:
     PWidget(const std::string &tag, QWidget *widget);
     void Update(VAPoR::ParamsBase *params, VAPoR::ParamsMgr *paramsMgr = nullptr, VAPoR::DataMgr *dataMgr = nullptr);
-    const std::string &GetTag() const;
+    
+    //! tag must be a key referencing a long value in the Params Database. If the associated value is equal
+    //! to whenEqualTo, the current widget will be shown/enabled, and hidden/disabled otherwise.
+    PWidget *ShowBasedOnParam(const std::string &tag, int whenEqualTo = true);
+    //! @copydoc PWidget::ShowBasedOnParam()
+    PWidget *EnableBasedOnParam(const std::string &tag, int whenEqualTo = true);
```
clyne: Document public methods.

apps/vaporgui/PWidgetHLI.h
```diff
@@ -0,0 +1,132 @@
...
+template <class PW, typename T, class P, typename... Args> typename std::enable_if<std::is_same<T, bool>::value, T>::type
+class PWidgetHLI2 : public PWidgetHLI<PW, T, P, Args...>
+{
+    virtual long getParamsLong() const override
+    {
+    }
+};
+
+//template <class PW, class P, typename... Args>
+//class PWidgetHLI2<PW, std::string, P, Args...> : public PWidgetHLI<PW, std::string, P, Args...> {
```
clyne: Remove commented out code. Use ifdef if it is something we might want to preserve for later.

apps/vaporgui/PWidget.h
StasJ: What are you referring to?

apps/vaporgui/PWidgetHLI.h
StasJ: Done

apps/vaporgui/PVariableSelector.cpp
StasJ: Sure but does this need to support every possible use case? Currently we manually code/configure every variable selector. This can take care of 90% of cases and the last 10% we can continue hard-coding as we do now. It could make sense to have the variable dimension to be 2D, 3D, or auto.

apps/vaporgui/VLineEdit.h
```diff
@@ -30,6 +30,7 @@ class VLineEdit : public VContainer {
     // Note: VLineEdit does NOT support an integer type at this point.
     // When isDouble == false, it represents a string. 
     void SetIsDouble( bool isDouble );
+    void SetReadOnly(bool b);
```
clyne: Document all public methods

apps/vaporgui/PRegionSelector.cpp
```diff
@@ -0,0 +1,35 @@
...
+    
+    layout->addWidget(new VLineItem("X", _sx = new QRangeSliderTextCombo));
+    layout->addWidget(new VLineItem("Y", _sy = new QRangeSliderTextCombo));
+    layout->addWidget(new VLineItem("Z", _sz = new QRangeSliderTextCombo));
+    
+    _sx->setEnabled(false);
+    _sy->setEnabled(false);
+    _sz->setEnabled(false);
+    
+//    _vgeometry->SetRange(<#const std::vector<float> &range#>)
```
clyne: Remove commented out code.

apps/vaporgui/VLineEdit.h
StasJ: Done but shouldn't the other methods of this class be documented then?

apps/vaporgui/PTFEditor.h
StasJ: Done

apps/vaporgui/PTFEditor.h
StasJ: Done

apps/vaporgui/PWidget.h
StasJ: Done

apps/vaporgui/VLineEdit.h
clyne: yes

apps/vaporgui/PRegionSelector.cpp
StasJ: This code is not finished because at the time of writing VGeometry selector was broken. I added code to prevent the class from being created but I need to revisit it to check if VGeometry has been fixed.

apps/vaporgui/VLineEdit.h
StasJ: I only added the method SetReadOnly, I did not write the rest of this class.

StasJ:
@clyne 

1. I am not sure what you are referring to as there are no range selectors in this PR and all of the widgets that allow the user to enter a value have a `SetRange(min, max)` method that can be used to extend that range.

2. I documented PWidget and it is not a wrapper for QWidget. It only inherits from QWidget because of limitations from existing code. The only public classes that do not have documentation are the ones I assumed were self-explanatory such as PCheckbox, PColorSelector, PFileOpenSelector, PFileSaveSelector, etc. I will add documentation to these but I don't know if it will add much, e.g. `\class PCheckbox is a GUI element that displays a checkbox which is synced with the params database`. I assume that can be inferred from the fact that it is a params widget called Checkbox

apps/vaporgui/PVariableSelector.cpp
clyne: Which use cases would it not support? It's not clear how you would "manually configure" this. The salient thing is we should be detecting the dimensionality by assuming that its dimensionality matches that of the *default* Box class.

clyne:
> @clyne
> 
> 1. I am not sure what you are referring to as there are no range selectors in this PR and all of the widgets that allow the user to enter a value have a `SetRange(min, max)` method that can be used to extend that range.

Whether it is a range selector with two controls or a value selector with a single control (slider) there are instances when we may allow the user to enter values that are outside of the default range that we have chosen for them. 

> 2. I documented PWidget and it is not a wrapper for QWidget. It only inherits from QWidget because of limitations from existing code. The only public classes that do not have documentation are the ones I assumed were self-explanatory such as PCheckbox, PColorSelector, PFileOpenSelector, PFileSaveSelector, etc. I will add documentation to these but I don't know if it will add much, e.g. `\class PCheckbox is a GUI element that displays a checkbox which is synced with the params database`. I assume that can be inferred from the fact that it is a params widget called Checkbox

Back when I used to write code if I were using an API my preferred method for figuring out what the API was capable of was to consult the documentation, not wade through all of the code. The latter can be pretty inefficient. I assume that I am not alone and others will on occasion RTFM :-). Yes, you need to document what each class does.

apps/vaporgui/PRegionSelector.cpp
StasJ: VGeometry is still broken (I tried VGeometry2 which I assume is the current version)

apps/vaporgui/PVariableSelector.cpp
StasJ: I don't know but I think it can support most of them without requiring a function pointer to initialize the dimensionality. Also that would not work in the case of the Flow or Wireframe renderers which change their dimensionality. I say the constructor creates a selector which always shows 2D vars, 3D vars, or automatically selects the var based on the determined dimension of the renderer. So for example:

Contour renderer:
```
VarSel("main", Auto);
VarSel("height", 2D);
```

Flow renderer:
```
VarSel("X", Auto);
VarSel("Y", Auto);
VarSel("Z", Auto);
VarSel("Color", Auto);
VarSel("Height", 2D);
```

If you have any more specific edge case rather than adding support for function pointers to allow VariableSelector to support everything just don't use it and do what we currently do or extend it.

StasJ:
@clyne 

1. Ah I see what you're talking about. My range selector widgets already support this but it is up to Vapor Widgets to add support for this. Params widgets only take the values given to them by the vapor widgets and sync them with the params database.

2. The idea is to write code that is itself the documentation when possible but I am adding doxygen to every class and public method as requested.

apps/vaporgui/PVariableSelector.cpp
clyne: I was focused on the issue with using the Box to determine the dimensionality and it just occurred to me that you don't have a "HLI" interface to the PVariableWidget. Is there a reason for this? Once implemented, isn't it simply a matter of calling the Params getter method for the variable and then querying the variable's dimensionality via the DataMgr?

StasJ: What would be the correct way to get the dimensionality? I tested various methods and this seemed to be the most reliable. Querying the dimensionality of the currently selected variable does not seem like a better solution. For example many variables slots have no variable as a valid option, e.g. the height variable or the Z variable for the flow or barbs. The reason there is no HLI is because I wrote this class as a specialized element that I've been using in my code, not a generic building block. If you need to call a specific function, for example to set the main variable for a renderer, a better solution would be to create a subclass, PMainVariableSelector, or something of that sort that can be used whenever you need to set the main variable rather than passing in the setter and getter functions for RenderParams::SetVariable for every renderer's GUI. If someone really needs an HLI version for some reason they can just call `CreateHLI(PVariableSelector,  std::string);`

apps/vaporgui/PWidget.h
```diff
@@ -17,17 +28,65 @@ class PWidget : public QWidget {
...
+    bool _usingHLI = false;
+    std::function<void(void*, long)>               _setterLong;
+    std::function<long(void*)>                     _getterLong;
+    std::function<void(void*, double)>             _setterDouble;
+    std::function<double(void*)>                   _getterDouble;
+    std::function<void(void*, const std::string&)> _setterString;
+    std::function<std::string(void*)>       _getterString;
+    
 public:
     PWidget(const std::string &tag, QWidget *widget);
```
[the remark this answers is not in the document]
StasJ: I think pass by const reference is fine.

apps/vaporgui/PCheckbox.h
```diff
@@ -1,12 +1,17 @@
...
+
+//! \class PCheckbox
+//! Creates a Qt Checkbox synced with the paramsdatabase using the PWidget interface.
+//! \copydoc PWidget
+
+class PCheckbox : public PLineItem {
     Q_OBJECT
     
-    QCheckBox *_qcheckbox;
+    VCheckBox *_qcheckbox;
```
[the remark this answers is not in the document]
StasJ: No.

apps/vaporgui/PDoubleInput.cpp
```diff
@@ -1,21 +1,23 @@
...
-    // _doubleInput->SetValue(value);
+    double value = getParamsDouble();
+    _doubleInput->SetValue(to_string(value));
 }
 
-void PDoubleInput::doubleInputValueChanged(double v)
+void PDoubleInput::doubleInputValueChanged(const std::string &v)
 {
-    getParams()->SetValueDouble(GetTag(), GetTag(), v);
+    double d = stod(v);
```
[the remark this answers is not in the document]
StasJ: Since VLineEdit is the vapor widget for entering doubles I was waiting for it to have a signal that gives a double. @sgpearse do you plan on adding this, does VLineEdit always guarantee the string is a double if SetIsDouble is true, or should I add safety checks here?

apps/vaporgui/PEnumDropdown.cpp
```diff
@@ -1 +1,38 @@
 #include "PEnumDropdown.h"
+#include <vapor/ParamsBase.h>
+#include <vapor/VAssert.h>
+#include "VComboBox.h"
+
+PEnumDropdown::PEnumDropdown(const std::string &tag, const std::vector<std::string> &items, const std::vector<long> &itemValues, const std::string &label)
+: PLineItem(tag, label, _vComboBox = new VComboBox(items)), _enumMap(itemValues)
```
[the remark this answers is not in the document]
StasJ: Then there will be duplicate entries.

apps/vaporgui/PGroup.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+PGroup::PGroup()
+: PWidget("", _widget = new QWidget)
+{
+    QVBoxLayout *layout = new QVBoxLayout;
+    layout->setMargin(0);
+    _widget->setLayout(layout);
+}
+
+PGroup *PGroup::Add(PWidget *pw)
```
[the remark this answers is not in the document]
StasJ: To allow call chaining.

apps/vaporgui/PColorSelector.h
```diff
@@ -4,19 +4,23 @@
...
-    PColorSelector(const std::string &tag, const std::string &label);
-    
-    static QColor VectorToQColor(const std::vector<double> &v);
-    static std::vector<double> QColorToVector(const QColor &c);
+    PColorSelector(const std::string &tag, const std::string &label="");
     
 protected:
     void updateGUI() const override;
+    
+    static QColor VectorToQColor(const std::vector<double> &v);
```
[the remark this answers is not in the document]
StasJ: No

apps/vaporgui/PVariableSelector.cpp
clyne: First, determining the variable dimensionality should only be an issue for the "Auto" configuration, correct? If the selector is configured 2D or 3D there is no need to figure out the dimensionality of the currently selected variable because it can't change.

In the case that the selector is configured "Auto" the dimensionality should be determined by querying the dimensionality of the currently selected variable using DataMgr:GetNumDimensions(). If there is no currently selected variable and the configuration is "Auto" then the choice is arbitrary. 

Relying on the dimensionality of the default Box returned by RenderParams::GetBox() might be working for you currently, but there is no guarantee that it will continue to work in the future.

StasJ: Ok. Done.

apps/vaporgui/PCheckbox.h
sgpearse: Unless there's a reason I don't see, could we rename _qcheckbox to _vcheckbox, since it's an instance of VCheckBox, not QCheckBox?

apps/vaporgui/PIntegerInputHLI.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+#include "PWidgetHLI.h"
+#include "PIntegerInput.h"
+
+CreateHLI(PIntegerInput, long);
```
sgpearse: I'm guessing there's a good reason for this, but I'm curious why each PWidget has its own file for these 4 lines of code?  Could this be consolidated in the relevant PWidget header file?   Or define individual instances of CreateHLI in a single header file?

apps/vaporgui/PWidget.cpp
```diff
@@ -17,21 +20,149 @@ void PWidget::Update(VAPoR::ParamsBase *params, VAPoR::ParamsMgr *paramsMgr, VAP
...
+        _dynamicUpdateInsideGroup = false;
+    }
+}
+
+void PWidget::_setParamsDouble(double v)
+{
+    if (_usingHLI)
+        _setterDouble(_params, v);
+    else
+        getParams()->SetValueDouble(getTag(), "", v);
```
sgpearse: I believe this will break if we intend on making the Params setters protected.

apps/vaporgui/PCheckbox.h
StasJ: done

apps/vaporgui/PGroup.cpp
```diff
@@ -0,0 +1,28 @@
...
+}
+
+void PGroup::updateGUI() const
+{
+    auto params = getParams();
+    auto paramsMgr = getParamsMgr();
+    auto dataMgr = getDataMgr();
+    
+    for (PWidget *child : _children)
+        child->Update(params, paramsMgr, dataMgr);
```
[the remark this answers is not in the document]
StasJ: That is correct.

apps/vaporgui/PSliderEdit.h
```diff
@@ -0,0 +1,62 @@
...
+//#include "VaporWidgetsFwd.h"
+
+class VSliderEdit;
+
+
+//! \class PSliderEdit
+//! Creates a Qt slider and text input combo synced with the paramsdatabase using the PWidget interface.
+//! \copydoc PWidget
+
+class PSliderEdit : public PLineItem, public PDynamicMixin {
```
sgpearse: Are there other PWidgets that make use of the PDynamicMixin?

apps/vaporgui/PCheckbox.h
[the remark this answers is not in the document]
shaomeng: Conversations of this fashion in multiple occasions render code review impossible to proceed. I'll revise my recommendation to "Approval."

apps/vaporgui/PVariableSelector.cpp
clyne: What happens in

`return getDataMgr()->GetNumDimensions(getParamsString());`

when 

`getParamsString()`

returns an empty value?

Also, you need to update the class documentation for PVariableSelector; it still references GetBox()

StasJ: It returns 0 as the number of dimensions so we get an empty drop down. I think this is a better option than it arbitrarily choosing 2D or 3D as then we may see other bugs caused by cases such as renderers getting 2D variables when they expect 3D or vice versa. I fixed the documentation.

apps/vaporgui/PWidget.h
```diff
@@ -17,17 +28,65 @@ class PWidget : public QWidget {
...
+    //! Follows the Vapor GUI update function convention. Update the element.
     void Update(VAPoR::ParamsBase *params, VAPoR::ParamsMgr *paramsMgr = nullptr, VAPoR::DataMgr *dataMgr = nullptr);
-    const std::string &GetTag() const;
+    
+    //! tag must be a key referencing a long value in the Params Database. If the associated value is equal
+    //! to whenEqualTo, the current widget will be shown/enabled, and hidden/disabled otherwise.
+    PWidget *ShowBasedOnParam(const std::string &tag, int whenEqualTo = true);
```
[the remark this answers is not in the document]
StasJ: I don't understand your comment. Are you saying the documentation is unclear?

apps/vaporgui/PIntegerInputHLI.h
StasJ: There are some minor benefits to doing this but mainly its just a design choice.

apps/vaporgui/PSliderEdit.h
StasJ: No, currently only PSliderEdit uses it.

apps/vaporgui/PIntegerInputHLI.h
sgpearse: Thanks.  Could you briefly elaborate on what the minor benefits are?  I'm fine resolving this, but I just don't know benefits.  I'm just concerned about the vaporgui directory getting unwieldy.

apps/vaporgui/PSliderEdit.h
sgpearse: Thanks.  I'm fine resolving this.  Do you think the mixin should be used by some or all other PWidgets?

apps/vaporgui/PIntegerInputHLI.h
StasJ: One would be potentially improved compile time. Many vapor files import way more than they need to leading to increased compile times and more files that need to be recompiled when a header is modified.

apps/vaporgui/PSliderEdit.h
StasJ: Only those that would benefit from supporting dynamic updates.

apps/vaporgui/PWidget.cpp
sgpearse: @clyne Have we decided on making the setters in params base protected, or to keep them public?

clyne: I'd like for them to be protected, but I think they are being used publicly throughout the application. Making them protected is probably outside the scope of this PR

clyne:
@StasJ as soon as you resolve these conflicts we an accept this. Thanks.

StasJ:
@clyne Done

PR #2229, opened 2020-02-13
Fix GUI flickering

StasJ:
This fixes the long standing issue of the GUI flickering on update as shown below.

before


StasJ:
This doesn't disable changes from being applied, it just delays redraws for the widgets until all changes are complete.

clyne:
> This doesn't disable changes from being applied, it just delays redraws for the widgets until all changes are complete.

sounds good

PR #2240, opened 2020-03-18
Fix overflow in raw2vdc and region size reporting in vdc2raw

pavolklacansky:
Converting large data sets (900 GB) would result in the overflow and then report either short read or error. I changed all affected types to size_t.

The vdc2raw would print the data dimensions instead of the region dimension. I added the extra print for the region case.


sgpearse:
This PR builds on all platforms, and introduces no new warnings on OSX.

PR #2256, opened 2020-03-25
Issue 2255 - Loading very large MPAS files results in an abort 

### LINKED ISSUE #2255: Loading very large MPAS files results in an abort
High resolution (~3km global) MPAS files cause vapor to abort when attempting to load them. 

To reproduce:
1. Import the MPAS files QVAPOR_MPAS.nc and x1.65536002.grid.nc from vapor/data/Source/MPAS/ThomasSchwitalla and vapor will abort after a couple of minutes

clyne:
Fixes #2255 

This PR fixes #2255 and also adds an initializer to the RenderParams class. The underlying problem was simply that there were portions of the code that were not 64bit clean. Compounding the problem was that, even after addressing the word size issue, the application asserted because insufficient memory was available to process 3D coordinate variables of this data set. An assert was thrown because the RenderParams class was attempting to compute variable extents inside of its constructor and constructors must be no-fail (this one was not). The addition of the RenderParams::Initialize() method allows for moving the extents query outside of the constructor and into Initialize() where error handling can take place. The addition of the initializer resulted in some changes upstream, namely in the VariablesWidget and various other RenderParams derived classes.

### DIFF
--- a/include/vapor/FlowParams.h
+++ b/include/vapor/FlowParams.h
@@ -42,6 +42,8 @@ public:
 
     virtual ~FlowParams();
 
+	virtual int Initialize();
+
     //
     // (Pure virtual methods from RenderParams)
     //
--- a/include/vapor/FlowParams.h
+++ b/include/vapor/FlowParams.h
@@ -42,7 +42,7 @@ public:
 
     virtual ~FlowParams();
 
-	virtual int Initialize();
+	virtual int Initialize() override;
 
     //
     // (Pure virtual methods from RenderParams)
--- a/include/vapor/FlowParams.h
+++ b/include/vapor/FlowParams.h
@@ -44,17 +44,6 @@ public:
 
 	virtual int Initialize() override;
 
-    //
-    // (Pure virtual methods from RenderParams)
-    //
-    virtual bool IsOpaque() const override
-    { 
-        return false; 
-    }
-    virtual bool usingVariable(const std::string& varname) override
-    {
-        return false;
-    }
 
     static std::string GetClassType() 
     {
--- a/include/vapor/ImageParams.h
+++ b/include/vapor/ImageParams.h
@@ -19,6 +19,8 @@ public:
 
   virtual ~ImageParams();
 
+  virtual int Initialize();
+
   static std::string GetClassType() 
   {
     return ("ImageParams");
--- a/include/vapor/ImageParams.h
+++ b/include/vapor/ImageParams.h
@@ -89,20 +89,6 @@ public:
     SetValueLong( _orientationTag, "set orientation value", val );
   }
 
-
-  //
-  // (Pure virtual methods from RenderParams)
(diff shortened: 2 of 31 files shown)


include/vapor/FlowParams.h
```diff
@@ -42,6 +42,8 @@ class PARAMS_API FlowParams : public RenderParams
 
     virtual ~FlowParams();
 
+	virtual int Initialize();
```
shaomeng:
```suggestion
	virtual int Initialize() override;
```

include/vapor/ImageParams.h
```diff
@@ -19,6 +19,8 @@ class PARAMS_API ImageParams : public RenderParams
 
   virtual ~ImageParams();
 
+  virtual int Initialize();
```
shaomeng:
```suggestion
  virtual int Initialize() override;
```

include/vapor/SliceParams.h
```diff
@@ -22,6 +22,8 @@ class PARAMS_API SliceParams : public RenderParams
 
  virtual ~SliceParams();
 
+ virtual int Initialize();
```
shaomeng:
```suggestion
 virtual int Initialize() override;
```

lib/params/FlowParams.cpp
```diff
@@ -64,6 +52,30 @@ FlowParams::~FlowParams()
     SetDiagMsg( "FlowParams::~FlowParams() this=%p", this );
 }
 
+int FlowParams::Initialize() {
```
shaomeng: This becomes interesting... RendererParams already has a private method called `_init()`. Now we introduce a public method called `Initialize()`. This design seems confusing...  Maybe one of them could be renamed to be more specific?

clyne: RenderParams::Initialize() is public and has a documented API. RenderParams::_init() is private. They both perform initialization.

include/vapor/ParamsMgr.h
```diff
@@ -432,6 +432,29 @@ class PARAMS_API ParamsMgr : public MyBase {
...
+ //! This method returns params that have been registered on the
+ //! ParamsMgr via RegisterAppParams() for the data set named
+ //! by \p dataSetName;
+ //!
+ //! \param[in] dataSetName  
+ //! \param[out] appRenderParams a vector of application render params associated with \p dataSetName  
+ //!
+ //! \sa ParamsMgr()
+ //
+ void GetAppRenderParams(string dataSetName, vector <RenderParams *> &appRenderParams) const {
```
shaomeng: Is it part of the c++ programming convention to place function implementations in the cpp file, instead of in the header ?

This particular header is widely included, meaning that it is repeatedly compiled many times throughout VAPOR.

clyne: Done

PR #2262, opened 2020-04-01
Change extension from .nc to .vdc

### LINKED ISSUE #2248: Change official VDC file extension to .vdc
We are currently in limbo supporting both .nc and .vdc as our official VDC file extension. This would involve changing help dialogues in the conversion utilities and possibly forcing them to use .vdc output files, as well as changing help dialogues in Vapor GUI.

StasJ:
In some cases I left the ability to use .nc extensions for usability. For example, in the Open Dataset dialog on OSX where disabling it would force users to rename their files. The sphinx documentation needs to be re-generated to include the changes but that can be done at the next release.

Fix #2248

### DIFF
--- a/apps/cfvdccreate/cfvdccreate.cpp
+++ b/apps/cfvdccreate/cfvdccreate.cpp
@@ -197,14 +197,14 @@ int	main(int argc, char **argv) {
 	}
 
 	if (argc < 3) {
-		cerr << "Usage: " << ProgName << " cf_files... master.nc" << endl;
+		cerr << "Usage: " << ProgName << " cf_files... master.vdc" << endl;
 		op.PrintOptionHelp(stderr, 80, false);
 		return(1);
 	}
 	
 
 	if (opt.help) {
-		cerr << "Usage: " << ProgName << " cf_files... master.nc" << endl;
+		cerr << "Usage: " << ProgName << " cf_files... master.vdc" << endl;
 		op.PrintOptionHelp(stderr, 80, false);
 		return(0);
 	}
--- a/apps/cfvdccreate/cfvdccreate.cpp
+++ b/apps/cfvdccreate/cfvdccreate.cpp
@@ -216,7 +216,7 @@ int	main(int argc, char **argv) {
 	string master = argv[argc-1];
 
     if (FileUtils::Extension(master) != "vdc")
-        printf("Warning: VDC files should the extension .vdc\n");
+        fprintf(stderr, "Warning: VDC files should the extension .vdc\n");
 
 	VDCNetCDF    vdc(opt.nthreads);
 
--- a/apps/cfvdccreate/cfvdccreate.cpp
+++ b/apps/cfvdccreate/cfvdccreate.cpp
@@ -215,6 +215,9 @@ int	main(int argc, char **argv) {
 	for (int i=0; i<argc-1; i++) cffiles.push_back(argv[i]);
 	string master = argv[argc-1];
 
+    if (FileUtils::Extension(master) != "vdc")
+        printf("Warning: VDC files should the extension .vdc\n");
+
 	VDCNetCDF    vdc(opt.nthreads);
 
 	if (vdc.DataDirExists(master) && !opt.force) {
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -61,6 +61,11 @@
 #include <vapor/STLUtils.h>
 #include <vapor/Proj4API.h>
 
+#include <vapor/VDCNetCDF.h>
+#include <vapor/DCWRF.h>
+#include <vapor/DCMPAS.h>
+#include <vapor/DCCF.h>
+
 #include "VizWinMgr.h"
 #include "VizSelectCombo.h"
 #include "TabManager.h"
@@ -251,8 +256,6 @@ void MainForm::_initMembers() {
 
 }
 
-#include <vapor/VDCNetCDF.h>
-#include <vapor/DCWRF.h>
 // Only the main program should call the constructor:
 //
 MainForm::MainForm(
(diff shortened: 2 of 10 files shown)


clyne:
@StasJ this PR changes the help messages issued such that the extension ".nc" is replaced by ".vdc", but that is all it does. I guess I was expecting the conversion utilities to automatically append a ".vdc", if none was present, when generating a .vdc file. What do @sgpearse and @shaomeng think?

sgpearse:
I don't have a rigid stance on this.  In my experience, other command-line file editing tools like NCO let users name generated files whatever they want.

I can't think of a reason why all .vdc files must always have only one extension, but maybe you guys can.  Conversly, enforcing .vdc adds more code to maintain.

StasJ:
@clyne this PR keeps the same behavior that we've had with the .nc extension. The functionality you suggested can be added but was not mentioned previously.

clyne:
> @clyne this PR keeps the same behavior that we've had with the .nc extension. The functionality you suggested can be added but was not mentioned previously.

The issue that you opened, #2248, states "possibly forcing them to use .vdc output files". I guess *possibly* is the operative word here. I would think that if we don't do something like automatically changing the output file to include the extension we probably won't accomplish much.

StasJ:
@clyne Fair point. I was referring to the fact that I didn't realize that was a requirement/expectation for this PR but I have no problem adding it.

There are 2 options:
1. We automatically append a .vdc extension if the file does not have one, i.e. if a user specifies out.nc, we output out.nc.vdc. This behavior mimics the system file save dialog.
2. We abort with an error code if the extension is not .vdc

1 seems more convenient to me however the advantage of 2 is that if a user has a script, this will clearly let them know we have started enforcing this file extension whereas option 1 may cause their script to silently fail and they have to debug it.

clyne:
@StasJ per our discussion this afternoon let's just add a warning message if the .vdc extension is not used.

apps/cfvdccreate/cfvdccreate.cpp
```diff
@@ -215,6 +215,9 @@ int	main(int argc, char **argv) {
 	for (int i=0; i<argc-1; i++) cffiles.push_back(argv[i]);
 	string master = argv[argc-1];
 
+    if (FileUtils::Extension(master) != "vdc")
+        printf("Warning: VDC files should the extension .vdc\n");
```
clyne, line 219: Should warnings go to stdout or stderr?

StasJ: Fixed.

StasJ:
> vapor doesn't seem to recognize the .vdc extension when the file is specified on the command line. The application starts, but the data set isn't loaded.

That was never actually a feature that vapor supported. I just threw in some basic support for this for my own convenience when testing so I did not bother updating it here. I can clean up that code a bit and update it if we want to make it a feature.

clyne:
I find it to be a highly useful feature. Please make it work.

StasJ:
Done

StasJ:
@clyne That came across wrong. I was just explaining why I didn't update it initially. I put that there when I was developing a feature a long time ago. Since I had already been using .vdc for a while it didn't work for me and I considered it old development code. It now works for vdc, netcdf, and wrf files.

apps/vaporgui/MainForm.cpp
```diff
@@ -395,18 +396,30 @@ MainForm::MainForm(
...
-	if (files.size() && files[0].endsWith(".nc")) {
-        VDCNetCDF vdc;
-        bool errReportingEnabled = Wasp::MyBase::EnableErrMsg(false);
-        int ret = vdc.Initialize(files[0].toStdString(), {}, VDC::R);
-        Wasp::MyBase::EnableErrMsg(errReportingEnabled);
-        if (ret < 0) {
-            loadDataHelper({files[0].toStdString()}, "NetCDF CF files", "", "cf", true);
-        } else {
+	if (files.size()) {
+        if (files[0].endsWith(".nc")) {
```
clyne, line 400: @stasj it's great to have this, but we may want to think about changing the file type detection logic a bit. Consider:

1. NetCDF CF files usually have .nc extensions, but not always
2. WRF files usually do not have any extension, but sometimes .nc is used
3. The DC*::Initialize() method should work reliably for detecting wrf files, VDC files, AND MPAS files, but DCCF::Initialize() will succeed for VDC, WRF, and MPAS files. 

Given that I would suggest NOT using extensions and testing for file types using DC*::Initialize() in this order:

VDC, WRF, MPAS, CF

As far as order goes the only one that is really important is to have CF NetCDF last. 

What do you think?

StasJ: I see, I can make that change. I included the file extensions here to mildly enforce the naming conventions however you seem to be saying that they are all over the place. Therefore I think it makes sense to ignore the file extension entirely and determine the file type based on DC*::Initialize().

StasJ: Done. We could consider replacing `Open VDC` and `Import` with just `Open Dataset` and use the method used here.

apps/vaporgui/MainForm.cpp
```diff
@@ -428,6 +435,30 @@ MainForm::~MainForm()
...
+    if (isDatasetValidFormat<VDCNetCDF>(paths))
+        *fmt = "vdc";
+    else if (isDatasetValidFormat<DCWRF>(paths))
+        *fmt = "wrf";
+    else if (isDatasetValidFormat<DCMPAS>(paths))
+        *fmt = "mpas";
+    else if (isDatasetValidFormat<DCCF>(paths))
+        *fmt = "cf";
+    else
+        return -1;
```
clyne, line 458: The convention in VAPOR is to return a boolean if the function does not call MyBase::SetErrMsg on failure.

StasJ: Done but this is not mentioned in the official Vapor coding conventions.

clyne: See https://github.com/NCAR/VAPOR/wiki/Proper-error-log-and-report-within-VAPOR

StasJ: I missed that, sorry.

PR #2275, opened 2020-04-11
Fixed #2274 - MPAS edge variables not handled correctly

### LINKED ISSUE #2274: MPAS edge variables not handled correctly
Variables that are sampled on cell edges in MPAS are not correctly processed.

To reproduce:
1. Load any of the data in /glade/p/cisl/vast/vapor/data/Source/MPAS/ValidationRuns
2. Create a slice renderer and load the "pv_edge" variable. All of the values above the bottom level will be zero

clyne:


sgpearse:
@clyne - I'm not able to see any difference between this branch and master.  The bug report says to put the Slice on the lowest level, but does not specify a value.  The lowest level I could find with actual (non-missing-value) data was at a height of 35.  

The following screenshot shows a side-by-side comparison (master on left, issue_2274 on right) of the MPAS file /glade/p/cisl/vast/vapor/data/Source/MPAS/ValidationRuns/x1.163842.output.2010-10-23_00.00.00.nc.

Screen Shot 2020-04-13 at 8 36 41 AM

clyne:
I've updated the steps to reproduce as follows:

1. Load any of the data in /glade/p/cisl/vast/vapor/data/Source/MPAS/ValidationRuns
2. Create a slice renderer and load the "pv_edge" variable. 
3. Change the colormap range to -0.0004 to 0.0004, and set the Z level to ~300.
All of the values be ~zero

PR #2277, opened 2020-04-12
Fixed 2273 - Missing values not correctly handled

clyne:


clyne:
Good catch. I was testing with raw data, not the VDC version. Apparently there was a bug that was preventing 0-dimensional variables from being copied into the VDC. It's been fixed, but you'll need to regenerate the VDC version of the data set to test.

StasJ:
Sounds good, thanks. Should I regenerate all our VDC files on little_glade?

clyne:
Two questions:

1. do we have a naming convention that will allow for the nightly builds?

2. How are we sync’ing with the actual glade? There is a fare bit of work that can’t be done on cisl-vapor because of limited memory, storage, etc., and it is important to keep big glade up to date.

thanks - jc


> On Apr 12, 2020, at 3:38 PM, StasJ <notifications@github.com> wrote:
> 
> 
> Sounds good, thanks. Should I regenerate all our VDC files on little_glade?
> 
> —
> You are receiving this because you authored the thread.
> 

John Clyne
clyne@ucar.edu

StasJ:
1. I don't know? We could come up with one. I didn't know we were planning to rebuild these nightly especially since they take like 8 hours on cisl-vapor.

2. We are not syncing with glade. There are issues with authorization that hinder automatic syncing. Also if there is issues with limited storage on cisl-vapor how would we keep such data on little_glade in sync with glade?

PR #2289, opened 2020-04-24
Ubuntu warnings

sgpearse:
This PR includes a few changes to conform to Qt5.13.4, adding guards to ungarded if statements, and casting unused variables to void.

### DIFF
--- a/lib/render/ShaderManager.cpp
+++ b/lib/render/ShaderManager.cpp
@@ -21,7 +21,7 @@ using std::pair;
 #ifdef NDEBUG
     #if SHADER_AUTORELOAD
         #ifndef WIN32
-            #warning Disabling shader autoreloading
+            //#warning Disabling shader autoreloading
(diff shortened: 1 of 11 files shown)


lib/render/ShaderManager.cpp
```diff
@@ -21,7 +21,7 @@ using std::pair;
 #ifdef NDEBUG
     #if SHADER_AUTORELOAD
         #ifndef WIN32
-            #warning Disabling shader autoreloading
+            //#warning Disabling shader autoreloading
```
StasJ, line 24: Why is this commented out?

sgpearse: Because it emits a warning.  Do you know of a better alternative?

shaomeng:
Hi Scott, I still see 3 warnings regarding deprecated declarations in Qt. Is it what you see as well?

PR #2290, opened 2020-04-30
Fixed #2276 - Add support for deriving zonal and meridional winds for MPAS model outputs

### LINKED ISSUE #2276: Add support for deriving zonal and meridional winds for MPAS model outputs
MPAS outputs do not have zonal and meridional winds. Instead the u and v variables in MPAS represent velocities that are orthogonal and tangential to cell edges. These values are not correlated as they change with the orientation of the edges around a cell. The zonal and meridional winds can easily be derived from these values. See:

https://www2.mmm.ucar.edu/projects/mpas/tutorial/Boulder2018/slides/5._mesh_structure.pdf

An image is attached showing the u field contained in an MPAS file and the zonal component derived from u and v

U
Uzonal

clyne:


lib/vdc/DCMPAS.cpp
```diff
@@ -353,6 +370,32 @@ int DCMPAS::initialize(
...
+
+		for (auto varname : {"Uzonal", "Umeridional"}) {
+			bool zonalFlag = (varname == string("Uzonal"));
+			DerivedZonalMeridonal *derivedVar = new DerivedZonalMeridonal(
+				varname, this, ncdfc, uNormalVarName, uTangentialVarName,
+				zonalFlag
+			);
+
+			rc = derivedVar->Initialize();
+			if (rc<0) return(-1);
```
shaomeng, line 389: if initialization failed, and this function is aborted by return, then who's responsible for freeing the memory allocated by new?

lib/vdc/DCMPAS.cpp
```diff
@@ -353,6 +370,32 @@ int DCMPAS::initialize(
...
+
+			_dvm.AddDataVar(derivedVar);
+
+			DC::DataVar dvarInfo;
+			(void) derivedVar->GetDataVarInfo(dvarInfo);
+			_dataVarsMap[varname] = dvarInfo;
+		}
+	}
+
```
shaomeng, line 398: Also, when this function finishes in a normal state, who's responsible for freeing the memory allocated by new?

lib/vdc/DCMPAS.cpp
```diff
@@ -1900,3 +1932,212 @@ bool DCMPAS::DerivedCoordVertFromCell::VariableExists(
...
+	string derivedVarName, DC *dc, NetCDFCollection *ncdfc,
+	string normalVarName, string tangentialVarName, bool zonalFlag
+
+) : DerivedDataVar(
+	derivedVarName
+) {
+	_dc = dc;
+	_ncdfc = ncdfc;
+	_normalVarName = normalVarName;
+	_tangentialVarName = tangentialVarName;
```
shaomeng, line 1954:
```suggestion
	_tangentialVarName = std::move(tangentialVarName); // pass-by-value and move 
```

The same applies to variable `_normalVarName`

lib/vdc/DCMPAS.cpp
```diff
@@ -1900,3 +1932,212 @@ bool DCMPAS::DerivedCoordVertFromCell::VariableExists(
...
+
+
+int DCMPAS::DerivedZonalMeridonal::ReadRegion(
+	int fd,
+    const vector <size_t> &min, const vector <size_t> &max, float *region
+) {
+	VAssert(min.size() == 2);
+	VAssert(min.size() == max.size());
+
+	DC::FileTable::FileObject *f = _fileTable.GetEntry(fd);
```
shaomeng, line 2038: Line 2016 checks `_fileTable.GetEntry()` returning a nullptr, while this line doesn't check it. Is this check necessary?

lib/vdc/DCMPAS.cpp
clyne: Good catch! Fixed.

lib/vdc/DCMPAS.cpp
clyne: The DCMPAS destructor frees all of the dynamically created derived variables.

lib/vdc/DCMPAS.cpp
clyne: _normalVarName is a string, not a raw pointer. So I don't understand the comment.

lib/vdc/DCMPAS.cpp
clyne: Indeed. I've fixed it.

lib/vdc/DCMPAS.cpp
shaomeng: "pass-by-value-and-move" is a programming idiom that avoids unnecessary memory allocations. It's especially useful when the parameters passed around have dynamically allocated memory space, like string in this case. Please see this document for more detail:
https://paper.dropbox.com/doc/Code-Snippet--Az4I~xnjxteyg~vhPGZWTU21Ag-H2oXdHWnNj5b2c4aMovCh

clyne: What are the practical benefits of pass-by-value-and-move for small strings and a function that will only be called a handful of times?

shaomeng: Well, I see it very powerful just because of its simplicity -- you do it in this fashion **whenever** you receive some complex object from the parameter and make a copy by yourself.

- Do you need to evaluate if the string/vector is a big one or small one? No. 
- Do you need to evaluate if this function is called repeatedly or occasionally? No. 
- Do you need to evaluate if the parameter is a const type or not? No
- Do you need to evaluate if you should pass by value or by reference? No

Plus, it's really not big difference in terms of typing. Why not?

shaomeng: To be clear, it guarantees the least number of dynamic memory allocations in every case without the need of the programmer to choose between design options such as values/references, const/non-const, etc.

clyne: I understand that it is more efficient in terms of memory space and maybe even computation, but this is not a place where any of that matters. I don't really care about making the change here, but I don't want to start down the road where every line of C++ code is scrutinized for optimal performance. We should focus our performance tuning efforts where memory usage or computational efficiency really matters. Where these considerations are not important we should emphasize readability, simplicity, robustness, and maintainability. Do we need to, for example, worry about the use of postfix increment operators (x++)? Prefix operators (++x) are more efficient, and should be used when looping over millions of elements (such as when using the Grid iterators), but are we going to require prefix operators for looping over n==3 elements because it saves a few nanoseconds :-). I hope not.

shaomeng: Hmm, I think your argument applies better to a scenario where option A and B both have some pros and cons, but in inconsequential cases, choose whatever you like instead of an optimal one. I agree with this philosophy.

In my view, the current scenario is that option A is superior to option B, all the time. As a result, I'm advocating that we always use option A and forget there's even an option B  there. Pass-by-value-and-move is option A in this scenario.

I'm advocating this practice because I often find making decisions from available choices imposes a mental burden on me, and I'm always happy to eliminate available choices when writing code. As a team, if all code follows this programming idiom, then even if another function comes to the reviewer and doesn't show any obvious pattern on its frequency being called and average size of objects being passed around, the reviewer can still be assured that there's no possible improvement. But afterall if you feel there are other reasons to keep it the way it currently works, I'm happy to mark this conversation resolved.

shaomeng: Let's move ahead without spending too much time on this issue :)

PR #2300, opened 2020-06-05
Fix dbus

sgpearse:
This fix includes dbus on all Unix platforms, on all build systems.  I also reverts our CentOS docker image to use the same version as used on Casper (7.4.1708).


site_files/site.NCAR
```diff
@@ -21,6 +21,7 @@ if (UNIX)
 	set (PYTHONVERSION 3.6)
 	set (PYTHONDIR ${THIRD_PARTY_DIR})
 	set (PYTHONPATH ${THIRD_PARTY_LIB_DIR}/python${PYTHONVERSION})
+	set (EXTRA_LIBS_SEARCH ${EXTRA_LIBS_SEARCH} dbus)
```
clyne, line 24: Not really sure how cmake works, but Is this going to present a problem if dbus does NOT exist in EXTRA_LIBS_SEARCH, but would still be found in the system libraries?

clyne: You say: "If possible, it is generally better to remove the dependency rather than include it this way."

Is there a preferable way to do this?

StasJ: dbus is an optional dependency of Qt so we could compile Qt without it, which would remove it as a dependency. It may not be worth recompiling Qt so we can postpone it by packaging dbus with the installer for now.

clyne: Understood. Thanks.

PR #2301, opened 2020-06-23
Issue2294

sgpearse:
This PR brings proposed changes described in the following document.

https://docs.google.com/document/d/1Uerdlt9V-jNjd3j6173Oo-oxF_jY37qV93sbpUFRPZk/edit


clyne:
A few questions 

I assume that the intent is to replace all of the existing V* widgets, not have the new and old V* widgets co-exist.

@stasj, @sgpearse, will these widgets play nicely with the params widgets? I.e. Are they essentially plug-n-play with the older V* widgets?

Can the lengths of the slides be adjusted (increased)? They're currently pretty short relative to our previous ones, and this limits the fidelity.

StasJ:
@clyne, I think @sgpearse could answer best but I believe since they add additional features like allowing users to change the number of decimal digits which need to be synced with the params database, these features will need to be added to PWidgets.

I agree about the slider lengths. I noticed this too but forgot to mention it.

StasJ:
To clarify, as far as I can tell they should be a drop in replacement for
the existing functionality. Adding support for the new configuration
options, however, will require work on PWidgets to sync these new options
with the params database in addition to the widget values themselves.

clyne:
I changed this to a draft PR until the test functions are removed from the slice renderer, and all of the public classes/methods in the V* .h files are documented with Doxygen.

sgpearse:
On the way to wrapping things up.  A few comments:

@clyne 

> I assume that the intent is to replace all of the existing V* widgets, not have the new and old V* widgets co-exist.

This is not true.  The refactoring of VLineEdit is being done because it was previously incomplete.  Additionally, it cobbled logic together to handle strings, ints, and doubles.  This PR breaks each type handler into its own class.

@StasJ 

> SetNumDigits will cause the updated truncated value to be emitted. Personally I think if the numdigits is lower than the precision of the stored value, it should just show a truncated version rather than truncating the actual value, however I don't know what the intended behavior is.

I debated about this too, but decided that we want the GUI to be a pure reflection of the Params database.  I wouldn't want a scientist to believe they're using a value that they're not.

sgpearse:
Hi @clyne and @StasJ, this is ready for re-review.

Fixes:
VDoubleSliderEdit and VIntSliderEdits can now have their min/max values (optionally) modified through user input
Size of the sliders in V*SliderEdits have been increased
Templatized VLineEdits have been removed
Renamed the old VLineEdit to VLineEdit_Deprecated
Doxygen documentation

clyne:
@sgpearse before I complete the review I want to have some discussion about the  design of the V*LineEdit. VIntLineEdit, VDoubleLineEdit, and VStringLineEdit all inherit from VAbstractLineEdit, which is essentially a wrapper for QLineEdit. QLineEdit works on strings; numerical values are formatted into strings and handed to a QLineEdit. 

We want building blocks that work on strings, ints, and floats, which is what the various V* specializations provide. However, much of the type (int, float, string) specialization seems to be handled (awkwardly) in VAbstractLineEdit, which provides type-specific versions of SetValue(), etc. If an instance of VStringLineEdit calls SetValue(int), or SetNumDigits() what does this mean?  It might make more sense to have VAbstractLineEdit only operate on strings (like QLineEdit), and move the type specialization into the respective V* classes. It might even makes sense to have VNumeric* class that inherits from VAbstractLineEdit, and is derived by VInt* and VDouble*. 

@stasj what are your thoughts on this?

StasJ:
@clyne I would agree. Since VStringLineEdit needs to have essentially the same functionality as QLineEdit (editing strings), you could have VStringLineEdit inherit from QLineEdit. The numeric editors could inherit from VStringLineEdit and just add on the numeric restrictions.

Technically you would want VIntEdit and VDoubleEdit to inherit separately from VStringEdit since the double input adds right-click menus which don't apply to the integer input, however, if you want to consolidate the number restricting code a simple solution would be to have VIntEdit inherit from VDoubleEdit and just restrict it to 0 decimals.

clyne:
> @clyne I would agree. Since VStringLineEdit needs to have essentially the same functionality as QLineEdit (editing strings), you could have VStringLineEdit inherit from QLineEdit. The numeric editors could inherit from VStringLineEdit and just add on the numeric restrictions.
> 
> Technically you would want VIntEdit and VDoubleEdit to inherit separately from VStringEdit since the double input adds right-click menus which don't apply to the integer input, however, if you want to consolidate the number restricting code a simple solution would be to have VIntEdit inherit from VDoubleEdit and just restrict it to 0 decimals.

@sgpearse does this make sense?

sgpearse:
@clyne I see what you mean.

@StasJ All of our VWidgets currently inherit from VContainer, which provides them consistent margins and size policies.  Inheriting from QLineEdit would unfortunately be inconsistent with the rest of the code base.  Implementing the VIntLineEdit as a VDoubleLineEdit with 0 decimals interestingly would not work, because integers displayed with scientific notation do have decimal precision.

I'll propose removing VAbstractLineEdit, and defining the SetValue and ValueChanged methods in concrete classes (VStringLineEdit, VIntLineEdit, VDoubleLineEdit).  

Then, implement VNumericLineEdit which would hold a menu and members to support it.  This would be inherited by the VIntLineEdit and VDoubleLineEdit.

Thoughts?

sgpearse:
Guys, I've made changes to the V*LineEdit relationships as shown in the diagram below.  I believe this addresses your concerns except for the points raised about significant figures.  I think that requires more discussion, so I'll leave this for review and if we agree that decimals is insufficient, we can revisit that issue next.

Line Edits v2 (1)

clyne:
Thanks for the diagram, @sgpearse. The only question I have is should VNumericLineEdit inherit from VStringLineEdit since ultimately the underlying QtLineEdit operates on strings? I.e. VNumericLineEdit will ultimately have to convert its stored value to a string, so should this be done by deriving from VStringLineEdit? I don't really have an opinion. Just asking.

sgpearse:
@clyne - VNumericLineEdit inheriting from VStringLineEdit doesn't make sense in my opinion.

VStringLineEdit has member functions for setting, getting, and emitting values of type std::string.  The numerics wouldn't need to inherit these members.

clyne:
> @clyne - VNumericLineEdit inheriting from VStringLineEdit doesn't make sense in my opinion.
> 
> VStringLineEdit has member functions for setting, getting, and emitting values of type std::string. The numerics wouldn't need to inherit these members.

VNumericLineEdit (or its derivatives) will have to handle setting, getting and emitting strings for QLineEdit. It could simply inherit the string typed functions as private and provide public versions of these functions that work on ints and floats. I'd like to hear @stasj thoughts.

StasJ:
I think it would be beneficial to have a single "wrapper" for QLineEdit rather than having two. This is because, for example, if we want to change when a line edit emits the "finished editing" signal, as we have in the past, we would have to make sure to change it for both VStringEdit and VNumericEdit. To address Scott's concern, which I agree is valid, we could delete the string setter/getter functions in VNumericEdit like John suggested. Another option is to have VNumericEdit own a VStringEdit rather than inherit from it, replacing its QLineEdit. This would address the concern as it would no longer inherit unused string setters/getters, provide the functionality we need (single access-point to QLineEdit), and require minimal changes.

clyne:
Both of those options sound reasonable to me.

sgpearse:
VStringLineEdit is now a base class for all line edits, and handles all interactions with the QLineEdit.  There is one exception: a VInt/DoubleLineEdit still need to reconnect the QLineEdit to their respective _valueChanged slot, since the _valueChanged function is being overridden in these two instances.  Aside from that, the numerics don't touch the QLineEdit.

Note: This is John's suggested approach, but without using private inheritance due to issues with privately inheriting Q_OBJECT.

Line Edits v2 (2)

apps/vaporgui/VDoubleLineEdit.cpp
```diff
@@ -0,0 +1,79 @@
+#include <string>
+#include <sstream>
+#include <iomanip>
+
+#include <QString>
+
+#include "VDoubleLineEdit.h"
+
+VDoubleLineEdit::VDoubleLineEdit( double value, bool useMenu ) :
```
StasJ, line 9: `bool useMenu` is not used. Same for VIntLineEdit

apps/vaporgui/VDoubleLineEdit.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+VDoubleLineEdit::VDoubleLineEdit( double value, bool useMenu ) : 
+    VNumericLineEdit(),
+    _value( value )
+{
+    // Disconnect the QLineEdit's interaction with VStringLineEdit::_valueChanged(),
+    // and reconnect it to VDoubleLineEdit::_valueChanged()
+    disconnect( _lineEdit, SIGNAL( editingFinished() ),
+        this, SLOT( _valueChanged() ) );
+    connect( _lineEdit, SIGNAL( editingFinished() ),
```
StasJ, line 17: Why do you disconnect the original disconnect and reconnect it to the same slot? (In VDoubleLineEdit and VIntLineEdit constructors). Shouldn’t this be done with inheritance? Another option would be to connect VStringEdit::changed to VIntEdit::changed. Also, VStringEdit::changed ensures that the changed signal is not emitted if the user re-enters the same value. This functionality is lost for the numeric edits in the current implementation because of the above.

apps/vaporgui/VDoubleLineEdit.h
```diff
@@ -0,0 +1,40 @@
...
+//! \class VDoubleLineEdit
+//! \ingroup Public_GUI
+//! \brief A wrapper for a QLineEdit that handles user input of type double,
+//! and provides Vapor's standard setters, getters, and signals
+
+class VDoubleLineEdit : public VNumericLineEdit {
+    Q_OBJECT
+
+    public:
+        VDoubleLineEdit( double value, bool useMenu=true );
```
StasJ, line 22: Feel free to reject this but does the constructor have an initial value? It will always be replaced by the update logic once the widget is shown to the user. Qt widgets do not have initial values in the constructor for I am assuming the same reason.

apps/vaporgui/VDoubleLineEdit.cpp
```diff
@@ -0,0 +1,79 @@
...
+    if (legalConversion) {
+        SetValueDouble( value );
+        emit ValueChanged( _value );
+    }
+    else {
+        SetValueDouble( _value );
+    }*/
+}
+
+std::string VDoubleLineEdit::_formatValue( double value ) {
```
StasJ, line 68: VIntLineEdit::_formatValue and VIntDoubleEdit::_formatValue are identical and VIntLineEdit::_formatValue casts the int to a double anyway so you could put this in the parent class.

apps/vaporgui/VNumericLineEdit.h
```diff
@@ -0,0 +1,63 @@
...
+    void SetSciNotation( bool sciNotation );
+
+    //! Remove the custom context menu
+    void RemoveContextMenu();
+
+public slots:
+    
+    //! Called whenever the line edit's value is changed.  Must be reimplemented by derived classes
+    //! to handle correct formatting
+    virtual void _valueChanged() = 0;
```
StasJ, line 47: _valueChanged is declared public

apps/vaporgui/VStringLineEdit.h
```diff
@@ -0,0 +1,43 @@
...
+        VStringLineEdit( std::string value = "" );
+
+        //! Set the current string value in the line edit
+        virtual void SetValueString( std::string value );
+   
+        //! Get the current value in the line edit 
+        virtual std::string GetValueString() const;
+
+    protected:
+        QLineEdit*  _lineEdit;
```
StasJ, line 31: VStringEdit::_lineEdit should be private because of the idea it VStringEdit is the single access-point to QLineEdit and other classes shouldn’t be able to modify it as the children of VStringEdit do.

apps/vaporgui/VSection.cpp
```diff
@@ -8,7 +8,9 @@ VSection::VSection(const std::string &title)
     QTabWidget::addTab(new QWidget, QString::fromStdString(title));
     _tab()->setLayout(new QVBoxLayout);
     layout()->setMargin(12);
+    layout()->setMargin(0);
```
StasJ, line 11: This breaks the layout.

apps/vaporgui/VStringLineEdit.h
```diff
@@ -0,0 +1,43 @@
...
+        std::string _strValue;
+
+        std::string _formatValue( std::string value );
```
StasJ, line 34: This function is unimplemented.

apps/vaporgui/VStringLineEdit.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+    _lineEdit->setContextMenuPolicy( Qt::DefaultContextMenu );
+    _lineEdit->setText( QString::fromStdString( value ) );
+    _lineEdit->setToolTip( QString::fromStdString( value ) );
+}
+
+void VStringLineEdit::SetValueString( std::string value ) {
+    _strValue = value;
+    _lineEdit->setText( QString::fromStdString( _strValue ) );
+    _lineEdit->setToolTip( QString::fromStdString( _strValue ) );
```
StasJ, line 27: The text input widgets set their tooltip to their value. Doesn’t this prevent us from giving them "help" tooltips?

apps/vaporgui/VStringLineEdit.h
```diff
@@ -0,0 +1,43 @@
...
+//! and provides Vapor's standard setters, getters, and signals
+
+class VStringLineEdit : public VContainer {
+    Q_OBJECT
+
+    public:
+        VStringLineEdit( std::string value = "" );
+
+        //! Set the current string value in the line edit
+        virtual void SetValueString( std::string value );
```
StasJ, line 25: The SetString functions can be disabled for the numeric widgets with delete.

apps/vaporgui/VActions.h
```diff
@@ -0,0 +1,168 @@
...
+#include <string>
+
+#include <QWidgetAction>
+
+class VIntSpinBox;
+class VCheckBox;
+class VStringLineEdit;
+class VIntLineEdit;
+class VDoubleLineEdit;
+
```
StasJ, line 12: You define a set of VActions where every action, A, creates its corresponding VWidget, W, within a VLineItem, passing its constructor arguments onto W. It then routes W's signals and slots through duplicates in A. This could be simplified by creating a single class VLineAction

```
VLineAction(string title, VWidget *W) : inherits QWidgetAction {
	setDefaultWidget(new VLineItem(title, W));
}
```

which is then used as follows:

```
VCheckbox *check = new VCheckbox;
connect(check, ValueChanged, this, checkChanged);
menu.AddAction(new VLineAction("Title", check));
```

which is almost the same as the original use case:

```
VCheckboxAction *check = new VCheckboxAction("Title");
connect(check, ValueChanged, this, checkChanged);
menu.AddAction(check);
```

while avoiding the need for individual wrappers for every widget.

apps/vaporgui/VDoubleLineEdit.cpp
sgpearse: Thanks.  Fixed.

apps/vaporgui/VStringLineEdit.cpp
sgpearse: >Doesn’t this prevent us from giving them "help" tooltips?

No - the help tooltips will be assigned to the VLineItem, taking up the label and the space between the label and VLineEdit.  

Assigning a separate tooltip that shows the current text is helpful for showing long numbers and directory paths that aren't visible because of size constraints.

apps/vaporgui/VSection.cpp
sgpearse: Good catch - thanks.

apps/vaporgui/VNumericLineEdit.h
sgpearse: Fixed.

apps/vaporgui/VDoubleLineEdit.cpp
sgpearse: There is actually a subtle difference.

The double version just operates on doubles, plain and simple.

However the integer version casts its value to a double _only if_ scientific notation is enabled.

apps/vaporgui/VDoubleLineEdit.cpp
sgpearse: Thanks.  I incorrectly assumed that the parent's implementation of _valueChanged() would hold the signal slot connection.

apps/vaporgui/VDoubleLineEdit.h
sgpearse: I debated this too.  In some cases it's useful to provide an initial value, such as for setting up the DVR's Volume Density, or the SliceRenderer's Quality widgets.  These don't get changed on Update.  But you're right that in other cases, an initial value just gets overwritten during Update.  I think that just using an optional default value satisfies both requirements.

apps/vaporgui/VNumericLineEdit.h
sgpearse: Fixed.

apps/vaporgui/VStringLineEdit.h
sgpearse: Fixed.

apps/vaporgui/VStringLineEdit.h
sgpearse: This function is actually used in the numeric line edits to apply a formatted number to the VLineEdit.

sgpearse: Fixed.

apps/vaporgui/VActions.h
sgpearse: Fair point.

sgpearse:
Thanks for catching the overflow.  It's now guarded, and reports an error to the user.  I thought about writing a VDoubleLineEdit that shows no decimals, but that gets sticky when it comes its logic for displaying scientific notation.  This would require a significant change to the PR.  I'm also unable to think of a place in the GUI where have users input large integer values.  I believe we always use doubles when the values exceed INT_MIN or INT_MAX.

Regarding the truncation, I debated about this too.  I believe that the GUI should be a pure reflection of what's in the Params database.  I think it could be misleading if a parameter value was 1.2345e9, but the GUI showed it to be 1.2e9.

StasJ:
@sgpearse, the fact that it truncates values isn't the issue in itself. The issue is that the precision that the user can specify is dependent on whether scientific notation is enabled, e.g. if scientific notation is disabled, they can enter 1234567 but when you enable it they cannot enter a number more precise than 1200000. Whether the number is displayed in scientific notation or not should not have an effect on this.

sgpearse:
> The issue is that the precision that the user can specify is dependent on whether scientific notation is enabled

This is correct.  For integer values, scientific notation is only relevant when using scientific notation.

> if scientific notation is disabled, they can enter 1234567 but when you enable it they cannot enter a number more precise than 1200000. 

If they have set the decimal precision to 1, then yes.  I believe that this is the correct behavior.  Users can either increase the decimal precision, or just not use scientific notation in this case.

apps/vaporgui/VDoubleRangeMenu.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
+
+#include "VNumericFormatMenu.h"
+
+class VDoubleLineEditAction;
+
+//! \class VIntRangeMenu
```
clyne, line 7: Wrong documentation for VDoubleRangeMenu class?

apps/vaporgui/VStringLineEdit.h
clyne: Agreed.

apps/vaporgui/VDoubleRangeMenu.h
sgpearse: Thanks John.  Fixed.

sgpearse:
Should be good to go now.  VStringLineEdit's _lineEdit is private, documentation fixed, and example widgets removed.

Issue #617, opened 2018-04-12
Calculating Image Renderer height is very slow


StasJ:
1. Open Katrina
2. Add image renderer
3. Select USFilled.tiff as the image
4. Set height variable to HGT
5. Enable

This will take 3-6 seconds to compute. The TwoDData renderer does the same in a fraction of a second.

clyne:
Adding a comment that Jeremy Sauer and Patrick Hawbecker are running into this.

clyne:
The performance issue appears to be due to the use of Grid::GetValue(), which still accepts STL vectors as arguments. Most of the other Grid accessors have been refactored to use raw pointers.

Issue #2250, opened 2020-03-20
Refactor grids to be all 3D in nature


shaomeng:
### Description
Refactor all data representations, mostly grids, to be fixed 3D representations. When the incoming data is inherently a 2D plane, it should be represented as a 3D volume with Z dimension equals to one.

### Follow up work
Users of grids, mostly renderers, examines the special handling of 2D and 3D cases, and decide if those special handling could be eliminated. This step is on a case-by-case basis.

PR #2308, opened 2020-07-12
Issue 617 - Calculating Image Renderer height is very slow

clyne:
This PR significantly improves, but does not completely resolve the performance problems with #617. Specifically, this PR augments the API of many of the public and private Grid class methods to offer versions that use raw pointers instead of STL vectors as arguments. The latter are problematic due to the vector classes use of heap memory, which results in a performance bottleneck when called many times (e.g. inside a large loop). These changes are general and will benefit other users of the Grid class. With regard to issue #617, the resulting speedup is ~3x. 

To further improve performance specifically with regard to #617 would require changes to the ImageRenderer class to allow more intelligent sampling of the displayed image. Currently, for images displaced by a height field sampling is done in image space at the resolution of the displayed image. This should probably be so that the image sampled in the space of the height field, which would provide equivalent visual results, but require less samples.


shaomeng:
John, it looks in most of the changes you've already known the size of the array to be passed in/out. In such cases, I recommend you consider using `std::array` instead of a raw pointer. (https://en.cppreference.com/w/cpp/container/array)
A few benefits of the container type over raw pointers: 
1. The length of the array is *strictly* enforced;
2. It plays well with all the STL algorithms, including the use of iterators, etc. 
3. It avoids all the pitfalls associated with the use of raw pointers.

clyne:
You're probably right. I'll have to play around with it and see how it performs. Maybe next weekend :-)

shaomeng:
> You're probably right. I'll have to play around with it and see how it performs. Maybe next weekend :-)

Oh, it'll be at least as fast as raw pointers. With stronger guarantees on memory bounds and overlapping and stuff like that, it'll be even faster in some cases.

shaomeng:
> You're probably right. I'll have to play around with it and see how it performs. Maybe next weekend :-)

Oh btw, when you play around with it, remember to set up something like this to reduce your typing: 
`using coord3 = std::array<float, 3>; using coord6 = std::array<float, 6>;`

clyne:
@StasJ , @shaomeng, I've converted most of the Grid API to use sdt::array for passing indices and coordinates. The biggest implication of this change is that std::array is a fixed length (3 elements). Many places in the code assume (wrongly) that the length of indices or coordinate vectors/arrays will match the dimensionality of the grid. To minimize the number of changes in the application I've preserved the old, slower performing std::vector interfaces. I've only migrated the application to the new std::array interface in places where performance matters. Another big todo is to refactor the Grid iterators. They are still using the std::vector interface, but in a way that typically does not impact performance. The only real motivation to address the iterators is for consistency. I think the time to address the iterators is when we address #2250

PR #2315, opened 2020-07-22
Progress Status Bar

StasJ:
Partly because of how Qt is implemented and partly because of how we use Qt, this code commits several atrocities. The consequences are then mitigated by more atrocities. In the end, however, it converges into something that seems to work.

The main issue is that the only way to get this to work is to manually processes Qt's event loop recursively. I later discovered that Qt does this itself on a few occasions (the Qt popup progress bar in the other PR works this way) however it has Band-Aid code in its backend implementation that prevents issues. 

If it ends up being problematic it can all be disabled with one line of code. There is a way to do this without hacks, however the amount of refactoring required would warrant Vapor 4.

With regards to performance, it limits the number of updates to either 100 per calculation or a maximum of 10 progress updates per second, whichever is lower. In my testing, showing the progress bar adds about 5-10% overhead to the computation time when doing a simple calculation with many updates. This would be lower for a heavier calculation with a higher ratio of calculation to progress updates of course. Also, the progress bar is only updated when the user is actively using the application. When rendering a sequence of timesteps, it is disabled.

This PR implements progress bars for the Wireframe, Flow, and Volume renderers.

Some pitfalls:

It sometimes breaks the Fidelity radio buttons but those were always somewhat glitchy (they disappear and reappear on updates) and they are being replaced soon (from my understanding) so I didn't look into fixing it.

I almost had this working for everything, i.e. it would show progress for loading data from disk, populating histograms, etc. however I could not get Qt to behave consistently when doing this and this project was already taking way too long. It would sometimes correctly process the updates and sometimes it would claim to process them but not do anything. Therefore the progress bar is limited to code within the visualizer rendering which works consistently for some reason.

Progress

### DIFF
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -84,6 +84,10 @@
 #include "MouseModeParams.h"
 #include "ParamsWidgetDemo.h"
 
+#include <QProgressDialog>
+#include <QOpenGLContext>
+#include <vapor/Progress.h>
+
 //Following shortcuts are provided:
 // CTRL_N: new session
 // CTRL_O: open session
@@ -371,7 +375,9 @@ MainForm::MainForm(
 
     createMenus();
 
-	createToolBars();	
+	createToolBars();
+    
+    _createProgressWidget();
 	
 	addMouseModes();
     (void)statusBar();
@@ -719,6 +725,76 @@ void MainForm::createToolBars(){
 	_createAnimationToolBar();
 	_createVizToolBar();
 	
+}
+
+void MainForm::_createProgressWidget()
+{
+    if (_progressEnabled)
+        return;
+    
+    // The modal version adds an animation that takes a little less than
+    // a second however on OSX the non-modal version is broken.
+    
+    if (!_progressDialog) {
+        _progressDialog = new QProgressDialog("Title", "Cancel", 0, 100, this);
+        _progressDialog->close();
+        _progressDialog->setWindowModality(Qt::WindowModal);
+        _progressDialog->setAutoClose(false);
+        _progressDialog->setMinimumDuration(1500);
+    }
+    
+    Progress::Start_t start = [this](const std::string &name, long total, bool cancelable)
+    {
+        if (QOpenGLContext::currentContext() && _progressSavedFB < 0) {
+            glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &_progressSavedFB);
+            glBindFramebuffer(GL_FRAMEBUFFER, 0);
+        }
+        
+        _progressDialog->reset();
+        if (cancelable)
+            _progressDialog->setCancelButtonText(QString("Cancel"));
+        else
+            _progressDialog->setCancelButton(nullptr);
+        _progressDialog->setLabelText(QString::fromStdString(name));
+        _progressDialog->setValue(0);
+        _progressDialog->setMaximum(total);
+    };
+    
+    Progress::Finish_t finish = [this]()
+    {
+        _progressDialog->close();
+        
+        if (QOpenGLContext::currentContext() && _progressSavedFB >= 0) {
+            glBindFramebuffer(GL_FRAMEBUFFER, _progressSavedFB);
+            _progressSavedFB = -1;
+        }
+    };
+    
+    Progress::Update_t update = [this, finish](long done, bool *cancelled)
+    {
+        _progressDialog->setValue(done);
+        if (_progressDialog->wasCanceled()) {
+            *cancelled = true;
+            finish();
+        }
+    };
+    
+    Progress::SetHandlers(start, update, finish);
+    _progressEnabled = true;
+    if (_progressEnabledMenuItem)
+        _progressEnabledMenuItem->setChecked(true);
+}
+
+void MainForm::_disableProgressWidget()
+{
+    if (!_progressEnabled)
+        return;
+    
+    Progress::Finish();
+    Progress::SetHandlers([](const string&, long, bool){}, [](long,bool*){}, [](){});
+    _progressEnabled = false;
+    if (_progressEnabledMenuItem)
+        _progressEnabledMenuItem->setChecked(false);
 }
 
 void MainForm::hookupSignals() {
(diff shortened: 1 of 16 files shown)


clyne:
@stasj this is a really nice usability feature, but I have a couple of questions before starting a review:

1. You've submitted two PRs with alternate implementations of a progress bar; this one and #2316. What is the intent here? 

2. You've made numerous caveats in the comments above concerning the robustness of this implementation. How confident are you in this code? Is it going to impact the reliability of the application?

StasJ:
@clyne 

1. I just included the other PR for comparison. It uses Qt's built in functionality to show a progress bar without running the GUI on a separate thread. It does not require as many hacks as this code to work but it also does not work as well.
2. I tested many potential use cases to make sure everything worked reliably before submitting the PR, however, I cannot guarantee it won't cause any issues since it does so many unorthodox things. If it ends up causing issues, there is a single flag that will disable all the code this PR introduces. I also added a checkbox to Vapor's developer menu that allows you to enable/disable it on the fly.

clyne:
@StasJ please verify that this compiles and runs on windows before we proceed further with a review. Thanks.

StasJ:
@clyne I verified that this works on Windows 8.

include/vapor/Progress.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+#include <string>
+#include <functional>
+#include <vapor/common.h>
+
+namespace VAPoR {
+class COMMON_API Progress {
```
clyne, line 7: Please document this class and methods

apps/vaporgui/MainForm.cpp
```diff
@@ -1712,12 +1885,20 @@ void MainForm::_setAnimationOnOff(bool on) {
 		enableAnimationWidgets(false);
 
 		_App->removeEventFilter(this);
+//        if (_progressEnabled) {
```
clyne, line 1888: If this code might be needed later, please ifdef and provide a comment. Otherwise delete.

apps/vaporgui/MainForm.cpp
```diff
@@ -1712,12 +1885,20 @@ void MainForm::_setAnimationOnOff(bool on) {
...
+//            _disableProgressWidget(); // Need to disable as no longer have event filter
+//            _needToReenableProgressAfterAnimation = true;
+//        }
 	}
 	else  {
 		_playForwardAction->setChecked(false);
 		_playBackwardAction->setChecked(false);
 		enableAnimationWidgets(true);
 		_App->installEventFilter(this);
+//        if (_needToReenableProgressAfterAnimation) {
```
clyne, line 1898: Ditto for this commented out code.

apps/vaporgui/VizWinMgr.cpp
```diff
@@ -345,11 +345,24 @@ void VizWinMgr::_vizAboutToDisappear(string vizName)  {
 
 
 void VizWinMgr::Update(bool fast){
-
+    // Certain actions queue multiple renders within Qt's event queue (e.g. changing the
```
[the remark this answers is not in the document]
StasJ: The way we have it set up, the render event processing is always done asynchronously from our perspective but synchronously from Qt's perspective. The can be changed to be synchronously from our perspective too but I don't know if that would be a good idea. Qt processes render events one at a time from the event loop, however the event loop can be triggered recursively when updating the progress bar, so if there are multiple render events queued, render can be called recursively. There is no method of preventing this that Qt exposes through its API AFAIK, but this check here prevents any issues. My other PR actually prevents multiple render events from being queued, however I think it would be a good idea to keep this check anyway as a failsafe.

StasJ:
@sgpearse I built on Windows 10. You need to update your VisualC++ compiler.

Great, I'll take a look at that warning, thanks.

apps/vaporgui/VizWin.cpp
```diff
@@ -647,6 +647,14 @@ void VizWin::setFocus(){
 }
 
 void VizWin::Render(bool fast) {
+    if (_insideRender)
```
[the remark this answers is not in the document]
StasJ: Done

include/vapor/Progress.h
StasJ: Done

apps/vaporgui/MainForm.cpp
StasJ: Done

apps/vaporgui/MainForm.cpp
StasJ: Done

clyne:
@sgpearse I received email stating: 

"Unfortunately we can't merge this PR until we've upgraded our windows compiler. This involves updating our CI system, and rebuilding our third party libraries."

However, the comment above states only:

"Unfortunately we can't merge this PR until we've upgraded our windows compiler."

Would we need to rebuild out 3rd party libs or not? If so, how much effort is this?

Alternatively, @StasJ can we simply not use the problematic language feature?

StasJ:
@clyne We shouldn't have to rebuild our third party libraries. It is possible to reimplement the offending code however it is already using pretty old features (C++11). The windows compiler Circle is using is 5 years old so we probably want to upgrade it anyway.

clyne:
> @clyne We shouldn't have to rebuild our third party libraries. It is possible to reimplement the offending code however it is already using pretty old features (C++11). The windows compiler Circle is using is 5 years old so we probably want to upgrade it anyway.

@sgpearse what kind of effort is it going to take to upgrade the windows compilers?

sgpearse:
@clyne I am uncertain about library compatibility.  I originally thought we would need to re-built the libraries, but I retracted this statement.  [This article](http://mingw.org/wiki/Interoperability_of_Libraries_Created_by_Different_Compiler_Brands#:~:text=Sometimes%20you%20can%20link%20a,runtime%20library%20provided%20with%20Windows.) talks about different schemes for mangling symbol names, so we might be ok, but there's a lot of uncertainty.  If we go down this route, I'd personally hedge on being safe and rebuilding the libraries.

When the libraries were rebuilt last year, Windows was the most effort by far.  That was partly due to the build process being undocumented.  However there are still inconveniences in the procedure.  Several build systems were used, including the use of pre-built binaries that were built with compilers currently unknown to us.

clyne:
@sgpearse my experience with libraries compiled under older compilers is that either they link or they don't. Have you tried to see if they'll link? It doesn't seem like it should be terribly difficult to test this. Thanks.

clyne:
Scott wrapping up testing with VS19. Will notify team when know outcome of using VS19 with extant 3rd party library builds

sgpearse:
Building on VS19 works when building on local hardware.  CircleCI's windows build system is currently broken, but the problem does not seem to be the compiler version being used.

PR #2323, opened 2020-07-27
Issue2241 minus flow and barbs

sgpearse:
This PR implements the Variables tab for all renderers except for Barbs, Flow, and WireFrame.  These excluded renderers require a variable dimensionality dropdown that is currently buggy.  The other renderers test well and this is a good point to submit a review.

Some notes:

1) Many of the new classes are undocumented.  I'm waiting to do in case things change.

2) I've added new functions to the RendererEventRouter that determine support for 2d/3d variables.  This was previously done with flags in the VariablesWidget.  This was done in the RenderEventRouter instead of RenderParams due to RenderParams not yet being initialized when the GUI requests 2d/3d variable support.

3) New PWidget and HLI classes have been added to a file called PVariableWidgets.* instead of their own individual files.

4) PVariablesGroup - A new class that is a derivation from PGroup that lets us add tabs to our event routers in the same fashion as we've done within our EventRouters in the past.  This may become unnecessary as we migrate all of our tabs to PWidgets.

5) The logic for the SliceRenderer's sampling rate has been moved from SliceSubtabs to the SliceEventRouter.

6) PGroup now needs to call the addStretch() upon construction to size it self appropriately.  This effectively adds a QSpacerItem to the layout.  Widgets that are subsequently added via addWidget() will be placed below the spacer, instead of on top of it.  Consolodating widgets to the top of the control panel is preferable, so insertWidget() is used to do this.  Without calling addStretch(), the 'Variable Selection" and "Data Fidelity" groups will expand to fill the entire control panel


include/vapor/RenderParams.h
```diff
@@ -156,12 +156,20 @@ class PARAMS_API RenderParams : public ParamsBase {
 	//
 	virtual void SetFieldVariableNames(vector <string> varNames);
 
+    void SetXFieldVariableName( std::string varName );
```
clyne, line 159: Doxygen document

include/vapor/RenderParams.h
```diff
@@ -156,12 +156,20 @@ class PARAMS_API RenderParams : public ParamsBase {
...
+    void SetYFieldVariableName( std::string varName );
+    void SetZFieldVariableName( std::string varName );
+
 	//! Get the field variable names, e.g. used in flow integration.
 	//! \retval vector<string> variable names. A vector of length 3 
 	//! containing variable names. The default is 3 empty variable names.
 	//
 	vector<string> GetFieldVariableNames() const ;
 
+    std::string GetXFieldVariableName() const;
```
clyne, line 169: doxygen document

lib/params/RenderParams.cpp
```diff
@@ -497,6 +497,39 @@ void RenderParams::SetFieldVariableNames(vector<string> varnames){
 //	setAllBypass(false);
 }
 
+std::string RenderParams::GetXFieldVariableName() const {
+    std::vector< std::string> fieldVars = GetFieldVariableNames();
```
clyne, line 501: It would be a good idea to assert in all of these setters and getters that fieldVars is of sufficient length before indexing into it.

sgpearse:
@clyne In regard to documentation - I'd like to reserve this until later, in case we find that the classes need to be changed.

@stasj - Sure.  In the ptest branch, when the Flow renderer is set to 2D, we render 5 pathlines above each seedpoint, shown in the image below.  However, the dim selector is interacting with the Box class in this instance.  We are intending to use the dimensionality of certain variables instead.

It gets tricky when we want to determine which "certain variable" we should be using to determine dimensionality.  I spent some time on this in the branch issue2241_dimSelector.  The PDimensionSelector works for the Flow renderer in this branch, but not the wireframe or barb renderer.  

Where I left off, I was getting the sense that we had to dynamic_cast the PDimensionSelector's RenderParams to find out what variable we should query in the _updateGUI() function.  At that point, I felt this PR was large enough to submit for review, and that we could discuss the PDimensionSelector next.

StasJ:
@sgpearse I checked out the ptest branch and the Barb, Flow, and Wireframe renderers all work correctly (well the flow renderer doesn't but that is because of a bug in the flow renderer which exists on master; the variable selector works correctly though).

clyne:
@StasJ please provide review comments regarding class design concerns. We will address dimensionality selector bug after we ok the class design for variable selectors

apps/vaporgui/PFidelitySection.h
```diff
@@ -0,0 +1,86 @@
...
+
+protected:
+    virtual void updateGUI() const override;
+    bool requireDataMgr() const override { return true; }
+    
+private slots:
+    void dropdownIndexChanged(int i);
+};
+
+class PLODSelectorHLI :
```
clyne, line 58: This appears to be dead code. This class isn't used anywhere. Note that PLODSelector::dropdownIndexChanged() is already using the Params HLI

apps/vaporgui/PFidelitySection.h
```diff
@@ -0,0 +1,86 @@
...
+        PLODSelector(),
+        PWidgetHLIBase<VAPoR::RenderParams, double> (
+            (PWidget*) this,
+            &VAPoR::RenderParams::GetCompressionLevel,
+            &VAPoR::RenderParams::SetCompressionLevel
+        )
+    {}
+};
+
+class PRefinementSelectorHLI :
```
clyne, line 73: This also appears to be an unused (and not needed) class.

apps/vaporgui/PGroup.cpp
```diff
@@ -8,6 +8,7 @@ PGroup::PGroup()
 {
     QVBoxLayout *layout = new QVBoxLayout;
     layout->setMargin(0);
+    layout->addStretch();
```
clyne, line 11: The PGroup widget has had layout options added. I don’t believe PWidgets should perform layout options, correct?

sgpearse: Unfortunately PWidgets are QWidgets themselves, so they do need to perform some layout management.  The addStretch() function handles resizing operations on the control panel.

include/vapor/RenderParams.h
sgpearse: Done

include/vapor/RenderParams.h
sgpearse: Done

lib/params/RenderParams.cpp
sgpearse: Done

apps/vaporgui/PFidelitySection.h
sgpearse: Removed

apps/vaporgui/PFidelitySection.h
sgpearse: Removed

clyne (commented):
LGTM. Just a couple of nits in-line. 

We still need to address the renderers that support 2D and 3D, correct?

StasJ (commented):
I think John's comment on PGroup adding spacers to the layout still needs to be addressed but I don't have any new comments.

apps/vaporgui/PFidelitySection.cpp
```diff
@@ -0,0 +1,156 @@
...
+    auto varName = rp->GetFirstVariableName();
+    auto dm = getDataMgr();
+    int nrf = dm->GetNumRefLevels(varName);
+    vector<string> items;
+    
+    for (int i = 0; i < nrf; i++) {
+        vector<size_t> dims;
+        dm->GetDimLensAtLevel(varName, i, dims);
+        if (dims.empty()) continue;
+        auto itr = dims.begin();
```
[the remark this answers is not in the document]
sgpearse: I don't know.  This is a question for @StasJ

sgpearse:
@stasj and @clyne,

I took a look at why PGroups are able to correctly display in our Appearance tabs.  The reason is that the Appearance tabs are still operating on UI file based GUI's.  

These UI files have spacers in them, which is effectively the same as calling the addStretch() function.  

If we don't do this in PGroup, all of our tabs will expand to take up the unused vertical space in our control panel.

apps/vaporgui/PVariableWidgets.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+PDimensionSelector::PDimensionSelector()
+: PLineItem("", "Variable Dimension", _vComboBox = new VComboBox({"2D", "3D"}))
+{
+    connect(_vComboBox, &VComboBox::ValueChanged, this, &PDimensionSelector::dropdownTextChanged);
+}
+
+void PDimensionSelector::updateGUI() const
+{
+    /*RenderParams *rp = dynamic_cast<RenderParams*>(getParams());
```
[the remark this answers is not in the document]
sgpearse: I'd like to keep this commented code for reference in the next task, which is to implement the dimension selector for Barbs, Flow, and the Wireframe renderers.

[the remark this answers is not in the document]
clyne: Please use ifdef's with a comment, not old style, multi-line C /**/ comments. They are too easy to miss.

clyne:
> @StasJ and @clyne,
> 
> I took a look at why PGroups are able to correctly display in our Appearance tabs. The reason is that the Appearance tabs are still operating on UI file based GUI's.
> 
> These UI files have spacers in them, which is effectively the same as calling the addStretch() function.
> 
> If we don't do this in PGroup, all of our tabs will expand to take up the unused vertical space in our control panel.

I don't have a deep enough understanding of the issue to comment. @stasj what are your thoughts?

apps/vaporgui/PFidelitySection.cpp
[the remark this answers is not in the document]
StasJ: It is used in the next line before the loop starts.

StasJ:
> > @StasJ and @clyne,
> > I took a look at why PGroups are able to correctly display in our Appearance tabs. The reason is that the Appearance tabs are still operating on UI file based GUI's.
> > These UI files have spacers in them, which is effectively the same as calling the addStretch() function.
> > If we don't do this in PGroup, all of our tabs will expand to take up the unused vertical space in our control panel.
> 
> I don't have a deep enough understanding of the issue to comment. @StasJ what are your thoughts?

@sgpearse It appears that spacers are being automatically added to all PGroups to support the case where a PGroup is used to replace a subtab. This may not be desired in most use cases in which it is not the top level subtab widget. To address this, we can instead have a PSubtab class which has the properly configured layout to act as a subtab (or rather inherits from a VSubtab which does so). Another option is to create a PSpacer which can be added to the end of a PGroup as necessary. Either way since PVariablesGroup is no longer a part of this PR, the solution can be addressed in a later PR.

sgpearse:
I've #ifdef'd commented code, and removed the addStretch() function from PGroup.  

I have some followup questions that don't need an immediate response.

>This may not be desired in most use cases in which it is not the top level subtab widget.

I'm trying to think of an example where it's not desirable.  Can you share one?  

The Python editor, Statistics, and Plot utilities don't use subtabs, but will also want to push their rows to the top of a given PGroup.  I don't believe we will ever want to distribute a PGroup's rows evenly within its vertical area.

>To address this, we can instead have a PSubtab class which has the properly configured layout to act as a subtab (or rather inherits from a VSubtab which does so).  

I believe this introduces the diamond problem.  PSubtab would inherit from VSubtab and PWidget, which both inherit from QWidget.

StasJ (approved):

Issue #2309, opened 2020-07-14
Automate code formatting


clyne:
Automate the formatting of C++ code using a code formatter such as clang-format, and an appropriate Git/GitHub pre-commit mechanism.

PR #2325, opened 2020-07-28
Automate code formatting #2309

sgpearse:
This PR is for us to review the checks we'd like to adopt in our pre-commit hook, as per #2309 .

The checks in this PR include running clang-format under LLVM's style guide, with overrides to column width (we use 100 instead of 80), tab width (we use 4 instead of 3), and indentation width (we use 4 instead of 2).

This also has checks for detecting non-ascii characters in filenames, and whitespace errors.  These are done within the hook file itself (named pre-commit), not through clang-format.


clyne:
@sgpearse please explore ways of putting this in place that won't disrupt the logging history (e.g. use of git blame) before we review.

clyne:
@stasj will test a method for  making formatting changes in a way that will preserve logging. After we're convinced the method works as hoped we will make the actual change.

clyne:
Update: Scott has updated formatter config so doesn't break compilation. Needs to verify and then will have Stas rerun his script

clyne:
Ready for @stasj to format entire repo

clyne:
@stasj will do a dry run because of concerns over disconnect between git and github

clyne:
We will table this  until after the 3.3 release

PR #2331, opened 2020-07-31
Flow Density Rendering

StasJ:
Implemented the technique in [Trajectory Density Projection for Vector Field Visualization](https://www.researchgate.net/publication/261329939_Trajectory_Density_Projection_for_Vector_Field_Visualization).

After doing so, however, I do not know if it is something we want to add.
- The rendering method means it won't work well (if at all) combined with other renderers
- It requires many particles to look good (from my testing)
- It introduces additional colors outside the colormap
- It can be finicky to get it looking right
- Its appearance is dependent on the background color (i.e. no matter the color of a flow line, it will be completely transparent when passing over a light background. The inverted option fixes this but then it inverts the colors. If you have changing background colors, (e.g. with other renderers present) this will not work).

That being said, as can be seen in the paper, there are scenarios where this technique can be useful.

### Tube Rendering
b

### Tube Rendering With Transparency
a

### Density Rendering
c

### Another Density Example
d

### DIFF
--- a/apps/vaporgui/CMakeLists.txt
+++ b/apps/vaporgui/CMakeLists.txt
@@ -172,6 +172,8 @@ set (SRCS
 	PDisplay.cpp
 	PDisplay.h
 	PDisplayHLI.h
+	PLabel.cpp
+	PLabel.h
 	PLineItem.cpp
 	PLineItem.h
 	PEnumDropdown.cpp
--- a/apps/vaporgui/FlowSubtabs.cpp
+++ b/apps/vaporgui/FlowSubtabs.cpp
@@ -138,7 +138,7 @@ FlowAppearanceSubtab::FlowAppearanceSubtab(QWidget* parent) : QVaporSubtab(paren
     PSection *ps;
     
     _pw->Add(ps = new PSection("Appearance"));
-    ps->Add(new PEnumDropdown(FlowParams::RenderTypeTag, {"Tubes", "Samples", "Density"}, {FlowParams::RenderTypeStream, FlowParams::RenderTypeSamples, FlowParams::RenderTypeDensity}, "Render Type"));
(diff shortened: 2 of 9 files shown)


clyne:
Team needs to review for functionality

clyne:
@StasJ I think the density renderer may have value, and given that it has a pretty small impact to the code I think we should include it. A couple of comments on the GUI/usability:

1. I don't think offering the density render as a "Render type" is the right place to put the selector (along with tubes and samples). To me the latter are display primitives - not really render types - and "density" is maybe a blending option, or different algorithm? 

2. I'm really not sure what "density falloff" and "tone mapping" are. Are they both needed?

3. It's not clear to me how the opacity selector interacts with the above two controls. Manipulating opacity seems to give similar results to changing density falloff and tone mapping. Can you elaborate on all three of these ?

StasJ:
@clyne 

1. I disagree that the tubes vs samples are just primitive types since they represent different things and show different types of information. In the case of the density rendering, it is not just a blending option because you need special primitives to make it work. The "type" also determines the GUI options shown to the user; the lines and samples have a different GUI, along with the density rendering which has a different GUI.
2. /3 They are the parameters the paper defines. Falloff scales the rate of fading of intensity along the width of the line. Tone exponentially scales the overall color intensity. There is a more detailed explanation in the paper. Opacity does not apply to the technique but it already exists in the renderer.

clyne:
Add help messsaes to sliders and a refernce to the paper

sgpearse (approved):

clyne (commented):
Nice work, @stasj . Please open an "enhancement" issue and link this PR to it. Closed issues is how we track development, test adequately, etc.

StasJ:
Fix #2364

clyne (approved):

PR #2339, opened 2020-08-13
Improved unique ptr cache

shaomeng:
This PR improves `unique_ptr_cache` so that 
1) its performance is a lot better, and
2) it's thread-safe.

### DIFF
--- a/include/vapor/unique_ptr_cache.hpp
+++ b/include/vapor/unique_ptr_cache.hpp
@@ -9,19 +9,27 @@
 // All structures stored in this cache are const qualified, so once a
 // structure is put in this cache, there is no more modification to this structure.
 //
-// This implementation follows std container naming conventions.
+// Caveat: This cache guarantees that a returned non-null pointer is valid at the
+//         query time, but cannot guarantee that the pointer keeps being valid after
+//         handing it over (i.e., not being evicted). Thus, this cache should be initialized 
+//         sufficiently big so that a returned pointer won't be evicted while that pointer
+//         is in use.
+//
+// Revision: (8/13/2020) it uses std::array<> to achieve the highest performance with
+//                       small to medium cache sizes.
+// Revision: (8/13/2020) it uses mutexes to achieve thread safety.
 //
 // Author: Samuel Li
 // Date  : 9/26/2019
+// Revision: 8/13/2020
 //-----------------------------------------------------------------------------
 
 #ifndef UNIQUE_PTR_CACHE_H
 #define UNIQUE_PTR_CACHE_H
 
-#include <cstddef>
-#include <list>
-#include <utility>
-#include <memory>
+#include <cstddef>  // size_t
+#include <utility>  // std::pair<>
+#include <memory>   // std::unique_ptr<>
 #include <array>
 #include <mutex>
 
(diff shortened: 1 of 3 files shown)


include/vapor/unique_ptr_cache.hpp
```diff
@@ -9,130 +9,135 @@
 // All structures stored in this cache are const qualified, so once a
 // structure is put in this cache, there is no more modification to this structure.
 //
-// This implementation follows std container naming conventions.
+// Caveat: This cache guarantees that a returned non-null pointer is valid at the
```
clyne, line 12: Hmm. Would a shared_ptr be a better choice so that the memory doesn't disappear out from under someone? I.e. Couldn't we remove this caveat by using shared pointers instead of unique pointers

include/vapor/unique_ptr_cache.hpp
```diff
@@ -9,130 +9,135 @@
...
     }
 
 private:
-    using list_type = std::list< std::pair<Key, std::unique_ptr<const BigObj>> >;
-    mutable list_type   m_list;
-    const   size_t      m_max_size;
+    using element_type = std::pair<Key, std::unique_ptr<const BigObj>>;
 
-};  // end of class unique_ptr_cache
+    std::array< element_type, CacheCapacity >   element_array;
```
clyne, line 135: These member variables should follow naming conventions (precede with an "_").

include/vapor/unique_ptr_cache.hpp
shaomeng: Well, in that case, the cache will have no control of how many objects are "alive." I.e., the cache could be holding 10 objects itself, but all objects that used to live in the cache and then got evicted are still "alive" and consuming system resources. In this scenario, you lose control of system resources, and the caching mechanism becomes meaningless. 

Neither of the scenarios (my caveat and evicted objects keep being alive) are ideal. But an undefined object lifespan is a major drawback associated with the use of raw pointers. Balancing the two, I feel my caveat is more acceptable. Thoughts?

shaomeng: Actually, a cache does *not* have the responsibility to manage the lifespan of objects out of the scope of the cache itself. For *any* cache, it is fair to *only* provide guarantee on the validity of objects at the time of query, and it's not quite reasonable to ask for more from a cache.

clyne: If the goal were to place a hard limit on the amount of memory used, then I agree giving ownership of memory to the "client" would not work as implemented. However, I'm not sure that a *hard* limit is a goal. Virtual memory works pretty well on most platforms :-)  Even if it were a goal, wouldn't it be better to make use of the shared_ptr.use_count() method and simply have the cache insert() fail if the cache was full and could not be purged? clients could detect and recover from this more easily than having pointers that might at some nondeterministic point become NULL. 

This is really very much an academic discussion because converting to  shared_ptrs would be a major change, and I want to make sure Abhishek can proceed. So with the proposed changes in this PR is it possible for VaporFlow to overrun the cache, or is it being sized in a way that this is not possible (regardless of the number of threads, etc).

include/vapor/unique_ptr_cache.hpp
shaomeng: This is addressed.

include/vapor/unique_ptr_cache.hpp
shaomeng: As we discussed in the meeting, the current implementation does the job of a cache. We might need more than what a cache is able to provide, but fulfilling those requirements would imply major code changes in other places. The bottom line is that Abhishek is now able to proceed, and this cache implementation satisfies one use case where it is deployed.

clyne: Most importantly: the cache size needs for the one user of this class, VaporFlow, are known a priori

PR #2340, opened 2020-08-13
Improved unique ptr cache documentation

shaomeng:


shaomeng:
John, I blieve this improved document makes the point very clear. Please take a look and consider merging it. Thanks! @clyne

PR #2344, opened 2020-08-21
Issue 2343 - Provide API to determine renderer dimension

### LINKED ISSUE #2343: Provide API to determine renderer dimension
Some renderers can operate on either 2D or 3D variables. API is needed to determine what dimension variable they are currently acting on to facilitate update the GUI to present the correct list of variable choices.

clyne:
Closes #2343

### DIFF
--- a/apps/vaporgui/PlotParams.h
+++ b/apps/vaporgui/PlotParams.h
@@ -65,6 +65,9 @@ public:
     {
         return("PlotParams");
     }
+    virtual size_t GetRenderDim() const override {
+        return(0);
+    }
 
 private:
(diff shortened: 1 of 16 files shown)


sgpearse:
The Flow and Barb renderers use their X field variable to determine dimensionality, but the X field variable is sometimes not present.  @clyne could this be a problem?

clyne:
> The Flow and Barb renderers use their X field variable to determine dimensionality, but the X field variable is sometimes not present. @clyne could this be a problem?

Good point. I've addressed it.

clyne:
@sgpearse please rereview

PR #2347, opened 2020-08-24
Particle advection refactor and OMP parallelization

ayenpure:
Updating particle advection code to loop over particles as a work unit.
The changes allow parallelization using libraries that support compiler directives (OpenMP, etc.)

The parallelism was tested on Casper using `gcc 7.4.0`
The requirements to use parallelism with OpenMP is to add the `-fopenmp` flag to your CMake compiler flags
`cmake -DCXX_COMPILER_FLAGS=-fopenmp`

In case the compiler does not support OpenMP the compiler directives for parallelism are ignored and the program compiles as if there were no directives.


ayenpure:
@shaomeng @clyne Can you review the particle advection related changes?

@StasJ I need to make sure how the Progress stuff works while using multiple threads. I have it commented for now.

test_apps/advection/README.md
```diff
@@ -0,0 +1,56 @@
...
+      .
+```
+
+## Running benchmarks
+
+The program requires an input config file for all the required parameters.
+A smaple file `demo10K.ini` is included in this repository.
+
+```
+datapath=/glade/u/home/abhisheky/NetCDF
```
clyne, line 34: It looks like the data set you have in your home directory is one of our canonical test data sets from /glade/p/cisl/vast/vapor/data/Source/CM1/Orf/24Maycontrol 

If so, can you update the README to use the canonical version (it will have more permanency than your home directory).

test_apps/advection/README.md
```diff
@@ -0,0 +1,56 @@
+# VAPOR Particle Advection Benchmarking
```
clyne, line 1: Can you provide the command invocation that is needed to run the benchmark program?


Also, can you include the results from the benchmark (either in this readme or a separate file). The results should have all of the pertinent information such as:  the parameters used to configure the benchmark, the host name where the benchmark was run, and any other information that might be needed to recreate the results.

test_apps/advection/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+VAPOR_PREFIX : path to the VAPOR build/installation
+VAPOR_INCLUDE_DIR : path to the include directory in the VAPOR repository
+VAPORDEPS_PREFIX : path to where the VAPOR dependencies are located
+```
+A sample CMake command to build this project would look like
+
+```
+cmake -DVAPOR_PREFIX=/Users/abhishek/repositories/VAPOR/build \
```
clyne, line 21: These macros point to files in your home directory. Is there anything special about these? I.e. have they been modified from the original?

clyne, line 21: How are the number of threads controlled?

test_apps/advection/README.md
ayenpure: Ok, I'll update the path to the one already in the repository.

test_apps/advection/README.md
ayenpure: Ok, I'll update all that information too.

test_apps/advection/README.md
ayenpure: There's nothing special about the VAPOR build as such.
But I did not want to add building of this application with the test_apps, I've used a couple of Boost modules which are not available with VAPOR CMake stuff after being built.

ayenpure: I'll add it to the `Running benchmarks` section

clyne:
One other thing: the current branch won't build. It looks like you may need to sync with master.

ayenpure:
> One other thing: the current branch won't build. It looks like you may need to sync with master.

So I verified the branch builds on my MacBook and on Casper, I don't know what CircleCI is complaining about.

lib/flow/Advection.cpp
```diff
@@ -42,14 +42,108 @@ Advection::CheckReady() const
...
+                {
+                    if( _isPeriodic[i] )
+                    {
+                        loc[i] = _applyPeriodic( loc[i], _periodicBounds[i][0], _periodicBounds[i][1] );
+                        locChanged = true;
+                    }
+                }
+
+                if( !locChanged )   // no dimension is periodic
+                    break;       // skip this particle, since it's out of the volume
```
shaomeng, line 79: Which loop does this break statement break?

test_apps/advection/Advection.cpp
```diff
@@ -0,0 +1,705 @@
...
+                {
+                    if( _isPeriodic[i] )
+                    {
+                        loc[i] = _applyPeriodic( loc[i], _periodicBounds[i][0], _periodicBounds[i][1] );
+                        locChanged = true;
+                    }
+                }
+
+                if( !locChanged )   // no dimension is periodic
+                    break;       // skip this particle, since it's out of the volume
```
shaomeng, line 80: I'm not sure why this line shows up again seemingly the same file... But still, is this break breaking an intended loop ?

clyne:
@ayenpure here is the compiler message I get when I try to build this branch on my Mac. 

> One other thing: the current branch won't build. It looks like you may need to sync with master.

I resolved the library build issue. However, the test driver itself will not build. Can you remove the boost dependency and make whatever changes are needed in CMakeLists.txt to enable compiling? Thanks!

clyne:
Abhishek will remove files from the benchmark directory that duplicate stuff in lib/flow

clyne:
@ayenpure is this PR ready for re-review, or are you still working on it?

ayenpure:
I think it’s ready. I’ve fixed everything we had planned on fixing.

-Abhishek

test_apps/advection/benchmark.cpp
```diff
@@ -0,0 +1,217 @@
...
+  // TODO : read field and variable information
+  // Let's assume we have read the data and field
+  const std::string filetype = "cf";
+  const size_t cache = 2000;
+  const size_t threads = 0;
+
+  std::vector<std::string> files;
+  files.push_back(datapath);
+  std::vector<std::string> fileopts;
+  //fileopts.push_back("-project_to_pcs");
```
clyne, line 148: please remove commented out code

test_apps/advection/benchmark.cpp
```diff
@@ -0,0 +1,217 @@
...
+  {
+    std::cerr << "Failed to intialize CF DataMGR" << std::endl;
+    exit(EXIT_FAILURE);
+  }
+  std::vector<double> timeCoords;
+  datamgr.GetTimeCoordinates(timeCoords);
+  double initTime = timeCoords.at(0);
+  // Create particles
+  std::vector<flow::Particle> seeds;
+  std::vector<double> xrake;
```
clyne, line 162: So the user doesn't have to determine suitable extents, how about by default you use the overloaded DataMgrUtils::GetExtents() function to get the intersection of all three variable extents? If the user specifies -minu or -maxu it can override the default which would be the entire domain?

ayenpure: Fixed

test_apps/advection/benchmark.cpp
ayenpure: I've left some comments which would make it easier to test the validity of the output, but mostly cleaned all of them.

lib/flow/Advection.cpp
ayenpure: The statement breaks the while loop which iterates over the maximum steps for a single particle.
If the particle goes out of domain, we no longer need to continue advecting the while loop.
In a serial setting this will enable proceeding to the next particle.

shaomeng: I see, yeah, I understand now. Thanks for the explanation!

PR #2348, opened 2020-08-26
Enable CentOS build test on submitted PRs

sgpearse:
This adds the compilation of Vapor on CentOS7 to the tests run on submitted PR's.  With this change, the following checks will be run on CircleCI:

- Build on Ubuntu18
- Check for warnings on Ubuntu18
- Build on CentOS7

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,15 +174,6 @@ jobs:
       - store_artifacts:
           path: /tmp/output.txt
 
-      - run:
-          name: check for warnings
-          command: |
-            if grep -q warning /tmp/output.txt; then
-               exit 1
-            else
-               exit 0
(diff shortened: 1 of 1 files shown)


clyne:
@sgpearse the circleci build on centos is not passing.

sgpearse:
@clyne - This is because this branch was derived from our current master, which is also failing its builds on CentOS.

However my configuration did not make this clear because I was directing make's output to a file instead of CircleCI's web interface.  The creation of this file is used to scan for warnings, as we do on Ubuntu18.

Since we don't currently have plans to check for warnings on CentOS, I've removed this step in the CentOS build process.  The output for our make command will now display on Circle's build page, so build failures should be easier to see.

PR #2352, opened 2020-08-27
Partial support for 3D WRF variables with dimension length of one

clyne:
This change allows the calculation of the WRF vertical coordinate when the
vertical dimension of all 3D varibles is of length one. This includes the 
staggered variables (e.g. W, PH, and PHB).


sgpearse (approved):
This passes my tests with the following file:

/glade/p/cisl/vast/vapor/data/Source/WRF/HRRR

PR #2353, opened 2020-08-27
Adding a test_app for FTLE calculation

ayenpure:
The PR contains all the files needed to calculate the FTLE field.
The README explains how to use this utility.

### DIFF
--- a/test_apps/CMakeLists.txt
+++ b/test_apps/CMakeLists.txt
@@ -1,5 +1,6 @@
 if (BUILD_TEST_APPS)
 	add_subdirectory (datamgr)
+	add_subdirectory (ftle)
 	add_subdirectory (grid_iter)
 	add_subdirectory (VDC)
 	add_subdirectory (params2)
--- a/test_apps/ftle/CMakeLists.txt
+++ b/test_apps/ftle/CMakeLists.txt
@@ -1,0 +1,3 @@
+add_executable (ftlecalc ftlecalc.cpp)
+
+target_link_libraries (ftlecalc common vdc wasp flow)
--- a/test_apps/ftle/FTLEHelper.h
+++ b/test_apps/ftle/FTLEHelper.h
@@ -1,0 +1,316 @@
+#ifndef ftle_helper_h
+#define ftle_helper_h
+#include <limits>
+#include <vector>
+
+#include "GridMetaData.h"
+
+namespace detail
+{
+
+class Vec3
+{
+public:
+  Vec3() = default;
+
+  Vec3(double x, double y, double z)
+  {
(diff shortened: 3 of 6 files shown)


clyne:
@ayenpure are you able to run the code with the changes you've made? I'm  getting the following:

CISL-UMATILLA[1062] ./build/bin/ftlecalc -datapath ~/Data/CM1/LeighOrf/directory -output /tmp/ftle.out -dimension 50x50x50 -duration 1.0 -length 0.1 -fieldname uinterp:vinterp:winterp
Advection for FTLE w/ : 
Data : /Users/clyne/Data/CM1/LeighOrf/directory
Field : uinterp | vinterp | winterp
Length : 0.1
Duration : 1
Dimensions : 50 | 50 | 50
nc_open(24Maycontrol.04001.000000.nc,) : No such file or directory
NetCDFSimple::Initialize(24Maycontrol.04001.000000.nc)
Failed to initialize netCDF data collection for reading
Failed to initialize data importer
Failed to intialize CF DataMGR


It looks like it's tripping over building the file list. I'd suggest that rather specifying a directory, and assuming that everything in the directory is a data file, allow the specification of files on the command line (without -datapath option)

ayenpure:
But consider the example of the data we got from the FastEddy team where they have multiple hundred time steps.
How can we handle that case?
In case of FTLE we’re doing time-varying advection.

-Abhishek

clyne:
If you are asking how to build the file list vector, try this:

    vector <string> files;
    for (int i=1; i<argc; i++) {
        files.push_back(argv[i]);
    }

If you are concerned about the number of files and and shell limits on command line length, I wouldn't think that would be an issue for this command; it only makes sense to have the FTLE calculation use a few time steps, correct? I'm fine with having -datapath as an alternate way of specifying the inputs, as long as it is working , and it doesn't currently seem to be :-)

ayenpure:
Haha, definitely not the first!
I should quit s/w development not if I asked you that question.

Yes, my concern was the second.
I just figured the problem.
Since I replaced Boost with VAPOR’s utils I did not realize one thing.

FileUtils::ListFiles only lists file names and not complete paths.
I’ll fix the boo boo and and update the PR, and it should work for the command that you tried to run it with.

Sorry for that :)

-Abhishek

clyne:
I thought you young guys might not know how to handle char** pointers in C++ :-)

I don't see the commit.... Was it pushed to the remote?

ayenpure:
I just pushed it.
Can you verify it works for you?
It worked for my test data.

-Abhishek

clyne:
Works now. Thanks.

PR #2358, opened 2020-09-04
Allow the user to specify when TMS images are downsampled

sgpearse:
This PR addresses #2360

We currently downsample TMS images when their width or height is > 1024 pixels.  This is currently hard-coded.

Sometimes we ~~want to downsample when the image size is larger~~ don't want to downsample the image at all.  This PR allows the user to specify the threshold at which we downsample TMS images in the Image renderer.


clyne (commented):
Please open an Enhancement issue and associate this PR with that issue. 

Allowing the user to change the image resolution by typing in a magic number is a pretty clunky interface. Most numbers entered will not result in any change as the resolutions available are all powers of two. The TMS images are simply an image pyramid. How about having the code determine how many LODs are present and simply providing the user with a slider that changes the currently selected LOD?

sgpearse:
>Allowing the user to change the image resolution by typing in a magic number is a pretty clunky interface.

I agree, but this is a result of how the image renderer is written.  See [line 514](https://github.com/NCAR/VAPOR/blob/18bc942522543e45fcde26c0300be71899504015/lib/render/ImageRenderer.cpp#L510) of ImageRenderer.cpp on master.  

I toyed with a version of this code that simply turns downsampling on or off, but in some cases we may still want to downsample, but a higher threshold, for performance reasons.  This PR still provides that option, while eliminating the hard coded magic number.

>How about having the code determine how many LODs are present and simply providing the user with a slider that changes the currently selected LOD?

I thought about this too, but it gets tricky.  

The LOD of the TMS image is determined within the GeoImageTMS class by the function [GeoImageTMS::_getBestLOD()](https://github.com/NCAR/VAPOR/blob/18bc942522543e45fcde26c0300be71899504015/lib/render/GeoImageTMS.cpp#L357).  

This get called within the pure virtual [GeoImage::GetImage()](https://github.com/NCAR/VAPOR/blob/18bc942522543e45fcde26c0300be71899504015/include/vapor/GeoImage.h#L93) function, which calculates the LOD with the clunky variable (maxWidthReq, maxHeightReq) which on master is hard coded to 1024.  

The first challenge is that the GeoImage classes have no access to the params database, so they can't look up a user defined value.

The second challenge is that if we were to pass a new LOD variable into the pure virtual GeoImage::GetImage() function, we would need to rewrite the function in the GeoImage base class and all subclasses.  There are other ways around this, but everything I can think of seems kludgy.

I hope that all makes sense.  Maybe we can follow up in today's meeting.

apps/vaporgui/ImageSubtabs.h
```diff
@@ -67,6 +79,24 @@ class ImageAppearanceSubtab : public QWidget, public Ui_ImageAppearanceGUI {
 		            VAPoR::RenderParams*  rParams) 
   {
     _rParams = (ImageParams*) rParams;
+    
+    _downsampleInput->Update( _rParams, paramsMgr, dataMgr );
+
+    // Disable the downsample threshold if we're not working with a TMS image
+    //
+    std::string path = _rParams->GetImagePath();
+    if (path.rfind(".tms", path.size()-4) != string::npos) {
```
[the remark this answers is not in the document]
sgpearse: The downsampling of the image only happens with TMS images, so we disable the option if we're operating on non-TMS files.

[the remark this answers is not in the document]
clyne: I understand that you need to know if it is a TMS file or not, but why do you need to actually open the file? Why not simply see if the _geoImage is of type GeoImageGeoTiff, or add a virtual method to the GeoImage base class to see if  the derived class provides LOD support

clyne:
> > Allowing the user to change the image resolution by typing in a magic number is a pretty clunky interface.
> 
> I agree, but this is a result of how the image renderer is written. See [line 514](https://github.com/NCAR/VAPOR/blob/18bc942522543e45fcde26c0300be71899504015/lib/render/ImageRenderer.cpp#L510) of ImageRenderer.cpp on master.
> 
> I toyed with a version of this code that simply turns downsampling on or off, but in some cases we may still want to downsample, but a higher threshold, for performance reasons. This PR still provides that option, while eliminating the hard coded magic number.
> 
> > How about having the code determine how many LODs are present and simply providing the user with a slider that changes the currently selected LOD?
> 
> I thought about this too, but it gets tricky.
> 
> The LOD of the TMS image is determined within the GeoImageTMS class by the function [GeoImageTMS::_getBestLOD()](https://github.com/NCAR/VAPOR/blob/18bc942522543e45fcde26c0300be71899504015/lib/render/GeoImageTMS.cpp#L357).
> 
> This get called within the pure virtual [GeoImage::GetImage()](https://github.com/NCAR/VAPOR/blob/18bc942522543e45fcde26c0300be71899504015/include/vapor/GeoImage.h#L93) function, which calculates the LOD with the clunky variable (maxWidthReq, maxHeightReq) which on master is hard coded to 1024.
> 
> The first challenge is that the GeoImage classes have no access to the params database, so they can't look up a user defined value.
> 
> The second challenge is that if we were to pass a new LOD variable into the pure virtual GeoImage::GetImage() function, we would need to rewrite the function in the GeoImage base class and all subclasses. There are other ways around this, but everything I can think of seems kludgy.
> 
> I hope that all makes sense. Maybe we can follow up in today's meeting.

I agree it may take some restructuring of the code to support easy-to-use LOD specification. So be it.

sgpearse:
@clyne - This is ready for re-review.  A couple notes:

1) The LOD range for NaturalEarth is {0:3}.  I set the current default to be 1, but I think this still looks grainy for Katrina and Duku.  We could calculate a better default based on the Box extents, or we could also bump the default LOD to 2.

2) We discussed using a slider for the LOD controller.  However I think that sliders should be used after a widget contains a certain threshold of options.  For example, we wouldn't want a slider if we are only presenting two or three numeric options to the user.  This is subjective but I think that 10 is a good number.  Since our packaged TMS files only have 4 options, I chose a combo box.

3) The function _tilePath has been changed to be a static inline function (TilePath) in GeoImageTMS.h

4) Two additional static inline functions (IsTMSFile and GetAvailableTMSLODs) have been added to GeoImageTMS.h.  This was done so that GeoImageTMS, and ImageSubtabs use the same logic to determine the information provided by these helper functions.

include/vapor/GeoImageTMS.h
```diff
@@ -9,13 +9,78 @@
...
+#include <fstream>
+#include <sys/stat.h>
 #include <vapor/MyBase.h>
 #include <vapor/UDUnitsClass.h>
 #include "GeoTileMercator.h"
 #include "GeoImage.h"
 
 namespace VAPoR {
 
+static inline bool IsTMSFile(
```
clyne, line 22: This function is trying to do too much. All that it needs to do is determine if the file is a TMS image data base. I.e. does it have a ".tms" extension. Also, our error policy is that functions that return booleans do not report errors as this function does.

include/vapor/GeoImageTMS.h
```diff
@@ -9,13 +9,78 @@
...
+        return false;
+    }
+
+    return true;
+}
+
+// Get the file path of a single tile from the TMS database
+// with the give lod and tile coordinates
+//
+static inline std::string TilePath(
```
clyne, line 45: We don't have a coding convention about this, but typically functions that are more than a couple of lines of code are defined in .cpp files, not .h files. Is there are reason for defining this function in the header file?

include/vapor/GeoImageTMS.h
```diff
@@ -9,13 +9,78 @@
...
+    path = base + ".tiff";
+
+    if (stat(path.c_str(), &statbuf) == 0) return (path);
+
+    // Tile does not exist
+    //
+    return("");
+}
+
+static inline int GetAvailableTMSLODs( std::string dir ) {
```
clyne, line 75: Rather than having this function operate on a directory how about having it operate on the .tms file? Otherwise the caller has to unnecessarily know the details about how TMS data bases are structured.

clyne (commented):
Scott, this looks pretty good, but one thing we've lost is the ability to automatically determine a reasonable LOD. The auto-selection algorithm did a reasonable job in most instances. With this implementation that default LOD (hardcoded to 1) is often not a good choice depending on the scope of the domain. Load Katrina, for example, and the image is way too coarse. A simple fix would be to use the old auto-selection method if LOD is -1, and then make appropriate changes to the GUI (possibly add a "manual select" button, or present a "auto" option in the check box).

include/vapor/GeoImageTMS.h
sgpearse: Fixed.

include/vapor/GeoImageTMS.h
sgpearse: > Is there are reason for defining this function in the header file?

Yes, TilePath() is a helper function used in the render and vaporgui libraries.  GeoImageTMS.cpp and ImageSubtabs.cpp are sharing the logic behind this function, by including GeoImageTMS.h.

GeoImageTMS.cpp is making use of TilePath() directly.

ImageSubtabs.cpp calls TilePath() indirectly by calling GetAvailableTMSLODs(), which is also found in GeoImageTMS.h.

include/vapor/GeoImageTMS.h
sgpearse: The logic here is legacy code, which was just moved from the .cpp file into a helper function in the .h file.

I agree that this is awkward.  I'm always trying to minimize the changes that come with my PRs, because more changes require more testing and more attention from reviewers.

I made a judgment call that refactoring this logic was outside of the scope of this PR.  But we can fix it in this PR too.  Let me know if this is something you'd like me to do.

sgpearse:
@clyne I've responded to your in-line comments.  The first was a quick-fix, the second requires some followup, and I can fix the third if you give a thumbs up.

I'll work on a better default value next.

include/vapor/GeoImageTMS.h
clyne: I'm a little confused. I don't see how these functions are any different from any other public static functions, and why they cannot simply be *declared* in the .h file and *defined* in the .cpp file

include/vapor/GeoImageTMS.h
clyne: Please fix

include/vapor/GeoImageTMS.h
sgpearse: Sorry - I misunderstood the question.

When we met last week I thought we planned on making these helper methods static, and they would be declared outside of the GeoImageTMS class.  They were inlined so they could be externally linked.

But if they're static public member functions, they don't need to be inlined, so I've made that change.

include/vapor/GeoImageTMS.h
sgpearse: Fixed.

sgpearse:
@clyne - I've addressed your two remaining comments, and added a default setting that falls back on our old logic for determining the LOD for the TMS image.

include/vapor/GeoImageTMS.h
```diff
@@ -27,8 +30,14 @@ class RENDER_API GeoImageTMS : public GeoImage {
  GeoImageTMS();
  virtual ~GeoImageTMS();
 
+ static bool IsTMSFile( std::string path );
+ static std::string TilePath( std::string dir, size_t tileX, size_t tileY, int lod );
```
clyne, line 34: dir -> 'file' or 'path'. 'dir' would imply directory,  wouldn't it, which the argument is not

clyne (commented):
Something is not right in your logic. When I load katrina and select the default I get a different result than what is possible with any of the manually selected LODs. Zoom in on Florida, for example, to see the difference. I'd guess the true maximum LOD is not available from the manual selector. Note: it looks like GetAvailableTMSLODs returns the maximum LOD (starting from 0), not the number of LODs available.

include/vapor/GeoImageTMS.h
clyne: Whether they are static member functions or simply free functions they can still be made visible to other modules by simply declaring them in the .h files and defining them in a .cpp file that is linked as a library.

include/vapor/GeoImageTMS.h
sgpearse: The 'dir' argument in TilePath can either be a file or a directory.  'file' describes both, so I'll change it to that.

sgpearse:
>I'd guess the true maximum LOD is not available from the manual selector. Note: it looks like GetAvailableTMSLODs returns the maximum LOD (starting from 0), not the number of LODs available.

You were right - thanks for catching this.

include/vapor/GeoImageTMS.h
sgpearse: True, however I'm pretty sure that static free functions must be inlined.  But I understand this is not what we wanted.

clyne:
It dumps core when I load CAM/AMIP_0.25degree/cam5_1_amip_run2.cam2.h1.1979-01-01-00000.nc  (or any other global data set) and select the maximum LOD

clyne:
> It dumps core when I load CAM/AMIP_0.25degree/cam5_1_amip_run2.cam2.h1.1979-01-01-00000.nc (or any other global data set) and select the maximum LOD

The core dump is due to a memory allocation error, which occurs in ImageRender.cpp:

```
    // (Re)allocate space for verts
    //
    _nverts = width * height * 3;
    _sb_verts.Alloc(_nverts * 3 * sizeof(GLfloat));
    _sb_normals.Alloc(_nverts * 3 * sizeof(GLfloat));
```

By default 'width' and 'height' are set to the width and height of the texture, which was reasonable before this change, but will not work now. A simple fix is the following:

diff --git a/lib/render/ImageRenderer.cpp b/lib/render/ImageRenderer.cpp
index 0d69a7c4..62a87cad 100644
--- a/lib/render/ImageRenderer.cpp
+++ b/lib/render/ImageRenderer.cpp
@@ -209,8 +209,8 @@ int ImageRenderer::GetMesh( DataMgr *dataMgr,
                // will be used to set the width and height of the mesh.
                //
         _getTexture(dataMgr);   // Ugh, this function is more than a get method...
-               _vertsWidth = _texWidth;
-               _vertsHeight = _texHeight;
+               _vertsWidth = min(1024, _texWidth);
+               _vertsHeight = min(1024, _texHeight);
                rc = _getMeshDisplaced(

Obviously the literal 1024 should be defined as a const such as MaxResamplingResolution.

include/vapor/ImageParams.h
```diff
@@ -95,11 +95,25 @@ class PARAMS_API ImageParams : public RenderParams
     SetValueLong( _orientationTag, "set orientation value", val );
   }
 
+  //
```
clyne, line 98: Needs Doxygen documentation

include/vapor/ImageParams.h
```diff
@@ -95,11 +95,25 @@ class PARAMS_API ImageParams : public RenderParams
...
+  // Get and set the threshold for TMS image downsampling
+  // 
+  int GetTMSLOD() const
+  {
+    int value = ( (int)GetValueLong( _TMSLODTag, -1 ) );
+    return value;
+  }
+  void SetTMSLOD( int val )
```
clyne, line 106: Needs Doxygen documentation

clyne (commented):
The LOD selector needs a context help message. Other than that (and inline comments), looks good.

apps/vaporgui/ImageSubtabs.h
```diff
@@ -40,6 +42,49 @@ class ImageVariablesSubtab : public QWidget {
...
+        if ( GeoImageTMS::IsTMSFile( imageFile ) ) {
+            _vComboBox->setEnabled( true );
+        }
+        else {
+            _vComboBox->setEnabled( false );             // Disable if not using a TMS image
+            return;
+        }
+
+        std::vector< std::string > options{"default"};
+        int lods = GeoImageTMS::GetNumTMSLODs( imageFile );
```
StasJ, line 72: GetNumTMSLODs does multiple file IO operations recursively. We may not want to do this every update.

clyne: Agreed this is a problem. File system metadata performance on large parallel filesystems (e.g. glade) is awful. A suggested fix would be to only call GetNumTMSLODs() when imageFile changes.

sgpearse: I tried caching the file name in this function (updateGUI()) but this function is const, and we can't update a private member here.

Instead, I've written a new param in ImageParams that holds the number of LODs available.  This only gets updated when the image file changes.

StasJ: Great! You are setting the number of LODs when the image TMS file is set. This needs to be wrapped in a Params group.

StasJ (commented):
Could you please move PTMSLODInput out of the Subtab file since those are all deleted in the refactor PR?

sgpearse:
> Could you please move PTMSLODInput out of the Subtab file since those are all deleted in the refactor PR?

No problem.  Moved to PTMSLODSelector.h.  Hell of an acronym...

apps/vaporgui/ImageSubtabs.h
sgpearse: Good catch.  done.

clyne: > I tried caching the file name in this function (updateGUI()) but this function is const, and we can't update a private member here.
> 
> Instead, I've written a new param in ImageParams that holds the number of LODs available. This only gets updated when the image file changes.

Rather than introduce a new parameter, and a getter that will return the wrong answer if it is called before SetImagePath, why not make use of the C++ 'mutable' keyword inside of updateGUI to allow an exception to the const declaration? This is what the mutable keyword is for, and the model that we use elsewhere for cached (expensive to compute) values returned by const getter functions. See, for example, the use of _minuCache inside of the Grid class.

include/vapor/ImageParams.h
```diff
@@ -95,14 +102,44 @@ class PARAMS_API ImageParams : public RenderParams
     SetValueLong( _orientationTag, "set orientation value", val );
   }
 
+  //! Get the TMS level of detail
+  //! \retval int Currently selected TMS level of detail.  Value of 0 means we are computing a default value.
```
clyne, line 106: The default value returned is -1. What does this mean?

clyne (commented):
Getting super close!

apps/vaporgui/ImageSubtabs.h
StasJ: @clyne If you have two image renderers, it will have to keep track of which renderer is currently active and manually keep its cache in sync.

StasJ (approved):

apps/vaporgui/ImageSubtabs.h
clyne: Good point.

clyne: @StasJ , on second thought, why would you have to keep track of the active renderer? If you cache the file name and # LODs then when the active renderer changes the information will still be valid if the current and previous renderers both use same image file. If the two render instances use different image files that will trigger a cache update, but switching renderers doesn't happen at a high enough frequency to be a performance problem.

include/vapor/ImageParams.h
sgpearse: Sorry - my comment was obsolete.  -1 means we're computing a default value.  Comment has been updated.

clyne (approved):
Nice work!

PR #2359, opened 2020-09-04
Issue 1636: Renderers are deleted and recreated for each undo

### LINKED ISSUE #1636: Renderers are deleted and recreated for each undo
This can result in unusable performance when performing undo with large 3D data

### LINKED ISSUE #2317: Closed visualizer windows not removed from params
1. Open a dataset in vapor
1. Create a new visualizer window
1. Close the visualizer window
1. Save the session
1. Open the session

The closed visualizer will reappear.

clyne:
Fixes #1636 - Renderers are deleted and recreated for each undo
Fixes #2317 - Closed visualizer windows not removed from params 

This PR is primarily a performance optimization that prevents data and visualizers from being unnecessarily reloaded every time there is an undo or redo event.


apps/vaporgui/RenderHolder.cpp
```diff
@@ -351,8 +351,6 @@ void RenderHolder::_newRendererDialogAccepted()
 	rendererName = uniqueName(rendererName);
 	qname = QString(rendererName.c_str());
 
-	paramsMgr->BeginSaveStateGroup("Create new renderer");
```
sgpearse, line 354: I haven't tested this, and I'm not sure if it's going to have a consequence.  There is still a call to paramsMgr->EndSaveStateGroup() at line 358 on the current branch, which does not have a BeginSaveStateGroup() associated with it..

include/vapor/ControlExecutive.h
```diff
@@ -123,13 +123,34 @@ class RENDER_API ControlExec : public MyBase {
  //
  int NewVisualizer(string name);
 
+ //! Return tag pushed to the undo stack when NewVisualizer() is called
+ //
+ string GetNewVisualizerUndoTag() const {
```
sgpearse, line 128: Just a question, not trying to nitpick:  Why does the ControlExecutive handle tags as hard coded strings, as opposed to the static const strings as is done with the Params classes?

StasJ (changes requested):
Regression:

1. Open a dataset
1. Create a renderer
1. Press undo (the renderer is not removed)
1. Press undo again (the renderer is still not removed)
1. Press undo again

Vapor will crash.

apps/vaporgui/RenderHolder.cpp
clyne: Good catch!

include/vapor/ControlExecutive.h
clyne: They are different "tags". The params classes required strings that were visible at compile time (at least it did when they were first implemented). There is no such requirement in this instance, so a traditional,  "safer" getter was used.

clyne:
> Regression:
> 
> 1. Open a dataset
> 2. Create a renderer
> 3. Press undo (the renderer is not removed)
> 4. Press undo again (the renderer is still not removed)
> 5. Press undo again
> 
> Vapor will crash.

Whoops. Missed some commits. Should be good now.

StasJ (commented):
This is looking great. Too many times I have accidentally hit undo with a large dataset open and instantly regretted it.

The code review is approved.

I found a couple more bugs. These may addressed later as separate bug reports.

Dimensionality change breaks undo.
1. Create a wireframe
1. Change the dimension
1. Press undo many times
Undo will no longer do anything. Switching back to the other dimension and pressing undo a few more times will cause Vapor to crash. This may not be important with a new dimensionality selector on the way but this may reveal a more general bug.

Undoing opening sessions/datasets doesn't quite work. I don't know if these operations are supposed to be supported by undo. 
1. Open a session
2. Press undo a few times (5-10?)
Vapor will crash.

Undoing a projection string change results in some odd behavior:
1. Open JSmall
1. Create/enable a 2D data renderer
1. Change the projection string to have a different lat_0 parameter
1. Undo 
The extents of the original renderer will have changed.

I think there are a bunch of these throughout vapor and it may just require adding params change groups but I happened to notice these as I was testing so I included them here:
- Deleting a renderer takes 2 undos.
- Creating a new visualizer window takes 4 undos.

clyne:
> This is looking great. Too many times I have accidentally hit undo with a large dataset open and instantly regretted it.
> 
> The code review is approved.
> 
> I found a couple more bugs. These may addressed later as separate bug reports.
> 
> Dimensionality change breaks undo.
> 
> 1. Create a wireframe
> 2. Change the dimension
> 3. Press undo many times
>    Undo will no longer do anything. Switching back to the other dimension and pressing undo a few more times will cause Vapor to crash. This may not be important with a new dimensionality selector on the way but this may reveal a more general bug.

Let's save this for later

> 
> Undoing opening sessions/datasets doesn't quite work. I don't know if these operations are supposed to be supported by undo.
> 
> 1. Open a session
> 2. Press undo a few times (5-10?)
>    Vapor will crash.
> 

Fixed by disabling undo/redo for session loads. Supporting these would take more effort than it is probably worth.

> Undoing a projection string change results in some odd behavior:
> 
> 1. Open JSmall
> 2. Create/enable a 2D data renderer
> 3. Change the projection string to have a different lat_0 parameter
> 4. Undo
>    The extents of the original renderer will have changed.

Ditto. Disabled undo/redo for these as well.

> 
> I think there are a bunch of these throughout vapor and it may just require adding params change groups but I happened to notice these as I was testing so I included them here:
> 
> * Deleting a renderer takes 2 undos.
> * Creating a new visualizer window takes 4 undos.

I've wrapped a number of the multi-state param change events, but I'm sure that we'll find more with time.

StasJ (approved):

sgpearse (approved):

PR #2365, opened 2020-09-11
Fix  #1927 - regular grids defined in NetCDF are treated as stretched grids

### LINKED ISSUE #1927: regular grids defined in NetCDF are treated as stretched grids
NetCDF has no means to explicitly describe a Regular grid (one where all cells are the same size). Hence, Regular grids are defined in NetCDF using 1D coordinate variables, with uniform spacing between each value. Currently VAPOR does not examine the values inside 1D coordinate variables and thus assumes that a 1D coordinate variable describes a "stretched" dimension. Thus when accessing variables with 1D coordinate variables from NetCDF these variables are represented using the StretchedGrid class (as opposed to the RegularGrid class). A number of algorithms in VAPOR (e.g. DVR) operate more efficiently on the RegularGrid class. 

Reported by @leighorf

clyne:
Fix #1927

### DIFF
--- a/lib/common/utils.cpp
+++ b/lib/common/utils.cpp
@@ -1,6 +1,7 @@
 #include "vapor/VAssert.h"
 #include <iostream>
 #include <algorithm>
+#include <cmath>
 #include <vapor/utils.h>
 
 #define MAXCOORDS 4
@@ -250,4 +251,25 @@ bool Wasp::BinarySearchRange(
 	}
 	i = i0;
 	return(true);
+}
+
+
+bool Wasp::NearlyEqual(float a, float b, float epsilon) {
+	float absA = fabs(a);
+	float absB = fabs(b);
+	float diff = fabs(a-b);
+	
+	// shortcut, handles infinities
+	//
+	if (a == b) return (true);
+
+	if (a == 0.0 || b == 0.0 || (absA + absB < std::numeric_limits<float>::min())) {
+		// a or b is zero or both are extremely close to it
+		// relative error is less meaningful here
+		//
+		return (diff < (epsilon * std::numeric_limits<float>::min()));
+	}
+
+	// use relative error
+	return (diff / min((absA + absB), std::numeric_limits<float>::max()) < epsilon);
 }
--- a/lib/common/utils.cpp
+++ b/lib/common/utils.cpp
@@ -255,9 +255,9 @@ bool Wasp::BinarySearchRange(
 
 
 bool Wasp::NearlyEqual(float a, float b, float epsilon) {
-	float absA = fabs(a);
-	float absB = fabs(b);
-	float diff = fabs(a-b);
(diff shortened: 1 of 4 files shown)


lib/common/utils.cpp
```diff
@@ -251,3 +252,24 @@ bool Wasp::BinarySearchRange(
 	i = i0;
 	return(true);
 }
+
+
+bool Wasp::NearlyEqual(float a, float b, float epsilon) {
+	float absA = fabs(a);
+	float absB = fabs(b);
+	float diff = fabs(a-b);
```
StasJ, line 260: Ironically, `fabs` is the double abs function. The float abs function is `fabsf`.

lib/vdc/DCCF.cpp
```diff
@@ -418,6 +418,40 @@ int DCCF:: _get_latlon_coordvars(
...
+		return(false);
+	}
+	ncdfc->Close(fd);
+	EnableErrMsg(enabled);
+
+	float epsilon = 0.0001;
+	for (int i = 0; i<buf.size()-2; i++) {
+		if (! Wasp::NearlyEqual((buf[i+1] - buf[i]), (buf[i+2] - buf[i+1]), epsilon)) {       
+			return(false);
+		}
```
StasJ, line 448: This can be simplified and made more accurate by comparing the first delta against all others rather than all neighboring deltas.

lib/common/utils.cpp
clyne: As of C++11 fabs is overloaded to work on 32bit floats (and other types)

lib/vdc/DCCF.cpp
clyne: True enough. Done.

lib/common/utils.cpp
StasJ: Oh, cool.

StasJ (approved):

lib/common/utils.cpp
shaomeng: From a good programming habit standpoint, if you want to use C++11 version of `fabs`, you need to specify `std::fabs()`. That's because, on Linux, the code above will in fact invoke the C language version of fabs.

From a practical standpoint, you can argue the wasted cycles dont matter.

P.s. you can run the following code to verify which fabs is invoked.

```C++
#include <iostream>
#include <type_traits>
#include <cmath>

int main()
{
    float f = 3.5f;
        
    auto v1 = fabs( f );
    auto v2 = std::fabs( f );

    std::cout << std::boolalpha;
    std::cout << std::is_same< decltype(v1), double >::value << '\n';
    std::cout << std::is_same< decltype(v2), double >::value << '\n';
}
```

clyne: Good point. Done.

shaomeng (approved):

PR #2373, opened 2020-09-26
Closes Issue #1377 - Undo open dataset doesn't fully close it

### LINKED ISSUE #1233: Incorrect behavior after closing/opening datasets
The following procedure has two problems.  1) All renderers get deleted, and 2) the wrong projection string is used.

Procedure:
1) Spin up Vapor, and import the CAM file /little_glade/data/Source/CAM/Gaussian/b30.030a.cam2.h0.TS.1870-01_cat_1999-12.nc
2) Close the imported CAM file
3) Import the WRF file /little_glade/data/Source/WRF/DUKU/wrfout_d02_2007-04-01_00\:00\:00.nc
4) Create a 2D renderer
5) Under the Navigation tab, select the Viewpoint tab.  Note that the projection string is referencing that of our closed dataset.  It should be invalid at this point.  Now select the inactive projection string, that references our WRF dataset.
6) Our TwoDData renderer is now deleted

### LINKED ISSUE #1377: Undo open dataset doesn't fully close it
1. Open a dataset
2. Press undo until vapor stops undoing
3. Notice that the dataset is gone from the `Close Dataset` menu
4. Open the same dataset

Vapor will prompt you whether you want to import the data into an existing dataset (the one you undid) or a new one. Whichever you chose, vapor will be in a glitched state and all renderers will have empty variable selectors.

Also, if you undo opening a dataset, you will still be able to create renderers with no dataset loaded.

clyne:
Closed #1377 - Undo open dataset doesn't fully close it
Fixed #1233 - Incorrect behavior after closing/opening datasets 

Supporting undo/redo for open/close of data sets is not possible with the current architecture. This PR disables undo/redo for these operations. It also ensures that data sets are properly removed from the database when they are closed. As a byproduct it also fixes #1233

### DIFF
--- a/lib/params/ParamsMgr.cpp
+++ b/lib/params/ParamsMgr.cpp
@@ -182,9 +182,10 @@ void ParamsMgr::LoadState(const XmlNode *node) {
 
 	// If data loaded set up data dependent parameters from new state.
 	//
-	map <string, DataMgr *>::const_iterator itr;
-	for (itr = _dataMgrMap.begin(); itr != _dataMgrMap.end(); ++itr) {
-		addDataMgrMerge(itr->first);
+	vector <string> dataSetNames = GetDataMgrNames();
+	for (auto dataSetName : dataSetNames) {
+		map <string, DataMgr *>::const_iterator itr = _dataMgrMap.find(dataSetName);
+		if (itr != _dataMgrMap.end()) addDataMgrMerge(itr->first);
 	}
 	EndSaveStateGroup();
 }
@@ -308,6 +309,17 @@ void ParamsMgr::AddDataMgr(string dataSetName, DataMgr *dataMgr) {
 		addDataMgrNew();
 	}
 
+	ParamsSeparator windowsSep(_rootSeparator, _windowsTag);
+	XmlNode *winSepNode = windowsSep.GetNode();
+	for (int i=0; i<winSepNode->GetNumChildren(); i++) {
+		string winName = winSepNode->GetChild(i)->GetTag();
+
+		ParamsSeparator windowSep(&windowsSep, winName);
+
+		ParamsSeparator renderSep(&windowSep, _renderersTag);
+
+		ParamsSeparator dataSep(&renderSep, dataSetName);
+	}
 }
 
 void ParamsMgr::RemoveVisualizer(string winName) {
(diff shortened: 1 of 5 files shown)


lib/render/Visualizer.cpp
```diff
@@ -584,7 +584,7 @@ int Visualizer:: _captureImage(std::string path)
         goto captureImageEnd;
     
     if (geoTiffOutput) {
-        string projString = _dataStatus->GetDataMgr(_dataStatus->GetDataMgrNames()[0])->GetMapProjection();
+        string projString = _dataStatus->GetDataMgr(_paramsMgr->GetDataMgrNames()[0])->GetMapProjection();
```
sgpearse, line 587: At line 171, we guard against the case where there is no DataMgr, and _paramsMgr->GetDataMgrNames() returns an array of size 0.  Do we need to guard against that case here as well?

sgpearse (commented):
I get a crash with the following procedure:

1) Import any WRF file
2) Rotate the scene
3) Undo
4) Create a Volume Renderer

It appears that the undo step is clearing the list of DataMgrs, held by the ParamsMgr.

clyne:
> I get a crash with the following procedure:
> 
> 1. Import any WRF file
> 2. Rotate the scene
> 3. Undo
> 4. Create a Volume Renderer
> 
> It appears that the undo step is clearing the list of DataMgrs, held by the ParamsMgr.

Good catch. Thanks. Fixed.

lib/render/Visualizer.cpp
clyne: Never hurts to add an assert. Done.

lib/params/ParamsMgr.cpp
```diff
@@ -308,6 +309,17 @@ void ParamsMgr::AddDataMgr(string dataSetName, DataMgr *dataMgr) {
 		addDataMgrNew();
 	}
 
+	ParamsSeparator windowsSep(_rootSeparator, _windowsTag);
```
StasJ, line 312: This code is not super clear as to what it accomplishes. A comment may be useful. Especially since it seems to be doing things by creating instances of classes which are then discarded which is not standard. (I am assuming the constructors modify the inputs?)

StasJ (approved):

lib/params/ParamsMgr.cpp
clyne: Agreed. I've added a comment. Ultimately it would be nice to refactor the ParamsBase classes to be more intuitive...

sgpearse (approved):

PR #2376, opened 2020-09-29
Issue2241 minimal

sgpearse:
A demonstration of the bugs found during PDimensionSelector, relating to MainForm's eventFilter logic.


sgpearse:
To reproduce the bug:
1) Load data (I use WRF/DUKU)
2) Create and enable a Wireframe renderer
3) Change the dimensionality to 2D with the PDimensionSelector, located at the bottom of the Variables tab
4) Set the 2D variable to HGT
5) Change the dimensionality back to 3D with the PDimensionSelector, and PWidgets will no longer update correctly

PR #2377, opened 2020-09-29
Issue 2350

shaomeng:
Many changes related to Flow Advection performance. 

In one test with steady advections, this PR reduces execution time from 56s to 15s. 

One caveat is that the progress bar is gone at this point. It'll require some thinking on how to bring it back.


clyne:
@shaomeng , this is good progress but I believe substantial more performance improvement is possible (and needed). There is still way too much redundant (and expensive) work unnecessarily being performed in the inner most loop. When I profile with only 1000 seed points only ~0.2% of the time is spent performing RK4 integration, while 38% is spent calling _getAGrid, a function that only needs to be called when the time step advances and new time steps are needed, but is instead called for every seed and every deltaT. Similarly, RenderParams::GetCurrentTimeStep and FlowParams::GetVelocityMultiplier are in the inner most loop and account for 2.3% and 1.3% of the time, respectively. The inner most loop, which will potentially be executed millions, or even billions of times, must avoid function calls, branching, and redundancy as much as possible. 

Also, another ~17% of the time is spent calling VaporField::InsideVolumeVelocity, which in turn calls Grid::InsideGrid. The latter is an inherently expensive operation because it has to find the cell containing a point; the same operation that Grid::GetValue performs. The duplicate cell search could be avoided by simply ensuring that the code calling GetValue() handles missing values, which are what GetValue returns if the point is outside the grid.

shaomeng:
I agree that there are some duplicate calculations, but I'm also
considering the balance between high-performance and a modularized design.
I.e., as a general-purpose module, the advection code shouldn't expect too
much from others, and should maintain its internal states rather
independent.

Let me dig a bit and shift the balance towards the performance side a bit ;)

clyne:
We need to shift it a LOT toward the performance side :-)

shaomeng:
Hi John,

The PR is ready to be review again. The new changes are mostly regarding a caching mechanism that avoids checking in with the params database in the middle of an advection. 

Regarding `VaporField::InsideVolumeVelocity`, actually, it's not called as often, since there's a boolean passed to `VaporField::GetVelocity()` controlling if a separate inside volume check is performed, and that boolean is `false` every time `VaporField::GetVelocity()` was invoked.

Please let me know if you spot other optimization opportunities. 

Best,
Sam

clyne:
That’s great, Sam. I obviously will not have a chance to review it until I
get back on the 12th. Have you profiled the code to make sure all of the
bottlenecks have been Eliminated?

Sent from my iPhone

On Oct 1, 2020, at 3:53 PM, Samuel Li <notifications@github.com> wrote:

﻿

Hi John,

The PR is ready to be review again. The new changes are mostly regarding a
caching mechanism that avoids checking in with the params database in the
middle of an advection.

Regarding VaporField::InsideVolumeVelocity, actually, it's not called as
often, since there's a boolean passed to VaporField::GetVelocity()
controlling if a separate inside volume check is performed, and that
boolean is false every time VaporField::GetVelocity() was invoked.

Please let me know if you spot other optimization opportunities.

Best,
Sam

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/NCAR/VAPOR/pull/2377#issuecomment-702417733>, or
unsubscribe

shaomeng:
Hi John,

There are no more simple improvements as far as I could tell. I've attached a screenshot of a tree view from VTune, as well as the VTune profiling log files. I believe you can navigate the profiling files if you simply download a copy of VTune.

tree_view

Profile log files:
[r009hs.tar.gz](https://github.com/NCAR/VAPOR/files/5324112/r009hs.tar.gz)



On a separate note, it looks `Wasp::BinarySearchRange()` is taking a lot of time. 
hotspot_view

I believe that for querying inside a `StretchedGrid`, a bounding box check should be sufficient. I reported it as issue #2383 . I suspect that resolving that issue will greatly shift the balance between the current hotspots.

clyne:
@shaomeng the current implementation is improved, but still makes unnecessary (and expensive) function calls to retrieve invariant information deep inside the execution kernel. As noted earlier this is particularly problematic with the call to getAGrid inside of VaporField::GetVelocity(), which can consume nearly 25% of the execution time. This can be seen by using Leigh's CM1 data (24Maycontrol), which are sampled on a regular grid, and therefore have very inexpensive grid access operations. Yes, there are inefficiencies in the Grid access methods, and these should be addressed, but they can be handled outside of this PR and will be beneficial elsewhere in the code. With regard to #2383 in particular, yes a bounding box check can (and should) be added, but this will not help with Curvilinear grids. Hence, VaporField::InsideVolumeVelocity() becomes an expensive call (consuming ~10% of execution time with regular grids) that should be avoided. Per my earlier comment I'd suggest having the advection kernels (e.g. _advectEuler) return a status code if the particle leaves the volume (i.e. if components of the velocity field are missing values) rather than having an explicit call to InsideVolumeVelocity()

shaomeng:
@clyne The newest changes eliminates a `insideVolume` check at the beginning of each advection step. Instead, it checks such situations when advection fails.It's ready to be reviewed now.

clyne:
> @clyne The newest changes eliminates a `insideVolume` check at the beginning of each advection step. Instead, it checks such situations when advection fails.It's ready to be reviewed now.

It's a significant improvement, and I've added a commit to this PR that saves a few more cycles, but _getAGrid still consumes ~33% of the time when processing a regular grid. Specifically, if I load Leigh's 24MayControl data and define a 50x50x50 grid for seeds, _getAGrid consumes 9 of the total 25 seconds. Computing an FTLE field would require much denser seeding (typically one seed point per grid point). I.e. even if parallelized interactive FTLE calculation won't be possible for even modest sized grids. We've got to remove _getAGrid from the inner most loop (or make it much, much faster).

FWIW the version on master takes over 17 minutes to run this configuration, so you've made a huge improvement!

shaomeng:
@clyne 
The vast majority of time inside of that function is spent on mutex handling. As the screenshot shows, in this 110s run, only 3s is spent on that function itself, and 17s spent on mutex handling. So if you think it's a good idea, we can remove mutexes for now and reduce the cost of getAGrid from 20s to 3s. 

Btw, in my test, even with mutexes in, it takes 20s out of 110s, still far lower percentage than your test. I don't know if it's a Linux vs. MacOS thing, or the percentage simply goes down as you run larger problems.

getAGrid

clyne:
I am assuming the code is not thread safe and that is why it needs mutexes. If you remove the mutex it won't parallelize. If mutexes are required parallel performance will be negatively impacted. Sam, please remove _getAGrid from the inner most loop. The requirements for this algorithm are high performance, and we're not going to get there with a complicated function that computes redundant information in the inner most loop of the kernel.

shaomeng:
@clyne I've reduced the times mutexes are invoked, now in my test, mutex handling doesn't make to my top-hotspot list. Would you try it on your machine as well?

clyne:
@shaomeng, I've already pointed out that mutexes should be avoided in performance sensitive code as described here:

https://link.springer.com/chapter/10.1007/978-1-4842-4398-5_5

and many other places. If they're inside a tight kernel they will kill parallel performance. Please remove the unnecessary call to _getAGrid, and any other code that will greatly impact parallel performance (e.g. mutexes) from the inner kernel.

shaomeng:
@clyne I completely agree with you that mutexes and other expensive operations should be avoided in a performance critical kernel. However, I have the following observations.

1. The current getAGrid() function constitutes less than 3s in a 108s run. Even assuming 100% of its calculations are useless (which of course isn't the case), your gain is still less than 3%. At the same time, there are many function calls with great improvement potential. For example, `StructuredGrid::ClampCoord` takes more than 10s, and `RegularGrid::InsideGrid` takes more than 4s. 

getAGrid

2. The requirement of extremely high performance isn't compatible with the rich functionality requirement of a flow renderer we've known of. For example, a generic flow renderer supports dozens of controls and types of input, including 2D/3D grids, missing values, boundary periodicity, etc. These controls can change at various points of the execution pipeline. To really achieve the high performance that you were looking for, we need to explicit separate the generic and high-performance implementations with much stronger assumptions applied to the high-performance implementation. By doing that, even more optimization opportunities (beyond the scope of slow) could be unlocked.

clyne:
Removing the dynamic cast definitely helped. Now getAGrid takes ~15% of the time. I say we move on with the rest of the code review. We may have to revisit this once the parallelization pragmas are reintroduced (or if we are able to make substantial improvements in the Grid access functions).

include/vapor/Field.h
```diff
@@ -32,30 +33,33 @@ class FLOW_API Field
...
     // It is 3 when the object is newly created, or is used to represent a scalar field
     // 
     int GetNumOfEmptyVelocityNames() const;
 
+    //
+    // Provide an option to cache and lock certain parameters.
+    // Both functions return 0 on success.
+    //
+    virtual auto LockParams()   -> int = 0;
+    virtual auto UnlockParams() -> int = 0;
```
StasJ, line 57: Is there a reason this is written this way? It is inconsistent with the reset of the code base. Also is it confirmed that this works with the windows compiler?

include/vapor/VaporField.h
```diff
@@ -11,6 +11,62 @@
...
+    // 0        uint32_t    currentTS
+    // 4        int32_t     refLevel
+    // 8        int32_t     compLevel
+    // 12       float[3]    extent_min
+    // 24       float[3]    extent_max
+    // 36       uint32_t    variable_dimension
+    // 40       float       DefaultZ
+    // 44       uint8_t     "if this buf should be compared"
+    // Total size: 45 
+    std::array<uint8_t, 45> buf;
```
StasJ, line 35: Why is this an anonymous array of bytes rather than a struct?

lib/render/FlowRenderer.cpp
```diff
@@ -297,43 +298,20 @@ FlowRenderer::_paintGL( bool fast )
...
-                rv = flow::ADVECT_HAPPENED;
-                for( size_t i = _2ndAdvection->GetMaxNumOfPart() - 1; 
-                     i < numOfSteps && rv == flow::ADVECT_HAPPENED; i++ )
-                {
-                    Progress::Update(done++);
-                    if (Progress::Cancelled())
-                        return 0;
-                    rv = _2ndAdvection->AdvectOneStep( &_velocityField, deltaT2 );
-                }
-                Progress::Finish();
```
StasJ, line 334: There is a Progress::StartIndefinite method you could use here if you cannot maintain a list of completed jobs.

include/vapor/Field.h
shaomeng: It's just a new syntax that facilitates some meta programming use cases. Here it simply looks different.

It's part of the standard, and a lot of people advocate for it (I kind of like it too). I'd be surprised if visual studio doesn't support. If that's the case, I'd have no problem to convert to the old style.

include/vapor/VaporField.h
shaomeng: It's just a storage space. In this use case, I only care if the content is the same, but don't care if they're interpreted as floats/ints/doubles etc. As a result, I could use a single `std::strcmp()` function call to compare the entire buffer space, instead of performing a dozen individual comparisons between ints/floats/doubles etc.

shaomeng:
@clyne I implemented a cache of cache which seems to greatly reduced the invocations of `_getAGrid()`, just FYI.

lib/flow/Advection.cpp
```diff
@@ -43,102 +43,146 @@ Advection::CheckReady() const
...
+            float dt = deltaT;
+            if( s.size() > 2 )  // If there are at least 3 particles in the stream and
+            {                   // neither is a separator, we also adjust *dt*
+                const auto& past1 = s[ s.size()-2 ];
+                const auto& past2 = s[ s.size()-3 ];
+                if( (!past1.IsSpecial()) && (!past2.IsSpecial()) )
                 {
-                    loc[i] = _applyPeriodic( loc[i], _periodicBounds[i][0], _periodicBounds[i][1] );
-                    locChanged = true;
+                    float mindt = deltaT / 20.0f,   maxdt = deltaT * 20.0f;
```
clyne, line 78: Please comment on what this does and the use of the literal 20.0. Of concern is whether assumptions are being made about the velocity field range

lib/flow/Advection.cpp
```diff
@@ -43,102 +43,146 @@ Advection::CheckReady() const
...
             }
+            else if( rv == MISSING_VAL ) {
+
+                // This is the annoying part: there are multiple possiblities.
+                // 1) past0 is really located at a missing value location;
+                // 2) past0 is inside the volume, but really close to the boundary,
+                //    causing RK4 method to fail;
+                // 3) past0 is not at a missing location, but out of the volume.
+                // Now we need to detect and deal with each of these possibilities.
+
```
clyne, line 110: Please add a comment indicating why periodic boundaries are being handled here, and not by the Grid class itself.

lib/flow/VaporField.cpp
```diff
@@ -1,37 +1,165 @@
...
+//
+// Class GridKey
+//
+void GridKey::Reset( uint32_t ts, int32_t ref, int32_t comp, std::string var,
+                     const std::vector<double>& min,
+                     const std::vector<double>& max,
+                     float dz )
+{
+    // See the header file for information stored at each offset.
+    std::memcpy( buf.data(),      &ts,        4 );
```
clyne, line 18: Using sizeof() (or a C++ equivalent) would be preferable to hard-coding the data element size

include/vapor/VaporField.h
```diff
@@ -11,6 +11,60 @@
...
+    // Offset   Data_Type   Information
+    //
+    // 0        uint32_t    currentTS
+    // 4        int32_t     refLevel
+    // 8        int32_t     compLevel
+    // 12       double[3]   extent_min
+    // 36       double[3]   extent_max
+    // 60       float       DefaultZ
+    //
+    // Total size: 64
```
clyne, line 33: Using hard-coded data sizes is error prone and hard to maintain. Can't you define a struct for these quantities and make use of sizeof? Alternatively, does 'buf' need to be an std::array? 

Perhaps the bigger question is why pack the quantities into an array at all? Why not just retain and compare the individual quantities? Some additional documentation would be helpful.

lib/render/FlowRenderer.cpp
```diff
@@ -29,8 +30,8 @@ FlowRenderer::FlowRenderer( const ParamsMgr*    pm,
                       FlowRenderer::GetClassType(),
                       instName,
                       dataMgr ),
-            _velocityField     ( 8 ),
-            _colorField        ( 2 ),
+            _velocityField     ( 9 ),
```
clyne, line 33: How were these numbers determined? Please document.

lib/render/FlowRenderer.cpp
```diff
@@ -297,43 +298,20 @@ FlowRenderer::_paintGL( bool fast )
             /* If the advection is single-directional */
             if( params->GetFlowDirection() == 1 )           // backward integration
                 deltaT *= -1.0f;
-            int numOfSteps = params->GetSteadyNumOfSteps();
-            int total = numOfSteps - (_advection.GetMaxNumOfPart() - 1);
-            int done = 0;
+            long numOfSteps = params->GetSteadyNumOfSteps();
+            //long stepsToGo  = numOfSteps - long(_advection.GetMaxNumOfPart() + 1);
```
clyne, line 302: Can this commented out code be removed?

clyne (commented):
Looks good, @shaomeng . Just a few minor in-line comments.

include/vapor/Field.h
shaomeng: C++11. This is something required to support multiple C++11 meta programming features, so it's unlikely to be left out.

lib/render/FlowRenderer.cpp
shaomeng: Cool. Does it mean the progress bar stays at 0% for a while, and then jumps to 100% all of a sudden?

include/vapor/VaporField.h
shaomeng: My thinking was that when comparing two buffers, I could use one single function call `std::strcmp` to compare the content of two buffers, instead of calling approximately a dozen individual comparisons. A small advantage is that it's also faster this day.

Regarding to documetation, I've clearly documented what offset stores what information of what size. Any suggestion on further documentation is welcome.

lib/flow/Advection.cpp
shaomeng: more explanation is added.

lib/flow/Advection.cpp
shaomeng: A more descriptive comment is added.

lib/flow/VaporField.cpp
shaomeng: I actually think this is a case hard coded values are more suitable, because there are consecutive memcpy() called with offset values determined by the size value from a previous memcpy() function call.

lib/render/FlowRenderer.cpp
shaomeng: more document is added.

lib/render/FlowRenderer.cpp
shaomeng: Good call!

include/vapor/VaporField.h
clyne: @shaomeng please see, for example,

https://stackoverflow.com/questions/28858359/should-i-use-memcmp-or-chained-equal-to-operations-when-both-give-the-same-resul

for one discussion on how to handle class equality operators. Of course you could simply just define a member-wise equality operator, but it is certainly possible to use std::memcmp in a way that follows best programming practices (e.g. avoids having to know the size of each data element, results in more maintainable code, etc.)

shaomeng: @clyne 

The discussion seems comprehensive, but I don't take anything away that I think I could use in my use case. The current implementation is using std::memcmp to directly comparing two memory buffers. This approach results in fairly readable code, and should produce the highest performance,  but the downside is that the hard coded memory offsets look ugly. In this particular case I decided that it's an acceptable price to pay. Going back to the discussion thread, maybe I'm missing something? If so, please specify.

clyne:
@shaomeng there does not appear to be any need or benefit from manually doing the memory placement, and the result is error-prone, difficult to read code. Please just use a struct as @stasj suggested and let the compiler to the work. That's what it is for :-)

shaomeng:
> @shaomeng there does not appear to be any need or benefit from manually doing the memory placement, and the result is error-prone, difficult to read code. Please just use a struct as @StasJ suggested and let the compiler to the work. That's what it is for :-)

To make sure I understand the requirement, the suggested approach is to do individual value comparisons for approximately a dozen times, is it correct?

clyne:
> > @shaomeng there does not appear to be any need or benefit from manually doing the memory placement, and the result is error-prone, difficult to read code. Please just use a struct as @StasJ suggested and let the compiler to the work. That's what it is for :-)
> 
> To make sure I understand the requirement, the suggested approach is to do individual value comparisons for approximately a dozen times, is it correct?

For a generic class you could define an "==" operator that compares the individual members. For a class (struct) of PODs like you have you could also simply compare classes with memcmp if you want to avoid an equality test for each individual member. E.g.

```
struct {
  int a;
  float b[3]
  ..
 } MyClass

bool isEqual(const MyClass &a, const MyClass &b) {
  return(std::memcmp(&a,&b,sizeof(MyClass));
}
```
The salient point is that there is no need to pack the data into a buffer and manually handle the memory offsets

shaomeng:
> > > @shaomeng there does not appear to be any need or benefit from manually doing the memory placement, and the result is error-prone, difficult to read code. Please just use a struct as @StasJ suggested and let the compiler to the work. That's what it is for :-)
> > 
> > 
> > To make sure I understand the requirement, the suggested approach is to do individual value comparisons for approximately a dozen times, is it correct?
> 
> For a generic class you could define an "==" operator that compares the individual members. For a class (struct) of PODs like you have you could also simply compare classes with memcmp if you want to avoid an equality test for each individual member. E.g.
> 
> ```
> struct {
>   int a;
>   float b[3]
>   ..
>  } MyClass
> 
> bool isEqual(const MyClass &a, const MyClass &b) {
>   return(std::memcmp(&a,&b,sizeof(MyClass));
> }
> ```
> 
> The salient point is that there is no need to pack the data into a buffer and manually handle the memory offsets

John, I'm afraid your approach isn't reliable because of memory packing. For example, in the following example, the size of the struct will be 16 instead of 12 because 4 bytes are packed in to satisfy memory alignment requirements. As a result, your approach will compare 4 bytes of memory that's not specified by anybody. 
```
struct A
{
  int a;      // 4 bytes
  long b;     // 8 bytes
}
```

shaomeng:
> The salient point is that there is no need to pack the data into a buffer and manually handle the memory offsets

Hmm, compilers are very well allowed to shuffle data members inside of a class, and many of them so so. For example, if a struct is defined with data members provided in the order of (a, b, c),
```
struct A
{
  int a;
  long b;
  int c;
}
```
when a certain level of optimization is turned on, it'll almost certainly re-organize to be in the order of (a, c, b). In other words, manually handle the memory offsets is the only way to make sure something is stored at a certain location.

clyne:
The optimizations should be consistent in your use case. But if you are concerned define an overloaded equality operator that does member-wise comparison. That is the C++ convention for comparing classes.

shaomeng:
> The optimizations should be consistent in your use case. But if you are concerned define an overloaded equality operator that does member-wise comparison. That is the C++ convention for comparing classes.

A member-wise comparison implementation is now pushed.

shaomeng:
> The optimizations should be consistent in your use case. 

I cannot agree with this statement. The packed memory contains unspecified values, most likely values from the last time that chunk of memory was used. A comparison of equality of those values simply doesn't make sense.

shaomeng:
> > The optimizations should be consistent in your use case.
> 
> I cannot agree with this statement. The packed memory contains unspecified values, most likely values from the last time that chunk of memory was used. A comparison of equality of those values simply doesn't make sense.

lib/flow/VaporField.cpp
```diff
@@ -1,37 +1,179 @@
...
+        return false;
+    if( this->timestep != other.timestep )
+        return false;
+    if( this->refLev   != other.refLev )
+        return false;
+    if( this->compLev  != other.compLev )
+        return false;
+    if( this->defaultZ != other.defaultZ)
+        return false;
+    for( int i = 0; i < 3; i++ ) {
```
clyne, line 47: Just an FYI that std::vector (and every other container I'm aware of) supports a '==' operator. I.e. you can just say:

`if( this->ext_min != other.ext_min )`

and avoid the element-wise comparison

clyne (approved):

lib/flow/VaporField.cpp
shaomeng: Well, I store the 3-element structure as a std::array instead of a std::vector, thus cannot use the built in == operator.

The reason to not use std::vector is to avoid 2 heap memory allocations, each 3-element in size. This decision was part of the effort to optimize grid retrieval from the grid cache.

clyne: Are you sure std::array doesn't support '=='? It looks like it does:

https://en.cppreference.com/w/cpp/container/array

shaomeng: You're right. I thought I was comparing a vector and an array. Now it's fixed, thanks for the reminder ;)

lib/render/FlowRenderer.cpp
StasJ: Yes, but it lets user know vapor has not frozen and you could still use it to allow the user to cancel a flow calculation.

clyne: @shaomeng were you able to restore the progress meter for flow vis? My recollection was that you disabled it when you were refactoring.

shaomeng: yeah, I can add it back but it's gonna show only two states: 0% and 100%, because the FlowRenderer currently won't have any knowledge on how far the advection has gone. Let me bring back the semi-useful progress bar for now.

shaomeng: A progress bar is added in 646d0eca .

Issue #2091, opened 2020-01-12
Flow: allow output of multiple variables sampled along flow lines


clyne:
The current flow output includes the value of a single variable sampled along the flow line. NEed to selection of multiple variables format output. Output format should have output
variable name as well

PR #2385, opened 2020-10-14
Field values

shaomeng:
This PR resolves issue #2091 and #1979.

Note that this PR contains changes from PR #2377. That means:
1. This PR could be reviewed for functions now.
2. Code of this PR should be reviewed after PR #2377 is merged.
3. This PR should be merged after PR #2377 is merged.


clyne:
converting this to a draft since it is dependent on #2377

clyne:
On hold pending completion of variable selector, #2400

clyne:
@StasJ, @sgpearse, @clyne  please review ASAP.

lib/flow/Particle.cpp
```diff
@@ -28,66 +27,55 @@ Particle::Particle( float x, float y, float z, float t, float val )
...
-
-int
-Particle::GetNumOfProperties() const
+    
+void Particle::RemoveProperty( size_t target_i )
 {
-    return (_nProperties);
+    size_t current_i = 0;
+    auto before_it = _properties.cbefore_begin();
+    for( auto it = _properties.cbegin(); it != _properties.cend(); ++it ) {
```
clyne, line 53: This seems like awfully complicated code to perform what appears to be simply removing the ith member of a list. Is there a reason for this complexity, or am I missing something?

lib/render/FlowRenderer.cpp
```diff
@@ -110,66 +113,98 @@ FlowRenderer::_initializeGL()
...
+int FlowRenderer::_paintGL( bool fast )
 {
     FlowParams* params = dynamic_cast<FlowParams*>( GetActiveParams() );
     int rv;     // return value
 
     _velocityField.DefaultZ = Renderer::GetDefaultZ( _dataMgr, params->GetCurrentTimestep() );
 
-    if( params->GetNeedFlowlineOutput() )
-    {
+    if( params->GetNeedFlowlineOutput() ) {
```
clyne, line 124: The _paintGL method is getting awfully long and complex. Can you make this 'if' clause into a helper function to improve readability?

lib/render/FlowRenderer.cpp
```diff
@@ -110,66 +113,98 @@ FlowRenderer::_initializeGL()
...
+
+        // Retrieve additional variables that users require to sample
+        const auto addiVars = params->GetFlowOutputMoreVariables();
+
+        // Identify variables that an advection already has, but the user doesn't 
+        // include in addiVars. Remove them.
+        auto removeVars = _advection.GetPropertyVarNames();
```
clyne, line 131: This seems like a lot of code to simply remove duplicate variables. Maybe _advection should have a RemoveParticleProperty function that operates on variable names.

clyne (commented):
Looks good, Sam. A couple of in-line comments and one question: Is there a reason to prefer the 'trailing return type' syntax:

`auto f(int x, int y) -> int {}`

over the conventional:

`int f(int x, int y)`

The conventional is shorter, and is conventional. Even the C++ Core Guidelines: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md
continue to use the traditional syntax. Just wondering.

lib/flow/AdvectionIO.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+        size_t step = 0;
+        for( const auto& p : stream ) {
+            if( !p.IsSpecial() ) {
+                std::fprintf( f, "%lu, %f, %f, %f, %f, %f", s_idx, p.location.x, p.location.y,
+                                  p.location.z, p.time, p.value );
+
+                auto props = p.GetPropertyList();
+                // A quick sanity check
+                assert( std::distance(props.cbegin(), props.cend()) == propertyNames.size() );
```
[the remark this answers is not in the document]
shaomeng: I intend to have a sanity check when debugging, but don't want to perform that check in Release mode. Does VAssert support that use pattern?

lib/flow/Particle.cpp
shaomeng: I couldn't think of a simpler way to remove the ith member. Am I missing something?

lib/render/FlowRenderer.cpp
shaomeng: In my design advection class does one thing and only one thing, which is to advect particles. I think it's appropriate for the flow renderer to tell all the particles which particular property to remove, rather than the advection class to do so. 

Regarding your comment on lots of code, it's a so called "[erase–remove idiom](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom)" which is probably already the simplest approach to achieve the same goal. Also note that once you're familiar with this pattern, it doesn't look like lots of code anymore.

lib/render/FlowRenderer.cpp
shaomeng: > Looks good, Sam. A couple of in-line comments and one question: Is there a reason to prefer the 'trailing return type' syntax:
> 
> `auto f(int x, int y) -> int {}`
> 
> over the conventional:
> 
> `int f(int x, int y)`
> 
> The conventional is shorter, and is conventional. Even the C++ Core Guidelines: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md
> continue to use the traditional syntax. Just wondering.

There are new capabilities in C++ that would only work with a trailing return syntax. But for returning an integer as your example shows, it makes no difference. 

One benefit why I like this syntax is that it allows your function declarations in the header look well aligned. E.g., every function name starts with auto and column 6 will be the start of the function name. (Suppose you put function implementations in the .cpp file.)

StasJ (commented):
Looks good. My only note is that I agree with John's comment about the function definition syntax.

lib/render/FlowRenderer.cpp
shaomeng: > The _paintGL method is getting awfully long and complex. Can you make this 'if' clause into a helper function to improve readability?

This request is addressed in b75ef7ba .

sgpearse (commented):
~~The code looks good, but I can't seem to be able to write VaporFlow.txt, or select another file to write to.  Is anyone else experiencing this?~~

~~1) Start up vapor, and import one of Leigh's MayControl files~~
~~2) Reduce the streamline grid to 1x1x1~~
~~3) Select 'dbz' and then press "Write to file" and nothing will happen~~

I found vaporFlow.txt.  It looks like the first letter of the is being set to lower case.

When I try to select a new file, I cannot enter a new filename, only select existing files.  The PFileSelector does not seem to be using its complimentary VWidget (VFileSelector).

lib/flow/AdvectionIO.cpp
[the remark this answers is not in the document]
clyne: No, VAssert behaves the same whether compiled with or without debugging.

sgpearse:
One last thing - it looks like the color-mapped variable is written to the file by default.  We may want to check it in our variable selection list.  Or we could just not write the color mapped variable values unless the user selects that variable.

lib/flow/Particle.cpp
clyne: How about simply:

`_properties.erase_after(_properties.cbefore_begin()+target_i)`

lib/render/FlowRenderer.cpp
clyne: I'm familiar with the idiom, Sam. I was referring to the 'for' loop that comes after it.

clyne:
> One last thing - it looks like the color-mapped variable is written to the file by default. We may want to check it in our variable selection list. Or we could just not write the color mapped variable values unless the user selects that variable.

Agreed. Furthermore, if you do the color mapped variable  from the multi-variable selector it will be output twice.

clyne (commented):
Getting closer. A couple more in-line comments.

lib/flow/Particle.cpp
shaomeng: Hmm, it seems 
1. If `target_i` is out of range, it'll result in an undefined behavior? 
2. Are you sure this syntax works? It seems the iterator type of a forward_list is a `ForwardIterator`, which only supports self increment operators.  Operators such as `auto itr2 = itr1 + n` is something only a `RandomIterator` supports to my understanding. 

image

lib/render/FlowRenderer.cpp
shaomeng: the for loop after it takes every variable name to remove, and then
1) find the index that the variable name is stored in an advection object, and
2) remove a particle property at that index.

It's a little cumbersome because the identification of variable to remove is by name, while issuing the remove command is by index number, so there's a sort of conversion happening. If you like, I can change the code so that we remove a particle property by name, and we don't need to do a conversion anymore.

shaomeng:
> ~The code looks good, but I can't seem to be able to write VaporFlow.txt, or select another file to write to. Is anyone else experiencing this?~
> 
> ~1) Start up vapor, and import one of Leigh's MayControl files~
> ~2) Reduce the streamline grid to 1x1x1~
> ~3) Select 'dbz' and then press "Write to file" and nothing will happen~
> 
> I found vaporFlow.txt. It looks like the first letter of the file is being set to lower case.
> 
> When I try to select a new file, I cannot enter a new filename, only select existing files. The PFileSelector does not seem to be using its complimentary VWidget (VFileSelector).

Scott, can you remove that file and try to output again see if you experience problems? On my Ubuntu system it is capital V instead of a small v. If it still doesn't work, then I'll need to look more closely.

lib/render/FlowRenderer.cpp
shaomeng: The code has been updated in bab957fb so that it's removing particle properties by name, eliminating a name index conversion.

shaomeng:
> One last thing - it looks like the color-mapped variable is written to the file by default. We may want to check it in our variable selection list. Or we could just not write the color mapped variable values unless the user selects that variable.



> > One last thing - it looks like the color-mapped variable is written to the file by default. We may want to check it in our variable selection list. Or we could just not write the color mapped variable values unless the user selects that variable.
> 
> Agreed. Furthermore, if you do the color mapped variable from the multi-variable selector it will be output twice.

I have address this issue in 9c2aa5fc203f0a7bc6c2cab04ab96d33df9eee02 so that no duplicate output anymore. I kept the behavior that the color mapping value is always output, since I feel that's a value calculated already, why not output it too?

lib/flow/Particle.cpp
clyne: If the offset it out of range I'd guess it behaves just like any other container. You can always check the offset against the size. 

Re whether it works: my understanding is that an offset added to an iterator results in another iterator: i.e. there is no difference between incrementing an iterator by 'n' steps or adding 'n' to an iterator: they both result in an iterator that is 'n' steps from wherever you started.

clyne:
> > One last thing - it looks like the color-mapped variable is written to the file by default. We may want to check it in our variable selection list. Or we could just not write the color mapped variable values unless the user selects that variable.
> 
> > > One last thing - it looks like the color-mapped variable is written to the file by default. We may want to check it in our variable selection list. Or we could just not write the color mapped variable values unless the user selects that variable.
> > 
> > 
> > Agreed. Furthermore, if you do the color mapped variable from the multi-variable selector it will be output twice.
> 
> I have address this issue in [9c2aa5f](https://github.com/NCAR/VAPOR/commit/9c2aa5fc203f0a7bc6c2cab04ab96d33df9eee02) so that no duplicate output anymore. I kept the behavior that the color mapping value is always output, since I feel that's a value calculated already, why not output it too?

Sounds good to me.

lib/flow/Particle.cpp
shaomeng: Ok, no need to guess anymore. I tested that your suggested syntax will fail when compiling. 

The reason it fails is that it's a `ForwardIterator` rather than a `RandomAccessIterator`. The former is a subset of the latter. The behavior you described is a requirement of a `RandomAccessIterator` but not a requirement of a `ForwardIterator`.

sgpearse:
> Scott, can you remove that file and try to output again see if you experience problems? On my Ubuntu system it is capital V instead of a small v. If it still doesn't work, then I'll need to look more closely.

I see the capital V now.  But I seem to be unable to perform multiple writes.  I'm on OSX, using the DUKU dataset right now.  I can help test with other configurations.

Case 1
1. Delete VaporFlow.txt
2. Create a default flow renderer, and write VaporFlow.txt with the first variable in the list (P in DUKU's case)
3. Note the file's timestamp
4. Wait a minute and try to write the file again, and the timestamp does not update

Case 2
1. Delete VaporFlow.txt
2. Create a default flow renderer, and write VaporFlow.txt with the first variable in the list (P in DUKU's case)
3. Delete VaporFlow.txt again
4. Write VaporFlow.txt again, but it does not seem to write

lib/flow/Particle.cpp
clyne: Makes sense. Now we know. Thanks for checking.

clyne (approved):

shaomeng:
> > Scott, can you remove that file and try to output again see if you experience problems? On my Ubuntu system it is capital V instead of a small v. If it still doesn't work, then I'll need to look more closely.
> 
> I see the capital V now. But I seem to be unable to perform multiple writes. I'm on OSX, using the DUKU dataset right now. I can help test with other configurations.
> 
> Case 1
> 
>     1. Delete VaporFlow.txt
> 
>     2. Create a default flow renderer, and write VaporFlow.txt with the first variable in the list (P in DUKU's case)
> 
>     3. Note the file's timestamp
> 
>     4. Wait a minute and try to write the file again, and the timestamp does not update
> 
> 
> Case 2
> 
>     1. Delete VaporFlow.txt
> 
>     2. Create a default flow renderer, and write VaporFlow.txt with the first variable in the list (P in DUKU's case)
> 
>     3. Delete VaporFlow.txt again
> 
>     4. Write VaporFlow.txt again, but it does not seem to write

I observe the behavior you reported too. It seems in some occasions clicking the "Write To File" button alone isn't enough to trigger a paint event. Something else needs to change to trigger a paint event, for example, if you rotate the scene. I.e., after you rotate the scene, and then click on that button, the output function is called. I think it's a bug, but probably out of the scope of this PR.

sgpearse:
> I think it's a bug, but probably out of the scope of this PR.

No problem - I'll make a ticket and accept the PR.

sgpearse (approved):

StasJ (approved):

PR #2392, opened 2020-10-21
Issue 927: Add support for vertical coordinate transforms 

### LINKED ISSUE #927: Add support for vertical coordinate transforms
Add support to convert from native vertical coordinate (e.g. pressure, Z, hybrid, etc.) to meters. In particular, transforms for vertical coordinate systems used by MOM, ROMS, and CAM are needed

### LINKED ISSUE #1927: regular grids defined in NetCDF are treated as stretched grids

clyne:
Closes #927 

This PR closes #1927, Add support for vertical coordinate transforms. Specifically, it adds for Atmosphere hybrid sigma pressure coordinates used by CAM (e.g. AdamPhilips/F2000C5_new120_025) and other data. 

This PR also generalizes the LayeredGrid class so that it supports non-uniformly sampled (stretched) horizontal coordinates. This was necessary to support the grid resulting from the conversion of Atmospheric hybrid coordinates to meters.

### DIFF
--- a/lib/vdc/DerivedVar.cpp
+++ b/lib/vdc/DerivedVar.cpp
@@ -3160,24 +3160,19 @@ int DerivedCoordVarStandardAHSPC::ReadRegion(
 	if (rc<0) return(rc);
 
 
+	float p0 = 1.0;
 	if (! _aVar.empty()) {
-		float p0;
 		myMin = {};
 		myMax = {};
 		rc = _getVar(
 			_dc, f->GetTS(), _p0Var, f->GetLevel(), f->GetLOD(),
 			myMin, myMax, &p0
 		);
+	}
 
-		compute_a(
-			min, max, a.data(), b.data(), ps.data(), p0, region
-		);
-	}
-	else {
-		compute_ap(
-			min, max, a.data(), b.data(), ps.data(), region
-		);
-	}
+	compute_a(
+		min, max, a.data(), b.data(), ps.data(), p0, region
+	);
 
 	return(0);
 }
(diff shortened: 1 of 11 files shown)


lib/vdc/GridHelper.cpp
```diff
@@ -276,12 +278,11 @@ LayeredGrid *GridHelper::_make_grid_layered(
...
-	for (int i = 0; i<2; i++) { 
-		float *coords = blkvec[i+1];
-		hminu.push_back(coords[0]);
-		hmaxu.push_back(coords[dims[i]-1]);
-	}
+	vector <double> xcoords;
+	for (int i=0; i<dims[0]; i++) xcoords.push_back(blkvec[1][i]);
+
+	vector <double> ycoords;
+	for (int i=0; i<dims[1]; i++) ycoords.push_back(blkvec[2][i]);
```
shaomeng, line 285: Depending on if this piece of code is performance critical, you could call the reserve() method of the vector before pushing back a bunch of values.

lib/vdc/LayeredGrid.cpp
```diff
@@ -129,98 +123,148 @@ void LayeredGrid::GetBoundingBox(
...
+			//
+			float zk = 
+				_zrg.AccessIJK(iv[0], jv[0], kk) * lambda[0] +
+				_zrg.AccessIJK(iv[1], jv[1], kk) * lambda[1] +
+				_zrg.AccessIJK(iv[2], jv[2], kk) * lambda[2];
+
+				zcoords.push_back(zk);
+		}
+
+		if (! Wasp::BinarySearchRange(zcoords, coords[2], indices[2])) return(false);
```
shaomeng, line 208: I believe it's better readability and maintenance if you call one of the std binary search functions. That is, std::lower_bound, std::upper_bound, or std::binary_search.

shaomeng (approved):
A comment in general, std::vector seems to be used pretty intensively still. Given that we just had a campaign to use std::array in applicable scenarios, I wonder if some of the std::vectors could be converted to be std::arrays ?

lib/vdc/DerivedVar.cpp
```diff
@@ -2899,3 +2899,390 @@ void DerivedCoordVarStandardOceanSCoordinate::compute_g2(
...
+    for (size_t j=0; j<max[1]-min[1]+1; j++) {
+    for (size_t i=0; i<max[0]-min[0]+1; i++) {
+
+		// We are deriving coordinate values from data values, so missing
+		// values may be present
+		//
+		float l_ps = ps[j*rDims[0]+i];
+		if (l_ps == _psVarMV) l_ps = 0;
+
+		float pressure = (a[k]*p0) + (b[k]*l_ps);
```
StasJ, line 3241: `compute_a` and `compute_ap` are identical other than the multiplication by `p0` here. You could just have one function with `p0` as an optional parameter initialized to `1`.

lib/vdc/GridHelper.cpp
clyne: Yup. Thanks. But the constructors in general were not designed to be performant. If the need ever arrises lots of changes will be required :-)

lib/vdc/DerivedVar.cpp
clyne: Doh! Good catch. Fixed.

lib/vdc/LayeredGrid.cpp
clyne: Ha ha. I did this, but the implementation with lower_bound/upper_bound is actually more code than what I had. Oh well. I've committed it.

StasJ (approved):

lib/vdc/LayeredGrid.cpp
shaomeng: Well, my understanding is that a user (e.g. a LayeredGrid) directly uses binary search functionality provided by std::lower_bound or std::upper_bound, but not use a wrapper (e.g. wasp::BinarySearchRange). Or in other words, if no new code uses wasp::BinarySearchRange() then that wrapper can go away after a few release cycles. Anyway, it's a minor issue.

PR #2395, opened 2020-10-23
fix circular dependency

sgpearse:
This PR gets around the circular dependency between ImageParams and the ImageRenderer that is seen in the renderer_gui_refactor PR.

This fixes the issue by defining three shared functions in GeoImageTMS in the file GeoImageTMS.h, instead of the .cpp file.

### DIFF
--- a/include/vapor/GeoImageTMS.h
+++ b/include/vapor/GeoImageTMS.h
@@ -30,9 +30,54 @@ public:
  GeoImageTMS();
  virtual ~GeoImageTMS();
 
- static bool IsTMSFile( std::string path );
- static std::string TilePath( std::string file, size_t tileX, size_t tileY, int lod );
- static int GetNumTMSLODs( std::string file );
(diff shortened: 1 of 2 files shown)


clyne:
@sgpearse a better way to resolve dependencies like this is to push the dependency lower down the dependency hierarchy. The lib/common library exists for this purpose: providing API needed by higher level functions. You can simply make your static GeoImageTMS functions free functions in lib/common. Put them in their own file if you'd like (*e.g. TMSHelper.{h,cpp}). 

There is lots of discussion on when and when not to put function definitions in a .h file such as:

https://stackoverflow.com/questions/583255/is-it-a-good-practice-to-place-c-definitions-in-header-files

But the real issue here is that both lib/params and lib/render require these functions, and the dependency order is render -> params -> common. You could also put these functions in lib/params, but it's really not parameter related function IMHO.

sgpearse:
Understood.  Will fix shortly.

PR #2399, opened 2020-10-23
Fixed #2398 - Support 3D variables from MPAS-O data

### LINKED ISSUE #2398: Support 3D variables from MPAS-O data
VAPOR currently does not support 3D variables from MPAS-Ocean. It can only read 2D variables.

clyne:
Closes #2398 

This PR adds support for reading 3D variables from MPAS-Ocean model outputs. Previously, only the 2D variables were imported.

@stasj, @shaomeng FYI I shoe-horned this largely cut-n-paste PR in because I hope to use MPAS-O data for a talk I'm giving at SC20.

### DIFF
--- a/lib/vdc/DCMPAS.cpp
+++ b/lib/vdc/DCMPAS.cpp
@@ -81,10 +81,12 @@ namespace {
 
 	// Vertical coordinate variables
 	//
-	const string zGridP1VarName = "zgrid";
+	const string zGridP1VarName = "zgrid"; // atmosphere vertical coordinate
+	const string zTopVarName = "zTop";	// ocean vertical coordinate
 
 	const vector <string> optionalVertCoordVarNames = {
-		zGridP1VarName
+		zGridP1VarName,
+		zTopVarName
 	};
 
 	// Time coordinate variables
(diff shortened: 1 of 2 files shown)


StasJ (approved):

lib/vdc/DCMPAS.cpp
```diff
@@ -1304,10 +1347,10 @@ bool DCMPAS::_HasVertical(
 
 		itr = find(varnames.begin(), varnames.end(), s);
 		if (itr == varnames.end()) {
-			return(false);
+			return(true);
 		}
 	} 
-	return(true);
+	return(false);
```
shaomeng, line 1353: Is this logic correct? If the iterator equals to the end, it usually means not finding that element.

shaomeng (commented):
I think there might be a logic flaw, but I don't understand the code enough to have a definitive answer. So, maybe you'll double check, and if it's correct, then this PR will be ready for merge.

lib/vdc/DCMPAS.cpp
clyne: Good catch! Thanks!

shaomeng (approved):

PR #2402, opened 2020-10-26
Fixed #2307 - vdccreate -ncvars2dxy is broken

### LINKED ISSUE #2307: vdccreate -ncvars2dxy is broken
For all the tests, the following code was run:
```
vdccreate -dimension 64x64x1 -[VAR_TYPE] test test.vdc
raw2vdc -varname test test.vdc data.raw
```

**[data.raw.zip](https://github.com/NCAR/VAPOR/files/4907917/data.raw.zip)**

With the following options for `VAR_TYPE`
- `vars3d`
- `ncvars3d`
- `vars2dxy`
- `ncvars2dxy`

All except for `ncvars2dxy` are correct

Correct:

good

`ncvars2dxy`

bad

clyne:
Fixed #2307

### DIFF
--- a/apps/raw2vdc/raw2vdc.cpp
+++ b/apps/raw2vdc/raw2vdc.cpp
@@ -149,23 +149,23 @@ int	main(int argc, char **argv) {
 	ProgName = FileUtils::LegacyBasename(argv[0]);
 
 	if (op.AppendOptions(set_opts) < 0) {
-		exit(1);
+		return(1);
 	}
 
 	if (op.ParseOptions(&argc, argv, get_options) < 0) {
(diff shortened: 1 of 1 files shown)


StasJ (approved):
The problem appears to be fixed but does this PR fix it or was it already fixed? It only changes exit calls to return calls?

clyne:
Changing exits to returns in main does indeed fix the problem. When you exit from main (as opposed to return) class destructors for locally scoped variables are not implicitly. In this case that resulting in the NetCDF library not being closed.

StasJ:
@clyne that makes sense, thanks.

PR #2405, opened 2020-10-27
Constrain when V*SliderEdit emits its signals

### LINKED ISSUE #2390: Changing the slider minimum value will result in the minimum changed signal being emitted 10 times
This will cause 10 undo steps and 10 renders which can potentially be very costly.

To reproduce, change the minimum value for the Volume ambient lighting slider in the renderer_gui_refactor branch.

sgpearse:
Fixes #2390, which reported that the V*SliderEdit emits unnecessary signals.

### DIFF
--- a/apps/raw2vdc/raw2vdc.cpp
+++ b/apps/raw2vdc/raw2vdc.cpp
@@ -149,23 +149,23 @@ int	main(int argc, char **argv) {
 	ProgName = FileUtils::LegacyBasename(argv[0]);
 
 	if (op.AppendOptions(set_opts) < 0) {
-		exit(1);
+		return(1);
 	}
 
 	if (op.ParseOptions(&argc, argv, get_options) < 0) {
-		exit(1);
+		return(1);
 	}
 
 	if (opt.help) {
 		cerr << "Usage: " << ProgName << " [options] vdcFile rawDataFile" << endl;
 		op.PrintOptionHelp(stderr);
-		exit(0);
+		return(0);
 	}
 
 	if (argc != 3) {
 		cerr << "Usage: " << ProgName << " [options] vdcFile rawDataFile" << endl;
 		op.PrintOptionHelp(stderr);
-		exit(1);
+		return(1);
 	}
 
 	string master = argv[1];	// Path to VDC master file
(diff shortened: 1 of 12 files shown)


StasJ (changes requested):
1. Pick any double slider
1. Set the value to the max
1. Right click and change the max to half the current max

The changed signal is emitted twice.

sgpearse:
This is a good catch.  I believe the correct fix lies with the P*SliderEdit.  

The current PR changes the P*SliderEdit::valueChanged() function to account for all situations where multiple V*SliderEdit signals are emitted; within a a ParamsGroup.

StasJ (changes requested):
Your changes to the PSliderEdit break its functionality. It is only supposed to change the custom range in the params if the use manually sets them. Otherwise it will never be able to update programmatically if needed.

sgpearse:
These new changes constrain the minimum and maximum range values to always include the currently selected value.  IE if the current value is 0, the user cannot change the min to 1, or max to -1.

These changes also resolve the slider bug reported in #2407

StasJ (approved):

StasJ:
#2407 is not fixed by the way. It changes the text in the textbox but the emitted value is still wrong.

sgpearse:
Hmm.  I cannot reproduce what you're seeing.

I've added a print statement in PDoubleSlierEdit for the emitted signal value and for me it aligns with what the text displays.  Are you seeing different values, or are you testing in a different way?

StasJ:
> Hmm. I cannot reproduce what you're seeing.
> 
> I've added a print statement in PDoubleSlierEdit for the emitted signal value and for me it aligns with what the text displays. Are you seeing different values, or are you testing in a different way?

When it emits the signal it is the max but it doesn't always emit the signal.

sgpearse:
I've reverted the print statement. 

The tests I'm running give correct results, but at this point #2407 should be discussed as a separate issue.

StasJ:
> I've reverted the print statement.
> 
> The tests I'm running give correct results, but at this point #2407 should be discussed as a separate issue.

I'm fine with postponing this issue. I've already approved the PR.

Note the inconsistent brightness below:

slider2

clyne:
@sgpearse this PR has been approved but isn't passing CircleCI checks. Please fix.

PR #2411, opened 2020-10-29
Multivar selector

### LINKED ISSUE #2400: multi-variable selector needed
The Python editor, Plot tool, Statistics tool, and now flow renderer, all support the selection of 0 to n variables. A multi-variable PWidget that could be used by all of these utilities would be useful. 

Requirements and comments

- It would be good to be configurable to support 2D, 3D, or 2D+3D lists of variables, the latter combining both 2D and 3D lists

- It should be possible to select (or deselect) 0 to N variables. 

- The current statistics and plot tools are pretty clunky; if the user wants to select multiple variables the selector has to be opened and closed for each variable. Removing variables is also a PIA. 

- The Python variable selector is a little more user friendly in that a list of variables is presented with a set of checkboxes so variables can easily be selected/deselected. 

@sgpearse @shaomeng, comments?

StasJ:
Fix #2400

VarSelector

### DIFF
--- a/apps/vaporgui/PMultiVarSelector.h
+++ b/apps/vaporgui/PMultiVarSelector.h
@@ -8,6 +8,8 @@ class QListWidgetItem;
 //! \class PMultiVarSelector
 //! \brief Allows the selection of multiple variables. To add a title, use a PLabel.
 //! \author Stas Jaroszynski
+//! By default this shows the vars with the same dimension as the renderer. This can be changed
+//! with DisplayVars(Mode)
 
 class PMultiVarSelector : public PWidget {
 public:
--- a/apps/vaporgui/PMultiVarSelector.h
+++ b/apps/vaporgui/PMultiVarSelector.h
@@ -10,10 +10,11 @@ class QListWidgetItem;
 //! \author Stas Jaroszynski
 
 class PMultiVarSelector : public PWidget {
-    QListWidget *_listWidget;
-    
 public:
+    enum Mode { Auto, D2, D3, Both };
+    
     PMultiVarSelector(std::string tag);
+    PMultiVarSelector *DisplayVars(Mode mode);
     
 protected:
     void updateGUI() const override;
@@ -21,6 +22,9 @@ protected:
     bool requireDataMgr() const override { return true; }
     
 private:
+    QListWidget *_listWidget;
+    Mode _mode = Auto;
+    
     QListWidgetItem *addVarToList(const std::string &var) const;
     void itemChanged(QListWidgetItem*);
 };
--- a/apps/vaporgui/PMultiVarSelector.h
+++ b/apps/vaporgui/PMultiVarSelector.h
@@ -1,0 +1,26 @@
+#pragma once
+
+#include "PWidget.h"
+
+class QListWidget;
+class QListWidgetItem;
+
+//! \class PMultiVarSelector
+//! \brief Allows the selection of multiple variables. To add a title, use a PLabel.
+//! \author Stas Jaroszynski
+
+class PMultiVarSelector : public PWidget {
+    QListWidget *_listWidget;
+    
+public:
+    PMultiVarSelector(std::string tag);
+    
+protected:
+    void updateGUI() const override;
+    QSize minimumSizeHint() const override;
+    bool requireDataMgr() const override { return true; }
+    
+private:
+    QListWidgetItem *addVarToList(const std::string &var) const;
+    void itemChanged(QListWidgetItem*);
+};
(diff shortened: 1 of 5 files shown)


clyne:
@StasJ is this thing integrated into anything that we can test it with?

StasJ:
@clyne Just add `new PMultiVarSelector(“test”),` to any renderer GUI.

apps/vaporgui/PMultiVarSelector.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "PWidget.h"
+
+class QListWidget;
+class QListWidgetItem;
+
+//! \class PMultiVarSelector
```
clyne, line 8: Please add documentation wrt how the class decides what variables to present.

clyne (commented):
@StasJ it may be limiting to have the class determine it's dimensionality via RenderParams::GetRenderDim(). There are instances where even though renderer operating in 2D or 3D mode, it might make sense to present variables of different dimensionality. For example, the Flow renderer will use this to allow the user to identify variables to be sampled along a trajectory. The advection might use 3D variables, but it would also make sense to the  user to sample 2D variables. Also, the statistic tool doesn't operate in one mode or the other. 2D and 3D variables can both be selected.Ditto for the Python editor. Any thoughts on how to support these use cases?

shaomeng:
> @StasJ it may be limiting to have the class determine it's dimensionality via RenderParams::GetRenderDim(). There are instances where even though renderer operating in 2D or 3D mode, it might make sense to present variables of different dimensionality. For example, the Flow renderer will use this to allow the user to identify variables to be sampled along a trajectory. The advection might use 3D variables, but it would also make sense to the user to sample 2D variables. Also, the statistic tool doesn't operate in one mode or the other. 2D and 3D variables can both be selected.Ditto for the Python editor. Any thoughts on how to support these use cases?

We have decided a while ago that distinguishing 2D vs. 3D variables presents multiple issues and avoidable complexities regarding how renderers process these variables. A proposed solution was to treat every grid as 3D unless there's a rare circumstance where it really matters to be 2D but not 3D with Z length equals 1. See #2250 .

shaomeng (approved):

apps/vaporgui/PMultiVarSelector.h
StasJ: Done

StasJ:
> > @StasJ it may be limiting to have the class determine it's dimensionality via RenderParams::GetRenderDim(). There are instances where even though renderer operating in 2D or 3D mode, it might make sense to present variables of different dimensionality. For example, the Flow renderer will use this to allow the user to identify variables to be sampled along a trajectory. The advection might use 3D variables, but it would also make sense to the user to sample 2D variables. Also, the statistic tool doesn't operate in one mode or the other. 2D and 3D variables can both be selected.Ditto for the Python editor. Any thoughts on how to support these use cases?
> 
> We have decided a while ago that distinguishing 2D vs. 3D variables presents multiple issues and avoidable complexities regarding how renderers process these variables. A proposed solution was to treat every grid as 3D unless there's a rare circumstance where it really matters to be 2D but not 3D with Z length equals 1. See #2250 .

Done.

apps/vaporgui/PMultiVarSelector.h
```diff
@@ -0,0 +1,32 @@
...
+    void updateGUI() const override;
+    QSize minimumSizeHint() const override;
+    bool requireDataMgr() const override { return true; }
+    
+private:
+    QListWidget *_listWidget;
+    Mode _mode = Auto;
+    
+    QListWidgetItem *addVarToList(const std::string &var) const;
+    void itemChanged(QListWidgetItem*);
```
shaomeng, line 31: The convention says private class method names start from an underscore. 

Should it be enforced? I don't have a strong opinion.

StasJ: You are correct. I have been writing all of my code with pascal case for public and camel case for private/protected functions and _camel for private/protected vars. I find it is useful to have a distinction between private variables and functions. I think it is a minor difference to someone reading the code, especially since my convention shares the same rules in its public interface. Much of the legacy code follows no convention at all so it is not a black sheep in the codebase.

clyne: @StasJ please follow the convention we've adopted.

clyne (commented):
Please see inline comments. Nice work!

sgpearse (approved):

apps/vaporgui/PMultiVarSelector.h
StasJ: Done. Just FYI, @clyne @shaomeng both of your last PRs also violate the Vapor coding convention so we may want to look into whether we want to revise it.

Issue #2155, opened 2020-01-24
Paint Fails if Flow renderer only has one variable


jvalan:
1. Load up flow renderer usiing a dataset that has one or two variable(s).
2. If dataset has only one variable, a paint fail error should appear.
3. If dataset has two variables, go to variables and set x or y to "none". Paint will fail.

### Desktop
 - OS and version: red hat linux enterprise 7
 - Version:  Vapor 3.2.0

sgpearse:
Hi @jvalan, this is not a bug.  Flow requires at least two vector fields to perform advection.  The error message should indicate this.

We haven't had time to sort through your other reports, but we will get to them soon.

shaomeng:
The current code has the following behavior:

- In case only 1 variable is provided, a window pops up saying that you need to provide at least 2 variables.
- In case of 2 variables are provided, advection is performed and painted on 2D planes.

I think this is a reasonable behavior. @clyne thoughts?

clyne:
@shaomeng I would suggest having the flow rendering treat unspecified variables (variable name = "none") consistently with the other renderers in VAPOR. I.e. an unspecified variable is treated as a constant field of zeros, not an error condition. This can be useful for debugging in some instances, and it's good to keep things consistent.

PR #2415, opened 2020-10-29
Issue 2155

shaomeng:
This PR addresses issue #2155.

Note: this PR should be merged after PR #2385 is merged.


shaomeng:
This branch is not able to be fixed with too many conflicts.

shaomeng:
Well, the conflicts are actually fixed. Now converting to a normal PR.

clyne (approved):

sgpearse (approved):
I think this is resolving the reported issue.  However I think this might be uncovering an unexpected use case that we haven't tested before. 

Changing the number of <none> variables seems to affect the gridded seed distribution.  I think this bug lies outside of this PR, which is only addressing a warning.  I can file another bug report and maybe we can fix that if we feel like it.

To reproduce:

1. Load our current DUKU VDC and create a Flow renderer
2. In the Seeding tab, reduce the Integration Steps to 10
3. Enable the renderer and note the 3x3x3 grid of seeds (we should see 5x5x5?)
4. In the Variables tab, set the V and W variables to <none>
5. Note that we now see a 5x3x4 grid of seeds (we should see 5x5x5?).

clyne:
> I think this is resolving the reported issue. However I think this might be uncovering an unexpected use case that we haven't tested before.
> 
> Changing the number of variables seems to affect the gridded seed distribution. I think this bug lies outside of this PR, which is only addressing a warning. I can file another bug report and maybe we can fix that if we feel like it.
> 
> To reproduce:
> 
> 1. Load our current DUKU VDC and create a Flow renderer
> 2. In the Seeding tab, reduce the Integration Steps to 10
> 3. Enable the renderer and note the 3x3x3 grid of seeds (we should see 5x5x5?)
> 4. In the Variables tab, set the V and W variables to
> 5. Note that we now see a 5x3x4 grid of seeds (we should see 5x5x5?).

@sgpearse I agree that this is probably outside of this PR, but please do submit a bug report. Thanks!

PR #2416, opened 2020-10-29
Fixed #2413 - Missing values not handled correctly in ROMS data

### LINKED ISSUE #2413: Missing values not handled correctly in ROMS data
Missing values are not being treated as such in some data sets

To reproduce:
1. Load Data/CoupledWrfRoms/kiost_roms_nwp12_avg_y1982-002.nc.
2 Create and enable a slice renderer with the "salt" variable (or any other). Note that the upper bounds of the TFE go up to 1e+37. This is the missing value and should be ignored.

clyne:
Fixed #2413

### DIFF
--- a/lib/vdc/Grid.cpp
+++ b/lib/vdc/Grid.cpp
@@ -139,6 +139,15 @@ void Grid::GetRange(float range[2]) const
     float missingValue = GetMissingValue();
     Grid::ConstIterator itr    = this->cbegin();
     Grid::ConstIterator enditr = this->cend();
+
+	// Edge case: all values are missing values.
+	//
+	range[0] = range[1] = missingValue;
+	while (*itr == missingValue && itr != enditr) {
+		++itr;
+	}
+	if (itr == enditr) return;
+
     range[0] = *itr;
     range[1] = range[0];
     while( itr != enditr )


lib/vdc/Grid.cpp
```diff
@@ -139,6 +139,15 @@ void Grid::GetRange(float range[2]) const
...
     Grid::ConstIterator enditr = this->cend();
+
+	// Edge case: all values are missing values.
+	//
+	range[0] = range[1] = missingValue;
+	while (*itr == missingValue && itr != enditr) {
+		++itr;
+	}
+	if (itr == enditr) return;
+
```
shaomeng, line 150: Three comments. 

1. This loop can be simplified by making use of an std algorithm: `std::all_of()`. 

2. This fix seems to break the logic of the rest of the function. Namely, if there exists non-missingVal elements, then `itr` will be pointing at an arbitrary location, and you won't find the true min and max. 

3. The code to look for min and max elements could be simplified by making use of another std algorithm: `std::minmax_element()`

shaomeng (changes requested):

lib/vdc/Grid.cpp
clyne: > Three comments.
> 
> 1. This loop can be simplified by making use of an std algorithm: `std::all_of()`.

The Grid operators unfortunately are not STL C++ operators, so they can't be used with the STL.

> 2. This fix seems to break the logic of the rest of the function. Namely, if there exists non-missingVal elements, then `itr` will be pointing at an arbitrary location, and you won't find the true min and max.

Hmm. Are you sure? I don't see how 'itr' will be pointing at an invalid location. AFter the first check for leading missing values 'itr' is either at the end of the list, or pointing to a value, no?


> 3. The code to look for min and max elements could be simplified by making use of another std algorithm: `std::minmax_element()`

Same comment as above.

shaomeng: Well, as long as the grid iterator meets a set of requirements, those iterators can be used in STL algorithms. A compiler can tell you if the iterator you give to it meets those requirements. That being said, it's only a suggestion to use STL algorithms.

Regarding the logic, you're right, it's not going to break anything. Now the PR is ready to merge!

shaomeng (approved):

PR #2417, opened 2020-11-03
Ospray3rd party

sgpearse:
This PR points site.NCAR at Ospray library targets.

CircleCI is up to date with Ospray, and this should allow ospray2-gui2 to pass Circle's tests.


StasJ (changes requested):
Can we make the ospray dir just “ospray” or “ospray2” so we don’t have to update the site file for every version change?

sgpearse:
We can.  It will take some more work, but I understand that we need to update this thing.

StasJ (approved):

shaomeng:
It looks like the same line is inserted to all 3 different conditions. Why don't have a single line that's out of the `if` block, so it's set for conditions?

PR #2421, opened 2020-11-06
Fixed #2225 - Broken Image Renderer

### LINKED ISSUE #2225: Broken Image Renderer
To reproduce:
1) Open Vapor, import the WRF file in /little_glade/bugs/2225
2) Create and enable an image renderer

Screen Shot 2020-02-10 at 9 06 00 AM

clyne:
Fixes #2225. Note that the problematic data set has a very small geographic region that maps to a single pixel in the NaturalEarth image bundled with VAPOR. The image renderer was not handling this edge case. Even though this PR fixes the problem, the image map still only covers a single pixel, so the results are not very thrilling.

### DIFF
--- a/include/vapor/GeoTile.h
+++ b/include/vapor/GeoTile.h
@@ -246,7 +246,7 @@ public:
 
  //! Compute the size of a map in pixels at a specified lod and coordinates
  //!
- //! Given a rectangular boundary described to two points (the north-west and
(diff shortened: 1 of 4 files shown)


sgpearse (approved):
Looks good and well explained.  Seems like a rare use case for now but now we are future proofed for higher res maps.

PR #2428, opened 2020-11-07
#2201 Slice Renderers look different after reloading session file

### LINKED ISSUE #2201: Slice Renderers look different after reloading session file
### Steps: (Updated)
1. Open VAPOR, load CDF compliant data: `/little_glade/shaomeng/cm1.nc`
2. Create a slice renderer. Set the Z location to 3.7 in the Geometry tab. Enable this renderer. 
3. Repeat step 2, with a different Z location value: 6.
3. Repeat step 2, with a different Z location value: 10.
4. Repeat step 2, with a different Z location value: 16.
4. Rotate the scene and observe the 4 planes that are rendered.
3. Save the current state into a session file.
3. Close VAPOR, and then re-open VAPOR, and load the saved session file.
3. Rotate the scene and observe the 4 planes that are rendered. They look different from step 6.

Platform: cisl-vapor
Version: Official 3.2 release. 

Before:
good

After:
bad

StasJ:
Fix #2201

### DIFF
--- a/include/vapor/SliceParams.h
+++ b/include/vapor/SliceParams.h
@@ -53,7 +53,8 @@ public:
  static const string SampleLocationTag;
     
 private:
-
+    bool _initialized = false;
+    
  void _init();
  std::vector<double> _cachedValues;
 
--- a/lib/params/SliceParams.cpp
+++ b/lib/params/SliceParams.cpp
@@ -39,7 +39,8 @@ SliceParams::SliceParams(
 
 SliceParams::SliceParams(
 	DataMgr *dataMgr, ParamsBase::StateSave *ssave, XmlNode *node
-) : RenderParams(dataMgr, ssave, node, THREED) {}
+) : RenderParams(dataMgr, ssave, node, THREED)
+{_initialized = true;}
 
 SliceParams::~SliceParams() {
 	SetDiagMsg("SliceParams::~SliceParams() this=%p", this);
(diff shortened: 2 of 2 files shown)


clyne:
@StasJ this may fix the problem, but it doesn't follow the programming model used by all of the other param classes, and may lead to inconsistent behavior. The params classes have two constructors, one that takes an XmlNode pointer as a argument and one that does not. The one that does NOT is used to initialize the class from scratch. The one that does is used to initialize the class from an existing Xml data base (session file). For some reason SliceParams don't follow this model. Note, that an Initialize() method that has been added to SliceParams that is not used by any other RenderParams class. I'm not sure why. @sgpearse may know more.

StasJ:
@clyne Unfortunately, because of how `RenderParams` is implemented, we need to use the `Initialize` function here. I can work around the issues by setting a flag to initialize only when creating a new params object from scratch.

clyne:
> @clyne Unfortunately, because of how `RenderParams` is implemented, we need to use the `Initialize` function here. I can work around the issues by setting a flag to initialize only when creating a new params object from scratch.

@StasJ using an initialization flag is fine, but I'm curious as to what is different about SliceParams that it requires use of the ::Initialize() method at all?

StasJ:
> > @clyne Unfortunately, because of how `RenderParams` is implemented, we need to use the `Initialize` function here. I can work around the issues by setting a flag to initialize only when creating a new params object from scratch.
> 
> @StasJ using an initialization flag is fine, but I'm curious as to what is different about SliceParams that it requires use of the ::Initialize() method at all?

The RenderParams Initialize function sets the orientation to XY for 2D renderers and XYZ for 3D renderers. The Slice renderer requires a 2D orientation despite being a 3D renderer so this breaks it. The RenderParams initialize function is called after the constructor so if Slice params sets its orientation in the constructor, it will be overridden.

clyne (approved):

PR #2430, opened 2020-11-09
time and geo coordinate conversion

shaomeng:
This PR implements time and geo coordinate conversions when writing flow lines to a file. 

@clyne The time conversion seems to be obviously correct, but I'm not 100% sure the geo coordinate conversion result is also correct and easy to read. Could you take a look and maybe provide feedback?

### DIFF
--- a/include/vapor/AdvectionIO.h
+++ b/include/vapor/AdvectionIO.h
@@ -33,6 +33,7 @@ namespace flow
     // Input a list of seeds from lines of CSVs.
     auto InputSeedsCSV( const std::string& filename, 
                         Advection*         adv ) -> int;
+
 };
 #endif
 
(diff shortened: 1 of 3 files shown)


clyne:
> This PR implements time and geo coordinate conversions when writing flow lines to a file.
> 
> @clyne The time conversion seems to be obviously correct, but I'm not 100% sure the geo coordinate conversion result is also correct and easy to read. Could you take a look and maybe provide feedback?

@shaomeng looks good except that I would not include the Z coordinate. No idea what it means, and is probably misleading.

shaomeng:
> > This PR implements time and geo coordinate conversions when writing flow lines to a file.
> > @clyne The time conversion seems to be obviously correct, but I'm not 100% sure the geo coordinate conversion result is also correct and easy to read. Could you take a look and maybe provide feedback?
> 
> @shaomeng looks good except that I would not include the Z coordinate. No idea what it means, and is probably misleading.

Gotcha. I have remove the Z conversion.

clyne (approved):

PR #2440, opened 2020-11-12
Fix #2418 Flow: Streamline data file can only be written once

### LINKED ISSUE #2418: Flow: Streamline data file can only be written once
I seem to only be able to write the flow streamline .csv once.

To reproduce on OSX:
1. Delete ~/VaporFlow.txt
2. Create a default flow renderer, and write VaporFlow.txt with the first variable in the list (P in DUKU's case)
3. Note the time of day that the file was written
4. Wait a minute and try to write the file again.  The timestamp does not update

StasJ:
Fix #2418

Fixed the bug by removing code that prevented changes to the params database from occurring from within a paint event. This would in the past have caused other bugs but I previously put checks in place that prevent paint events from queueing recursively.

Also added a mechanism for non-undoable actions which the write to file button now uses. This is also now automatically applied to all params changes from within a paint event since these are only supposed to be a side effect of an action the user performed in the GUI which will be undoable.

### DIFF
--- a/apps/vaporgui/PButton.h
+++ b/apps/vaporgui/PButton.h
@@ -19,6 +19,7 @@ class PButton : public PWidget {
     bool _disableUndo = false;
 public:
     PButton(std::string label, Callback cb);
+    // @copydoc VAPoR::ParamsMgr::SetSaveStateUndoEnabled(bool)
     PButton *DisableUndo();
 protected:
     void updateGUI() const override {}
--- a/apps/vaporgui/PButton.h
+++ b/apps/vaporgui/PButton.h
@@ -16,10 +16,13 @@ class PButton : public PWidget {
     typedef std::function<void(VAPoR::ParamsBase*)> Callback;
     VPushButton *_button;
     const Callback _cb;
+    bool _disableUndo = false;
 public:
     PButton(std::string label, Callback cb);
+    PButton *DisableUndo();
 protected:
     void updateGUI() const override {}
+    bool requireParamsMgr() const override { return _disableUndo; }
 private:
     void clicked();
 };
--- a/include/vapor/ParamsMgr.h
+++ b/include/vapor/ParamsMgr.h
@@ -509,10 +509,16 @@ public:
 
  bool GetSaveStateEnabled() const { return (_ssave.GetEnabled()); }
     
+    
+    //! Enable/Disable adding params changes to the undo list.
+    //! When enabled, behaves as normal.
+    //! When disabled, params are saved as normal, however the undo list is not updated.
+    //! An example use case is to store a computed value in the params database.
     void SetSaveStateUndoEnabled( bool enabled) {
        _ssave.SetUndoEnabled(enabled);
     }
 
+    //! Get whether updating the undo list is enabled.
     bool GetSaveStateUndoEnabled() const { return (_ssave.GetUndoEnabled()); }
 
  //! Restore state to previously saved state
--- a/include/vapor/ParamsMgr.h
+++ b/include/vapor/ParamsMgr.h
@@ -508,6 +508,12 @@ public:
  }
 
  bool GetSaveStateEnabled() const { return (_ssave.GetEnabled()); }
+    
+    void SetSaveStateUndoEnabled( bool enabled) {
+       _ssave.SetUndoEnabled(enabled);
+    }
+
+    bool GetSaveStateUndoEnabled() const { return (_ssave.GetUndoEnabled()); }
 
  //! Restore state to previously saved state
  //!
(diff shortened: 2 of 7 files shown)


apps/vaporgui/PButton.h
```diff
@@ -16,10 +16,13 @@ class PButton : public PWidget {
     typedef std::function<void(VAPoR::ParamsBase*)> Callback;
     VPushButton *_button;
     const Callback _cb;
+    bool _disableUndo = false;
 public:
     PButton(std::string label, Callback cb);
+    PButton *DisableUndo();
```
clyne, line 22: Please document.

include/vapor/ParamsMgr.h
```diff
@@ -508,6 +508,12 @@ class PARAMS_API ParamsMgr : public MyBase {
  }
 
  bool GetSaveStateEnabled() const { return (_ssave.GetEnabled()); }
+
```
clyne, line 511: Please document {Get,Set}SaveStateUndoEnabled AND {Get,Set}SaveStateEnabled so we have a clear understanding of what these closely related methods do and when you might use one versus the other.

clyne (commented):
Nice work, @stasj. I 've left a couple of in-line comments and have a couple of questions.

1. Does this PR resolve the problem we discussed yesterday where parameter changes were not being excepted by the Params data base?

2. It seems like having the ControlExec::Paint() method as the only method that the UI can use to interact with the renderers is becoming a real limitation, forcing us to do unnatural things like image capture, output of flow lines, etc. from within Renderer::_paintGL. Perhaps it's time to think about adding something like pre and post paint methods that could be triggered to perform these non-rendering tasks. What are your thoughts?

sgpearse (commented):
It's bizarre to me, but my QFileDialog still does not let me specify a filename.  We all saw Stas demonstrate this functionality yesterday.  

I've checked out issue2418 and have rebuilt from scratch, but I still don't see a way to enter a file name.  I'm testing on OSX 10.14.6.

@stasj I'll review this more tomorrow morning on some other platforms.  Could you share which OSX version you used to demonstrate the fix yesterday?

Screen Shot 2020-11-12 at 3 22 49 PM

StasJ:
> It's bizarre to me, but my QFileDialog still does not let me specify a filename. We all saw Stas demonstrate this functionality yesterday.
> 
> I've checked out issue2418 and have rebuilt from scratch, but I still don't see a way to enter a file name. I'm testing on OSX 10.14.6.
> 
> @StasJ I'll review this more tomorrow morning on some other platforms. Could you share which OSX version you used to demonstrate the fix yesterday?

@sgpearse Odd. I used Mojave.

include/vapor/ParamsMgr.h
StasJ: I documented {Get,Set}SaveStateUndoEnabled.

I did not write the {Get,Set}SaveStateEnabled functions and I don’t know what they are supposed to do. Based on the resulting behavior I am assuming their functionality is broken.

apps/vaporgui/PButton.h
StasJ: Done

StasJ:
> Nice work, @StasJ. I 've left a couple of in-line comments and have a couple of questions.
> 
> 1. Does this PR resolve the problem we discussed yesterday where parameter changes were not being excepted by the Params data base?

Yes

> 2. It seems like having the ControlExec::Paint() method as the only method that the UI can use to interact with the renderers is becoming a real limitation, forcing us to do unnatural things like image capture, output of flow lines, etc. from within Renderer::_paintGL. Perhaps it's time to think about adding something like pre and post paint methods that could be triggered to perform these non-rendering tasks. What are your thoughts?

I agree that some other mechanism for performing non-rendering tasks would be useful although it may likely require significant refactoring.

include/vapor/ParamsMgr.h
clyne: Well then let's either fix the older function or replace it, But let's not have two methods that are ostensibly supposed to provide the same functionality but for some reason are not.

StasJ: Well I can’t fix it if I don’t know what it’s supposed to do. For the same reason I also didn’t know that they are supposed to have the same functionality. I can look into whether we can simply remove/replace the old function since it is broken anyway but since it is so old there may be code that depends on its odd behavior.

StasJ: It seems to be used in multiple places and based on the context in which it is used I don't think my new function provides the intended functionality of the old function.

sgpearse:
@stasj my bad.  

I'm getting a core dump when I write the file multiple times with different variables.  But I think this problem lies outside of this bug, so I'll write a new bug and approve.

sgpearse (approved):

include/vapor/ParamsMgr.h
clyne: I'm going to approve this, but I've opened a ticket to address (refactor) some of the uglyness that is going on with state saving: #2444

clyne (approved):

PR #2442, opened 2020-11-13
Fix #2391, to prevent extra frame captures at the end of an image capture sequence

### LINKED ISSUE #2391: Capture Image Sequence: Extra frame captured
To reproduce:

1) Import the following WRF files
    /glade/p/cisl/vast/vapor/data/Source/WRF/Katrina/
        wrfout_d02_2005-08-29_00
        wrfout_d02_2005-08-29_01
2) Create an isosurface renderer, and enable it
3) Begin an image-capture-sequence and play the time steps forward
4) End the capture sequence, and there will be a duplicate of the last frame

sgpearse:
A band-aid to fix #2391, which demonstrated one or two extra frame captures at the end of a frame capture sequence.

### DIFF
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -2254,11 +2254,13 @@ bool MainForm::eventFilter(QObject *obj, QEvent *event) {
 
 		// force visualizer redraw
 		//
-        menuBar()->setEnabled(false);
-        _progressEnabled = true;
-		_vizWinMgr->Update(false);
-        _progressEnabled = false;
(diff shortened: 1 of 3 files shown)


clyne (approved):
Why do you describe this as a "band-aid"?

sgpearse:
I'd call this a band-aid because it's not clear why we don't want to call VizWinMgr::Update() while animating.  But it fixes the bug.  

The modified function is already very long and packed full of conditionals.  The MainForm is 16 years old right now and might be due for some TLC.

StasJ (approved):

PR #2492, opened 2020-11-18
Fixed  #2462 - redo broken

### LINKED ISSUE #2462: Redo broken
1. Do something
2. Undo it

Redo is greyed out.

clyne:
Fixed  #2462

### DIFF
--- a/lib/render/ControlExecutive.cpp
+++ b/lib/render/ControlExecutive.cpp
@@ -160,12 +160,12 @@ int ControlExec::Paint(string winName, bool fast){
 
 	// Disable state saving when generating the transfer function
 	//
-	bool enabled = _paramsMgr->GetSaveStateUndoEnabled();
-	_paramsMgr->SetSaveStateUndoEnabled(false);
+	bool enabled = _paramsMgr->GetSaveStateEnabled();
+	_paramsMgr->SetSaveStateEnabled(false);
 
 	int rc =  v->paintEvent(fast);
 
-	_paramsMgr->SetSaveStateUndoEnabled(enabled);
+	_paramsMgr->SetSaveStateEnabled(enabled);
 
 
 	if(rc) SetErrMsg("Error performing paint event");
--- a/lib/vdc/CurvilinearGrid.cpp
+++ b/lib/vdc/CurvilinearGrid.cpp
@@ -312,7 +312,9 @@ CurvilinearGrid::ConstCoordItrCG::ConstCoordItrCG(
 	_xCoordItr = rhs._xCoordItr;
 	_yCoordItr = rhs._yCoordItr;
 	_zCoordItr = rhs._zCoordItr;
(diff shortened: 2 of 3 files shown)


lib/vdc/CurvilinearGrid.cpp
```diff
@@ -312,7 +312,9 @@ CurvilinearGrid::ConstCoordItrCG::ConstCoordItrCG(
 	_xCoordItr = rhs._xCoordItr;
 	_yCoordItr = rhs._yCoordItr;
 	_zCoordItr = rhs._zCoordItr;
-	_terrainFollowing = rhs._terrainFollowing;
+	if (rhs._terrainFollowing) {
+		_terrainFollowing = rhs._terrainFollowing;
+	}
```
StasJ, line 317: This seems unrelated

lib/render/ControlExecutive.cpp
```diff
@@ -160,12 +160,12 @@ int ControlExec::Paint(string winName, bool fast){
...
 	//
-	bool enabled = _paramsMgr->GetSaveStateUndoEnabled();
-	_paramsMgr->SetSaveStateUndoEnabled(false);
+	bool enabled = _paramsMgr->GetSaveStateEnabled();
+	_paramsMgr->SetSaveStateEnabled(false);
 
 	int rc =  v->paintEvent(fast);
 
-	_paramsMgr->SetSaveStateUndoEnabled(enabled);
+	_paramsMgr->SetSaveStateEnabled(enabled);
```
StasJ, line 168: This will reintroduce #2418. I would assume the proper fix is to prevent changes performed when undo is disabled from clearing the redo list.

StasJ (changes requested):

lib/vdc/CurvilinearGrid.cpp
clyne: It is. I'm not sure how this snuck in here, but it does fix a memory access error.

lib/render/ControlExecutive.cpp
clyne: I'm not sure what the "proper" fix is, but if you want to modify SetSaveStateUndoEnabled per your description we can try it.

StasJ: I checked and this does not reintroduce that bug. After some head scratching I think I figured out why this is. Also I checked and preventing the redo list from clearing when undo is disabled fixes the issue as well, however I am not familiar enough with this code to say which is the proper fix.

StasJ (approved):

lib/render/ControlExecutive.cpp
clyne: Yeah, undo/redo needs some work. I've opened a ticket to do some refactoring. Hopefully the latest bandaid will get us through the release.

PR #2497, opened 2020-11-20
Fix #2470 - Prevent zero-vector 'view' and 'up' directions for camera

### LINKED ISSUE #2470: VAPOR crashes when 0 is input in
1. Open VAPOR, create a volume renderer, enable it.
2. Go to tab Navigation -> Viewpoint.
3. Select Projection Mode to be "Map Orthographic."
4. Change Z direction in "Camera Controls" field to 0.
5. Vapor crashes.

sgpearse:
This PR adds a return code in the TrackBall's ModelViewMatrix creator.  If a zero-vector is given as either the view or up direction, it returns 1 and sets MSG_ERR.  

The NavigationEventRouter then receives the return code and either applies a ModelViewMatrix from the TrackBall (rc==0), or reverts to the ModelViewMatrix of the first Visualizer (rc<0).

### DIFF
--- a/apps/vaporgui/NavigationEventRouter.cpp
+++ b/apps/vaporgui/NavigationEventRouter.cpp
@@ -1011,7 +1011,7 @@ void NavigationEventRouter::_setViewpointParams(
 	//
 	Trackball trackball;
     int rc = trackball.setFromFrame(posvec, dirvec, upvec, center, true);
-    if (rc > 0 ) {  // If trackball fails
+    if (rc < 0 ) {  // If trackball fails
 
         // Get the first available visualizer to retrieve a model view matrix we
         // can reset to.
--- a/apps/vaporgui/NavigationEventRouter.cpp
+++ b/apps/vaporgui/NavigationEventRouter.cpp
@@ -1012,7 +1012,6 @@ void NavigationEventRouter::_setViewpointParams(
 	Trackball trackball;
     bool rc = trackball.setFromFrame(posvec, dirvec, upvec, center, true);
     if (rc == false ) {  // If trackball fails
-        return;
         MSG_ERR( "Invalid camera settings" );
         
         // Get the first available visualizer to retrieve a model view matrix we
--- a/apps/vaporgui/NavigationEventRouter.cpp
+++ b/apps/vaporgui/NavigationEventRouter.cpp
@@ -1009,25 +1009,25 @@ void NavigationEventRouter::_setViewpointParams(
 	//
 	Trackball trackball;
     int rc = trackball.setFromFrame(posvec, dirvec, upvec, center, true);
-    std::cout << "rc " << rc << std::endl;
     std::vector<double> modelview, centerv;
-    if (rc > 0 ) {
+    if (rc > 0 ) {  // If trackball fails
+
+        // Get the first available visualizer to retrieve a model view matrix we
+        // can reset to.
 	    ParamsMgr *paramsMgr = _controlExec->GetParamsMgr();
 	    vector <string> winNames = paramsMgr->GetVisualizerNames();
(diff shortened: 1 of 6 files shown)


apps/vaporgui/TrackBall.cpp
```diff
@@ -368,11 +370,26 @@ bool Trackball::ReconstructCamera(
 
 // Set the quaternion and translation from a viewer frame
 // Also happens to construct modelview matrix, but we don't use its translation
-void Trackball::setFromFrame(
+int Trackball::setFromFrame(
```
clyne, line 373: Our convention for error handling is that functions return an integer status if they report errors  with MyBase. Otherwise they should return a boolean.

apps/vaporgui/TrackBall.cpp
```diff
@@ -368,11 +370,26 @@ bool Trackball::ReconstructCamera(
...
 	bool persp
 ) {
+    VAssert( posvec.size() == 3 );
+    VAssert( dirvec.size() == 3 );
+    VAssert( upvec.size() == 3 );
+    VAssert( centerRot.size() == 3 );
+
+    // Check for valid up vector and view direction
+    if ( upvec[0] == 0 && upvec[1] == 0 && upvec[2] == 0 ) {
+        MSG_ERR("Invalid up vector");
```
clyne, line 385: Leave the error reporting to the user of this class, or make use of MyBase to log the error if you think that it's important to provide a message.

apps/vaporgui/NavigationEventRouter.cpp
```diff
@@ -1004,20 +1004,38 @@ void NavigationEventRouter::_setViewpointParams(
...
-    trackball.setFromFrame(posvec, dirvec, upvec, center, true);
-    trackball.TrackballSetMatrix();
-
-	const double *m = trackball.GetModelViewMatrix();
-
-	vector <double> modelview (m, m + 16);
-	vector <double> centerv (center, center + 3);
+    int rc = trackball.setFromFrame(posvec, dirvec, upvec, center, true);
+    if (rc < 0 ) {  // If trackball fails
+
```
clyne, line 1015: Why is recovery needed here? If setFromFrame fails why not simply return and leave the params database unchanged?

apps/vaporgui/TrackBall.cpp
```diff
@@ -368,11 +370,26 @@ bool Trackball::ReconstructCamera(
...
+        return -1;
+    }
+    if ( dirvec[0] == 0 && dirvec[1] == 0 && dirvec[2] == 0 ) {
```
StasJ, line 388: Since these are floats we may want to check if they are very small rather than if they are 0. You can use FLT_EPSILON for that.

sgpearse: Correct me if I'm wrong, but the viewing and up vectors are only invalid if all 3 components are 0.  If the components are 0, 0, and FLT_EPSILON, then the vector would be valid.

apps/vaporgui/TrackBall.cpp
sgpearse: Fixed.

apps/vaporgui/TrackBall.cpp
sgpearse: Ok, error report has been bumped up into the NavigationEventRouter.

apps/vaporgui/NavigationEventRouter.cpp
sgpearse: This was to refresh the gui with valid values.  I agree that this shouldn't be modifying the params database.  It now just refreshes the gui with what's in params.

clyne (approved):

StasJ (approved):

PR #2502, opened 2020-11-21
fix #2477: crash

### LINKED ISSUE #2477: Flow crashes on 2D dataset
To reproduce:

1) Import the wrf file /glade/p/cisl/vast/vapor/data/Source/WRF/2D/wrf.nc
2) Create a Flow renderer, and use U10 and V10 for the X and Y field components
3) Enable the renderer and Vapor will crash

Test machine: Casper

shaomeng:
This PR fixes the crash problem by enforcing a color mapping variable to be specified. This color mapping variable is used by the code to decide a tube size when rendering.

An alternate solution is to use a different logic to decide rendering tube size. Input from Stas will be needed to implement that different logic.

### DIFF
--- a/lib/render/FlowRenderer.cpp
+++ b/lib/render/FlowRenderer.cpp
@@ -239,14 +239,6 @@ int FlowRenderer::_paintGL( bool fast )
         MyBase::SetErrMsg("Please provide at least 1 field variables for advection!");
         return flow::PARAMS_ERROR;
     }
-    // In case a color mapping variable wasn't specified, it's also considered an ill-formed
-    //   parameter. That is because, the color mapping variable is required in private method
-    //   _renderAdvectionHelper() and there isn't an obvious way to satisfy it without a
-    //   color mapping variable.
-    if( _colorField.ScalarName.empty() ) {
-        MyBase::SetErrMsg("Please provide a color mapping variable!");
-        return flow::PARAMS_ERROR;
-    }
 
     if( _velocityStatus == FlowStatus::SIMPLE_OUTOFDATE )
     {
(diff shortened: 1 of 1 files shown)


clyne:
@sam, @StasJ is it not possible to using the flow renderer without a color mapping variable specified? This seems like an unnecessary limitation that prevents (or makes difficult) a very simple use case: rendering the glyphs with constant color.

shaomeng:
> @sam, @StasJ is it not possible to using the flow renderer without a color mapping variable specified? This seems like an unnecessary limitation that prevents (or makes difficult) a very simple use case: rendering the glyphs with constant color.

I went ahead and changed the logic to determine tube size, so the code doesn't rely on a color mapping variable anymore. Now it is OK to have an empty color mapping variable and the program will execute as normal.

clyne (approved):

StasJ (changes requested):
> @sam, @StasJ is it not possible to using the flow renderer without a color mapping variable specified? This seems like an unnecessary limitation that prevents (or makes difficult) a very simple use case: rendering the glyphs with constant color.

@clyne There are certain variable types that are guaranteed to never be empty and the color mapping variable is one of them. There is a bug in the params database, however, where if you have a renderer that supports 2d and 3d data, and you load a 2d-only dataset, the initialization of this variable silently fails. This bug was unfortunately fixed by the old variable selection GUI which would band-aid the bug when it arose. Thus, with the removal of the old variable selection GUI, the bug has resurfaced and is causing the issue here. I have submitted a PR that fixes the original bug (#2506). As to whether this limitation is unnecessary: probably, but this is a limitation that was imposed many years ago and not by me. The code I wrote functions within this limitation since that is the framework provided by the legacy code. That being said, I think it would make sense to remove this restriction but the changes in this PR create a situation where the radius scalar is not based on the color map variable even when it is being used. This can be simply fixed by only resorting to the velocity field variables if the color map variable is not being used.

shaomeng:
Hey @StasJ , I see that you requested a change, but I have trouble to understand what exactly I should change. Can you be a little more specific about what change you want to see ?

StasJ (approved):

shaomeng:
Hi @StasJ  this branch is updated and ready to be reviewed again.

PR #2513, opened 2020-12-01
Fix #2505 PWidget PIntegerSliderEdit inconsistant with the value in Params

### LINKED ISSUE #2505: PWidget PIntegerSliderEdit inconsistant with the value in Params.
### Describe the bug
`PIntegerSliderEdit` sometimes does not reflect the value stored in Params.

### To Reproduce
It will require inserting a print statement in the source code, but here are the step what I discovered this bug.
1. Locate any line that has `GetPastNumOfTimeSteps()` in `FlowRenderer.cpp`.
2. Immediately insert a print statement to print out what `GetPastNumOfTimeSteps()` returns. 
3. Compile VAPOR.
4. Open VAPOR, load a series of data set with multiple time steps.
5. Create a Flow renderer, change the rendering type to "Pathlines`.
6. Enable the renderer. Notice the number being print out in the terminal. It should equal to the number of time steps of the data set being imported.
7. Notice the value shown in the `Pathline Length` widget. It shows 0.
8. However this widget should reflect values stored in the params database, and in accordance with the print out value.

StasJ:
Fix #2505

### DIFF
--- a/apps/vaporgui/FlowEventRouter.cpp
+++ b/apps/vaporgui/FlowEventRouter.cpp
@@ -122,8 +122,8 @@ void FlowEventRouter::_updateTab(){
     RenderEventRouterGUI::_updateTab();
     
     int numTS = GetActiveDataMgr()->GetNumTimeSteps();
-    _pathlineLengthSlider   ->SetRange(0, numTS-1);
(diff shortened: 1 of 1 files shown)


shaomeng (approved):
Good findings, thanks Stas!

Issue #2484, opened 2020-11-18
No flow pathline with MPAS-O data set


shaomeng:
1. Open VAPOR, import MPAS-O data: `/little_glade/data/Source/MPAS/MPAS-O`
1. Create a flow renderer, and enable it. Nothing gets rendered. Adjusting variable names doesn't seem to help.

shaomeng:
This particular data set and variable upon opening the software has interesting field values that ain't technically missing values, but probably are. More specifically, there are field values as big as `5.14008662e+31` but the official missing values are `inf`. As a result, calculations based on values that big blew up.

PR #2515, opened 2020-12-01
Mitigate Issue 2484: Pop up message. 

shaomeng:
This PR doesn't really fix issue #2484, but handles it more user-friendly. 

Note: this PR should only be merged after #2507  is merged.

### DIFF
--- a/lib/render/FlowRenderer.cpp
+++ b/lib/render/FlowRenderer.cpp
@@ -239,14 +239,6 @@ int FlowRenderer::_paintGL( bool fast )
         MyBase::SetErrMsg("Please provide at least 1 field variables for advection!");
         return flow::PARAMS_ERROR;
     }
-    // In case a color mapping variable wasn't specified, it's also considered an ill-formed
-    //   parameter. That is because, the color mapping variable is required in private method
-    //   _renderAdvectionHelper() and there isn't an obvious way to satisfy it without a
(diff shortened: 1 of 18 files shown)


lib/render/FlowRenderer.cpp
```diff
@@ -245,7 +245,7 @@ int FlowRenderer::_paintGL( bool fast )
         // First step is to re-calculate deltaT
         rv = _velocityField.CalcDeltaTFromCurrentTimeStep( _cache_deltaT );
         if( rv == flow::FIELD_ALL_ZERO ) {
-            MyBase::SetErrMsg("The velocity field seems to contain only zero values!");
+            MyBase::SetErrMsg("The velocity field seems to contain only zero or missing values!");
```
clyne, line 248: One small nit: rather than saying the flow field contains "missing values", which may or may not be true, how about saying something more generic and correct such as "unusable values",  or 'unsuitable values", or simply 'invalid values'?

shaomeng, line 248:
```suggestion
            MyBase::SetErrMsg("The velocity field seems to contain only invalid values!");
```

shaomeng: Hey @clyne , I have changed to simply say "invalid values." I think it's good to go now!

PR #2549, opened 2020-12-08
Fixes #2536 - Vapor hangs with some WRF-Fire data sets

### LINKED ISSUE #2536: Vapor hangs with some WRF-Fire data sets
Hi everybody,

I am trying to visualize some WRF-Fire idealized simulation output using Vapor 3.2. However, when I want to create barb renderer or some other renderers Vapor hangs and I have to force quit it! The WRF output file is relatively small, about 160 mb. and also only Wireframe renderer works correctly.

I am using WRF version 4.2 and Vapor version 3.1 on my MacBook Pro 16 running Catalina.

I appreciate every help, and thank you very much in advance!

- [*] Did VAPOR crash?

**Impact**
- [*] High - User productivity significantly degraded
- [ ] Medium - User productivity partially degraded

### Desktop
 - OS and version: macOS Catalina 10.15.5
 - Version: Vapor 3.2

clyne:
Fixes #2536 

Note: the test data are now on glade in WRF/WRF-Fire/kasrash

### DIFF
--- a/lib/vdc/DCWRF.cpp
+++ b/lib/vdc/DCWRF.cpp
@@ -812,10 +812,11 @@ bool DCWRF::_isIdealized(NetCDFCollection *ncdfc) const {
 	if (Wasp::StrCmpNoCase(s, "IDEALIZED DATA") == 0) return(true);
 
 	// Pre version 4.x WRF did not have an attribute to identify 
-	// idealized cases. However, these cases have constant valued 
-	// horizontal coordinates.
+	// idealized cases. However, these cases have constant-valued 
+	// XLONG and XLAT coordinates. N.B. The staggered grid horizontal
+	// coordinates (XLONG_U, XLAT_V, etc.) may not exist.
 	//
-	vector <string> cvars = {"XLONG", "XLAT", "XLONG_U", "XLAT_U", "XLONG_V", "XLAT_V"};
+	vector <string> cvars = {"XLONG", "XLAT"};
 	for (auto &c : cvars) {
 		if (_isConstantValuedVariable(ncdfc, c)) return(true);
 	}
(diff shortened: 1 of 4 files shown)


lib/vdc/DCWRF.cpp
```diff
@@ -748,42 +748,38 @@ int DCWRF::_GetProj4String(
...
+	ncdfc->Close(fd);
 
-		for (size_t i=0; i<Wasp::VProduct(dims)-1; i++) {
-			if (data[i] != data[i+1]) {
-				EnableErrMsg(enabled);
-				return(false);
-			}
+	for (size_t i=0; i<Wasp::VProduct(dims)-1; i++) {
+		if (data[i] != data[i+1]) {
+			EnableErrMsg(enabled);
```
StasJ, line 781: Do we want to account for "almost constant" since these are floats? Also, performance can be improved by comparing all values against the first.

StasJ (approved):

lib/vdc/DCWRF.cpp
clyne: Hmm. I think doing "almost constant" correctly would be expensive and might be overkill. You'd have to compare every value with every other value, wouldn't you? Consider

`a ~= b && b ~= c`

does not imply

`a ~= c`

StasJ: I don't think you would have to do that. a_0 ~= x in [a_1..a_n] implies the field is constant.

clyne: You are correct. Done.

StasJ (approved):

sgpearse (approved):

Issue #2543, opened 2020-12-07
Flow: List of seeds not working?


sgpearse:
Produced on Casper.

The session /glade/u/home/pearse/brokenListOfSeeds.vs3 uses a "List of seeds" file in its flow streamline renderer.  

The file being used is /glade/u/home/pearse/listOfSeeds.txt, and defines one seed at 0,0,3,0 (x,y,z,time).  This streamline should only be drawn at the first timestep, but it persists for all 9.

shaomeng:
@sgpearse The current behavior of the software is that when in "Streamline" mode, the seed/particle information is ignored, and the software uses velocities from whatever the current time step is to perform "streamline" integration. That is because, almost by definition, streamlines imply that time is constant. 

I guess my question for you is:
1) is the behavior described above understandable and reasonable?
2) if it is reasonable, does your test show behaviors otherwise?

sgpearse:
Hmm.  I understand what you mean, and I'm not certain what the right thing to do is.  

On one hand, having a consistent format for the list-of-seeds file would be less complex for the user.  Using different formats means there's more code for us to maintain, and more things for the user to remember when writing their file.

On the other hand, do users ever want to change streamline seed locations across time?  Perhaps...

One thought is that moving domains have changing coordinates.  If a user is defining streamlines in Katrina, will the X,Y,Z,T values be valid after the moving domain has shifted beyond the initial time step?

clyne:
Agreed that we would simply remove time coordinate from the list of seeds file and allow the GUI to control the time.

PR #2558, opened 2020-12-09
Fix #2537 by documenting the usage of the flow list-of-seeds file

### LINKED ISSUE #2537: Flow: file format for "list of seeds" not documented
There is no context help in the GUI describing the format for the file used when Seed Distribution Type is set to "list of seeds".

sgpearse:
Explained the usage of the list-of-seeds file by adding a demo file and tooltip.

Fixed #2537

### DIFF
--- a/apps/vaporgui/FlowEventRouter.cpp
+++ b/apps/vaporgui/FlowEventRouter.cpp
@@ -56,10 +56,11 @@ FlowEventRouter::FlowEventRouter(QWidget *parent, ControlExec *ce)
             }),
             (new PShowIf(FP::_seedGenModeTag))->Equals((int)FlowSeedMode::LIST)->Then({
                 (new PFileOpenSelector(FP::_seedInputFilenameTag, "List of seeds file"))->SetTooltip(
-                    "Seed injection points may be defined in space and time in this file. \n"
-                    "The format is X,Y,Z,Time.  The X,Y,Z spacial coordinates are in meters, \n"
-                    "and Time is expressed as the timestep increment in the dataset.  This \n"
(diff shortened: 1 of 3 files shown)


clyne (commented):
@shaomeng how is the time step value treated in the seed list file when generating *stream* lines? Is it simply ignored? The user may want to specify a spatial seed locations that apply to *all* time steps, particularly when generating stream lines. One option, if it doesn't already exist, is to have a special value (e.g. -1) indicate that seeds should be injected at all time steps, and this would apply to both stream lines and path lines.

shaomeng:
Let me investigate #2543 first so this tool tip can be as accurate as possible.

shaomeng:
> @shaomeng how is the time step value treated in the seed list file when generating _stream_ lines? Is it simply ignored? The user may want to specify a spatial seed locations that apply to _all_ time steps, particularly when generating stream lines. One option, if it doesn't already exist, is to have a special value (e.g. -1) indicate that seeds should be injected at all time steps, and this would apply to both stream lines and path lines.

Given the complexity of the current GUI and available controls, I don't think it's a good idea to invent a seed list syntax that is capable to express those GUI controls. I propose the following:

1) Seed list is used to only specify location information, but not temporal information.
2) In both steady and unsteady flow cases, the temporal information of imported seeds is fixed to whatever the current time step is. 
3) Use cases such as repeated seed injection is only controlled via GUI.

Note: this proposed behavior is consistent with how seed time is determined in other seed generation methods.

sgpearse:
See issue #2543 for desired behavior

clyne (approved):

shaomeng (approved):

PR #2560, opened 2020-12-09
New time policy for reading seeds from a file

### LINKED ISSUE #2543: Flow: List of seeds not working?

shaomeng:
This PR implements the policy that reading seeds from a file only specifies seed locations. Seed times are specified using the same logic as the rest of seeding modes. 

Also fixes #2543


lib/render/FlowRenderer.cpp
```diff
@@ -1117,25 +1095,54 @@ FlowRenderer::_genSeedsRakeUniform( std::vector<flow::Particle>& seeds ) const
...
+    FlowParams* params = dynamic_cast<FlowParams*>( GetActiveParams() );
+    VAssert( params );
+
+    // Read seed locations (X, Y, Z) from a file.
+    auto read_from_disk = flow::InputSeedsCSV( params->GetSeedInputFilename() );
+    if( read_from_disk.empty() )
+        return flow::NO_SEED_PARTICLE_YET;
+
+    // Set seed time to be the time stamp at step 0
+    auto timeVal = _timestamps.at(0);
```
clyne, line 1124: I think the use of 'auto' here and above really obfuscates the code. The reader has no idea what is being returned from InputSeedsCSV or _timestamps. Please see discussions on when and when not to use auto such as:

https://stackoverflow.com/questions/6900459/the-new-keyword-auto-when-should-it-be-used-to-declare-a-variable-type

https://www.acodersjourney.com/c-11-auto/

@StasJ, @sgpearse what do you think?

shaomeng: For the sake of argument, 
1) Herb Sutter advocates an [AAA Style (Almost Always Auto)](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/), and Herb Sutter is good.
2) The two instances in question above don't go nearly as far as what Herb advocates. 

In practice, I can change it to a non-auto style if reviewers desire ;)

StasJ: I would advocate for reserving the auto keyword for scenarios where either the type can be determined from the surrounding context without having to reference other parts of the code or the type is not important to understanding the code.

sgpearse: I sometimes find myself digging through more code than I want to, in order to figure out what auto'd datatype I'm tracking.  It's hard to argue with Herb though.  

I think that auto is easier to understand if our functions are short, and follow the single responsibility principle.  My issue isn't with auto itself, but auto'd variables that live for many lines of code.

sgpearse (approved):

lib/render/FlowRenderer.cpp
clyne: @shaomeng let's  avoid auto when the type is unclear from the immediate context. Thanks.

shaomeng: This request is addressed in 48dcaf62.

StasJ (approved):

clyne (approved):

PR #2561, opened 2020-12-09
Auto determine scalar based on version

StasJ:
This PR ensures an additional sampling rate is applied to ospray volumes only if the Ospray version is < 2.5. It also slightly improves memory usage while loading data but does not resolve the overall memory usage requirements of ospray since most of the memory is allocated later but it was something small I noticed could be improved.

### DIFF
--- a/include/vapor/OSPRay.h
+++ b/include/vapor/OSPRay.h
@@ -10,6 +10,7 @@ namespace VOSP {
 RENDER_API int Initialize(int *argc, char **argv);
 RENDER_API int Shutdown();
 RENDER_API std::string Version();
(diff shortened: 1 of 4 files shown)


clyne:
@StasJ as there is no issue associated with this PR can please summarize what it does.

StasJ:
This PR ensures an additional sampling rate is applied to ospray volumes only if the Ospray version is < 2.5. It also slightly improves memory usage while loading data but does not resolve the overall memory usage requirements of ospray since most of the memory is allocated later but it was something small I noticed could be improved.

clyne (approved):

Issue #2080, opened 2020-01-10
Flawed text box


shaomeng:
In the following range selector, anyone who's thinking the range is (0.019 -- 0.413) is wrong. It's from negative 0.019 instead. 

As we move towards VWidgets or PWidgets which will be reused many times, it's worth putting some effort into these widgets so they don't show such misleading displays. 
![flow](https://user-images.githubusercontent.com/814885/72192329-879b5f00-33c2-11ea-856e-81ea0e8e9322.png)

StasJ:
Can you provide instructions on how to reproduce?

shaomeng:
> Can you provide instructions on how to reproduce?

Well, I cannot find the specific data set that causes this problem, sorry about that. 

The idea is that when the text box hides leading digits, then it also hides the negative sign, which would cause big misunderstanding. It shouldn't hide tailing digits either, since there could be a `+e4` that got hidden.

If this information is still not enough to debug the text box, then simply reject it for now, and I'll file a more proper bug later when I run into it.

StasJ:
I see what you mean but I can't think of any application that does has such a feature so users expect text boxes to have potentially hidden information. Also, that is not easy to implement and literally every single text box in vapor has this issue. We can add this as a feature request for Vapor Widgets.

shaomeng:
Glad you see what the problem is. I personally consider it unacceptable for
a scientific application to display misleading numbers, but if that's a
difficult fix, we can postpone it to a later time.

clyne:
@stasj please explore manually truncating the precision

StasJ:
> @StasJ please explore allowing user to manually truncating the precision

Unfortunately, this is not feasible with the current VWidgets. VDoubleLineEdit is missing the required functionality. The needed features, such as the ability to set the valid input range and precision, are implemented in the VDoubleSliderEdit rather than the VDoubleLineEdit.

shaomeng:
The described behavior still exists. I.e., when a text box isn't able to display all digits, it hides the most significant (left most) instead of the most insignificant (right most) digits.

StasJ:
> The described behavior still exists. I.e., when a text box isn't able to display all digits, it hides the most significant (left most) instead of the most insignificant (right most) digits.
> 
> This "CursorPosition" property might help us achieve the same goal by placing the cursor at the left most position:
> https://doc.qt.io/qt-5/qlineedit.html#cursorPosition-prop

I tried using the CursorPosition property but it did not work.

[...]

PR #2564, opened 2020-12-10
Additional fix for #2080 - Set cursor position to better position in VStringLineEdit

sgpearse:
This change resets the VStringLineEdit's cursor to position 0 after a change is applied.  This results in more significant digits being displayed, before following digits of lesser magnitude.

### DIFF
--- a/apps/vaporgui/VDoubleLineEdit.cpp
+++ b/apps/vaporgui/VDoubleLineEdit.cpp
@@ -11,7 +11,7 @@ VDoubleLineEdit::VDoubleLineEdit( double value ) :
     _value( value )
 {
     std::string formattedNumber = _formatValue( _value );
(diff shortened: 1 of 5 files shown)


shaomeng (approved):
This PR fixes the described issue on Ubuntu Linux. Could be a good idea to test on more platforms.

StasJ (changes requested):
`VStringLineEdit::_setValueString` and `VStringLineEdit::SetValueString` are duplicates of each other and with this change will exhibit different behavior depending on which is called.

sgpearse:
You are correct.  

The duplicate private function has been removed.  This also implies un-deleting the SetValueString() function in VNumericLineEdit.

StasJ (approved):

PR #2582, opened 2020-12-16
fix #2581

### LINKED ISSUE #2581: Pathline length: maximum should be num_of_TS - 1.
Given the definition of displayed pathline length, which is how many time steps to display before the current time step, that number should max out at `num_of_TS-1`. The current implementation has that number at `num_of_TS`.

To use an example, when there are 2 time steps, the longest possible pathline is 1.

shaomeng:
This PR fixes a minor bug:  #2581  ( fix #2581 )

(This PR is re-introducing a previous bug (#2505) and a better resolution is being investigated by @StasJ and Sam.)

### DIFF
--- a/apps/vaporgui/FlowEventRouter.cpp
+++ b/apps/vaporgui/FlowEventRouter.cpp
@@ -132,7 +132,7 @@ void FlowEventRouter::_updateTab(){
     RenderEventRouterGUI::_updateTab();
     
     int numTS = GetActiveDataMgr()->GetNumTimeSteps();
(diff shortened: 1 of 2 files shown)


clyne:
@shaomeng please add "fixes @2582" or "closes #2581", etc. so that the PR and issue get linked. Thanks!

shaomeng:
@StasJ I believe I fixed the regression and also achieved the desired behavior. Please feel free to provide any review.

StasJ (approved):

PR #2589, opened 2021-01-08
Improve error handling during Python initialization

RemiLacroix-IDRIS:
Hi,

I compiled VAPOR for a user of our computing center and run into some issues.

I noticed that there are some problems regarding the way errors are currently handled during Python initialization.

I tried to fix some of the issues and I think there is still some room for improvements but I don't know the code base so I am not completely sure.

I think the main remaining issue is that although `SetErrMsg` is called, I don't think the error is actually ever displayed.

Best regards,
Rémi

### DIFF
--- a/lib/render/MyPython.cpp
+++ b/lib/render/MyPython.cpp
@@ -127,9 +127,9 @@ int MyPython::Initialize() {
 	//
 	std::string stdErr =
 	"try:\n"
-	"	import sys, os\n"
+	"	import sys\n"
 	"except: \n"
-	"	print >> sys.stderr, \'Failed to import sys\'\n"
+	"	print('Failed to import sys')\n"
 	"	raise\n"
 	"class CatchErr:\n"
 	"	def __init__(self):\n"
@@ -149,9 +149,9 @@ int MyPython::Initialize() {
 
 	std::string stdOut =
 	"try:\n"
-	"	import sys, os\n"
+	"	import sys\n"
 	"except: \n"
-	"	print >> sys.stderr, \'Failed to import sys\'\n"
+	"	print('Failed to import sys')\n"
 	"	raise\n"
 	"class CatchOut:\n"
 	"	def __init__(self):\n"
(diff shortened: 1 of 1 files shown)


clyne:
Thanks for your contribution, @RemiLacroix-IDRIS . Do you have a test case that demonstrates the problem this PR fixes?

@sgpearse can you see why this PR isn't passing CI testing?

RemiLacroix-IDRIS:
> Thanks for your contribution, @RemiLacroix-IDRIS . Do you have a test case that demonstrates the problem this PR fixes?

Well yes and no, it is supposed to make a difference when say the `matplotlib` package is missing. However VAPOR does not seem to display the error message it collects so you won't actually see what the fix does...

To debug my issue I had to replace the call to `SetErrMsg` by a `printf`:
https://github.com/NCAR/VAPOR/blob/33c787b6089ad845f6f989176d839e117fcdb03f/lib/render/MyPython.cpp#L181-L183

That's how I noticed that there was a problem with the error handling still being Python 2 instead of Python 3.

With the `printf`, you would see something like that before my changes:
```
PyRun_SimpleString() : VAPOR_PY: Traceback (most recent call last):
  File "<string>", line 2, in <module>
ModuleNotFoundError: No module named 'matplotlib'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 4, in <module>
TypeError: unsupported operand type(s) for >>: 'builtin_function_or_method' and 'CatchErr'. Did you mean "print(<message>, file=<output_stream>)"?
```
and now:
```
PyRun_SimpleString() : VAPOR_PY: Failed to import matplotlib
Traceback (most recent call last):
  File "<string>", line 2, in <module>
ModuleNotFoundError: No module named 'matplotlib'
```

RemiLacroix-IDRIS:
> @sgpearse can you see why this PR isn't passing CI testing?

I can only guess but I would think it's because I have opened this PR from a forked repository.

clyne (approved):

StasJ (approved):

RemiLacroix-IDRIS:
Hello,

Just wondering why this PR was closed without being merge. Is there a problem with it?

clyne:
> Hello,
> 
> Just wondering why this PR was closed without being merge. Is there a problem with it?

Whoops. Operator error. It's been merged. Thanks again!

RemiLacroix-IDRIS:
No worries and you're welcome.

Should I open an issue to discuss the fact that the error messages related to Python initialization are never displayed although there are stored with `SetErrMsg`?

clyne:
That would be great. Thanks.

PR #2598, opened 2021-01-21
Disable menu items for hidden transfer function maps #2517

### LINKED ISSUE #2517: DVR color map variable controls
I found myself confused by the DVR appearance tab when using the ColorMapVariable option.  There are two small changes that I think would improve its usability and prevent what was reported in #2151, where it appeared that selecting built-in colormaps no longer worked.

These small changes are:
1) Disable colormap options on opacity-only transfer functions
2) Move the colormap transfer function below the primary transfer function, instead of below the ray-tracing options.

Untitled drawing (1)

StasJ:
Fix #2517


clyne:
@stasj is this now ready for review?

StasJ:
> @StasJ is this now ready for review?

Yes

sgpearse (approved):
Thanks, stas.

clyne (approved):

Issue #2473, opened 2020-11-17
Annotation Tab uses different coordinate systems for different options


sgpearse:
In the Annotation tab, we are using user-coordinates for specifying axis annotations, and a unit square (0:1 on X Y and Z) for placing the RGB arrows.  We should probably use either one or the other.

clyne:
This may be a duplicate.

clyne:
This is not fixed. However, I'm not sure I agree that we should change the current behavior. The RGB arrows are in 2D image coordinates, and it makes sense to specify them in NDC coordinates (0.0 - 1.0). The tick marks are 3D and most of our 3D coordinates (e.g. Geometry) are specified in user coordinates. I'd be inclined to mark this as rejected rather than fixed.

PR #2599, opened 2021-01-22
Issue2473 - Improved drawing coordinates for axis arrows

sgpearse:
This PR addresses #2473 by doing the following:

1) Removing the old GUI elements for the axis arrow annotation, and replacing them with PWidgets

2) Instead of drawing the axis arrows inside of the domain bounding box, this PR draws the arrows as an overlay on top of the scene.  I don't think we want to be drawing the arrows inside of our renderings.  This kind of overlay approach is what's used by Visit, Paraview, Blender, and many others. Users can select an XY screen coordinate between 0 and 1 to draw the arrows to.  Arrow size can also be modified on a scale of 0 to 1.

Screen Shot 2021-01-22 at 10 09 51 AM


apps/vaporgui/AnnotationEventRouter.h
```diff
@@ -151,6 +149,8 @@ protected slots:
 
     AnimationParams *_ap;
     bool             _animConnected;
+
+    PGroup *_aa;
```
StasJ, line 153: Maybe a more descriptive name than `_aa`?

apps/vaporgui/AnnotationEventRouter.cpp
```diff
@@ -81,6 +82,12 @@ AnnotationEventRouter::AnnotationEventRouter(QWidget *parent, ControlExec *ce) :
 
     _animConnected = false;
     _ap = NULL;
+
+    _aa = new PGroup({new PSection("Orientation Arrows", {(new PCheckbox(AnnotationParams::AxisArrowEnabledTag, "Show arrows (XYZ->RGB)")),
```
StasJ, line 86: Clang-Format struggles with GUI code. You can manually format it as follows
```
// clang-format off
GUI code here
// clang-format on
```
Look at the event routers (you need to merge with master first) for an example. I don't think we've decided we're all expected to do this so its up to you or we can discuss in our meeting, but I think it makes it more readable.

StasJ (commented):
I like it!

Please merge with master before making a PR. It makes reviewing easier. Thanks.

apps/vaporgui/AnnotationEventRouter.cpp
sgpearse: Sounds good. Will do and note when it's done.

apps/vaporgui/AnnotationEventRouter.h
sgpearse: Sounds good. Will do and note when it's done.

sgpearse:
Interesting - I thought that this branch was from a fresh clone after the clang-format changes.  I'll let you know when re-review is ready.

shaomeng (commented):
Looking good Scott. Just one question, is there a particular reason to not align the 1st slider with the 2nd and 3rd? 

image

sgpearse:
@Sam  It currently grows to full the available space in the line, which is determined by the length of the label (ie Size, X Position, etc).  

I think this could be improved by limiting it to a maximum size, based on the size of its container, VDoubleSliderEdit.  We'd probably want this to happen during QWidget::repaint() function, to update the value when the user expands the control panel.  Sliders with very long labels will still need to shrink though, so I'm not sure if there's a way to make this 100% consistent.

StasJ (approved):

shaomeng:
> @sam It currently grows to full the available space in the line, which is determined by the length of the label (ie Size, X Position, etc).
> 
> I think this could be improved by limiting it to a maximum size, based on the size of its container, VDoubleSliderEdit. We'd probably want this to happen during QWidget::repaint() function, to update the value when the user expands the control panel. Sliders with very long labels will still need to shrink though, so I'm not sure if there's a way to make this 100% consistent.

I understand why it's happening, but it just looks to me 
1) not consistent locally between two lines next to each other, and 
2) not consistent with other three-widget-in-a-line composite widgets. What I mean here is, in contrary, the following screenshot shows such a widget that **_is_** able to pad white space in between the 1st and 2nd widget in that line.

Given that the the biggest selling point for the three-widget-in-a-line design is to provide consistency, I think the lack of consistency needs to be addressed. 

image

sgpearse:
> Given that the the biggest selling point for the three-widget-in-a-line design is to provide consistency, I think the lack of consistency needs to be addressed.

I agree, however I think its scope is outside of this PR.

shaomeng:
> > Given that the the biggest selling point for the three-widget-in-a-line design is to provide consistency, I think the lack of consistency needs to be addressed.
> 
> I agree, however I think its scope is outside of this PR.

Makes sense. Let's give it a ticket of itself  #2603

shaomeng (approved):

PR #2601, opened 2021-01-26
Fix #2388 Variables without data at the first timestep cause misleading behavior

### LINKED ISSUE #1705: Variable default range is based on its first timestep, even if it is uninitialized
1. Open Scott's WRF tutorial dataset (will be copied to Bugs directory when I have a better connection)
1. Set timestep to 5
1. Add a volume renderer
1. Select variable `QCLOUD`
1. Enable the renderer

The variable is uninitialized at the first timestep. The default data range will be 0 -> 0. This caused a decent amount of confusion at the WRF tutorial.

### LINKED ISSUE #2388: TFEditor: Variables without data at the first timestep cause misleading behavior
The TFEditor determines its minimum/maximum value extents based on the first timestep in a loaded dataset.  Sometimes a variable may have no data at the first time step.  When we advance the timestep to a point where a variable does have data, the TFEditor appears to show that there is no data.  This is because it's still using the bounds from the first time step, with no data.

This one is complicated.  One solution may be to normalize the TFEditor's data bounds to the range of the current variable.

To reproduce:
1) Load vapor, and import the two WRF files:
/glade/p/cisl/vast/vapor/data/Source/WRF/Katrina/wrfout_d02_2005-08-29_00
/glade/p/cisl/vast/vapor/data/Source/WRF/Katrina/wrfout_d02_2005-08-29_01
2) Create an isosurface renderer of QCLOUD
3) Note that the TFEditor range is 0 to 0
4) Advance to the next timestep.  The TFEditor range is still 0 to 0, and there appears to be no data at this timestep.
5) Drag the TFEditor data range slider to the right, and the histogram will now be correct.

StasJ:
Makes behavior of range selector clearer when uninitialized variable. When the range is 0-0 and both min/max are 0, the range selector shows that the whole range is selected. When you select the next timestep, now it shows that just 0 is selected but the tilmestep has more data. It also now has a tooltip which shows the entire possible range.

**Timestep 0**

**Timestep 1**

Fix #2388
Fix #1705

### DIFF
--- a/apps/vaporgui/AnimationEventRouter.cpp
+++ b/apps/vaporgui/AnimationEventRouter.cpp
@@ -32,6 +32,7 @@
 #include "AnimationParams.h"
 #include "RangeCombos.h"
 #include "AnimationEventRouter.h"
+#include <float.h>
 
 #include "images/playreverseA.xpm"
 #include "images/playforwardA.xpm"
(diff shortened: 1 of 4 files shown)


clyne (commented):
@stasj this change definitely improves the situation, but does it go far enough? Is there any other way that you can think of that we can bring to the user's attention that there has been a significant change in the range of values between time steps?

StasJ:
> @StasJ this change definitely improves the situation, but does it go far enough? Is there any other way that you can think of that we can bring to the user's attention that there has been a significant change in the range of values between time steps?

Is there anything specific you had in mind? Vapor does not have any mechanism that would be used for alerting the user of a change in the range of values between time steps. We could piggy back off of something else like checking for such things in the animation event router since it handles timestep changes. The PWidget GUI framework was specifically designed not to allow any temporal state (and will fail to compile if it does) so detecting a "change" from the GUI would not be possible without hacks. Also, when you say "significant change in the range" does that mean any time the range changes significantly or are we will just considering the original scope where it goes from an uninitialized variable to an initialized one? If it is the former, what would be the threshold? I think one way we could address this without looking for a change is to simply show a warning (e.g. making the histogram red) in the TF Editor when the range is 0 but the possible range is non-zero.

clyne:
Would it be reasonable to simply reinitialize the TFE if its range is zero and the current data have non-zero range? This would not provide a general solution to capture all anomalous cases, but it would address the constant-valued data cases that we've run into a few times.

sgpearse:
@stasj I made a PR - #2605 - for your PR, which will fix CircleCI's test configuration.  However it turned up a compilation error on Ubuntu, which isn't finding the fabsf() function.  CentOS passes though.

/root/project/apps/vaporgui/QRangeSliderTextCombo.cpp:76:9: error: 'fabsf' was not declared in this scope
     if (fabsf(_max - _min) < FLT_EPSILON)
         ^~~~~

Feel free to reject that PR and just merge with main.  Both will do the trick.

StasJ:
> Would it be reasonable to simply reinitialize the TFE if its range is zero and the current data have non-zero range? This would not provide a general solution to capture all anomalous cases, but it would address the constant-valued data cases that we've run into a few times.

Sounds good. I added this feature.

clyne:
Ready for rereview?

StasJ:
> @StasJ I made a PR - #2605 - for your PR, which will fix CircleCI's test configuration. However it turned up a compilation error on Ubuntu, which isn't finding the fabsf() function. CentOS passes though.
> 
> /root/project/apps/vaporgui/QRangeSliderTextCombo.cpp:76:9: error: 'fabsf' was not declared in this scope
> if (fabsf(_max - _min) < FLT_EPSILON)
> ^~~~~
> 
> Feel free to reject that PR and just merge with main. Both will do the trick.

Thanks @sgpearse, I rebased off of main and changed the function call causing the issue.

StasJ:
> Ready for rereview?

Yes

clyne (approved):
Looks great!

sgpearse (approved):

PR #2602, opened 2021-01-26
Fix #2530 Setting constant colors for glyphs is non-intuitive

### LINKED ISSUE #2530: Setting constant colors for glyphs is non-intuitive
The current mechanism for making a glyph (e.g. barb, contour line, flow tube, etc.) is non-intuitive. There is no way to set the color variable to 'none', and the only way to achieve constant color is to delete one of the control points from the TFE, which many users may not realize is possible. Let's go back to something like a check box that enables/disables constant color vs pseudo coloring.

StasJ:
Fix #2530

Adds an explicit "Use Constant Color" checkbox and color selector to all the mentioned renderers.

### DIFF
--- a/apps/vaporgui/BarbEventRouter.cpp
+++ b/apps/vaporgui/BarbEventRouter.cpp
@@ -25,6 +25,7 @@ BarbEventRouter::BarbEventRouter(QWidget *parent, ControlExec *ce) : RenderEvent
     }));
     
     AddSubtab("Appearance", new PGroup({
+        new PColormapTFEditor,
         new PSection("Barbs", {
             (new PIntegerSliderEdit(BP::_xBarbsCountTag, "X Barbs"))->SetRange(1, 50),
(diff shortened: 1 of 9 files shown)


sgpearse (approved):

clyne (commented):
In the case of the barb rendering, turning on constant color causes the TFE to be hidden. However, for the wireframe, contour, and flow renderer the TFE is not hidden. Is there a reason to not be consistent?

StasJ:
> In the case of the barb rendering, turning on constant color causes the TFE to be hidden. However, for the wireframe, contour, and flow renderer the TFE is not hidden. Is there a reason to not be consistent?

That was something I added when refactoring the GUI because in the barb renderer, when constant color is enabled, the color mapping var is not used at all which is not the case in the other renderers. I changed it to make it consistent, however.

clyne (approved):

clyne:
@stasj needs to resubmit to resolve issue with 26 files change from transition from master to main.

PR #2613, opened 2021-01-30
#2077 Change from slider to spinbox

### LINKED ISSUE #2077: Flow integration steps slider can easily lead to hang
The slider can easily allow the user to accidentally specify a very high value and then vapor will hang as it tries to compute the flow lines

StasJ:
Fix #2077

@sgpearse There are still a few things that need to be addressed with the VSpinBox:
- The changed signal is sometimes emitted multiple times for a single action with the same value.
- As you type, it will emit a value changed signal for every key press.
- Right-clicking on the widget causes the current value to be emitted as a changed value.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,9 +62,7 @@ jobs:
       - run:
           name: make installer
           command: |
-            cd /VAPOR/build
-            git pull
-            git checkout $CIRCLE_BRANCH
+            cd build
             cmake -DCMAKE_BUILD_TYPE:String=Release -DDIST_INSTALLER:string=ON ..
             make -j7
             make installer
@@ -125,9 +123,6 @@ jobs:
       - run:
           name: cmake3 and make
           command: |
-            ls -lrth /usr/local/VAPOR-Deps
-            cd /VAPOR
-            git pull
             cd build
             cmake3 -DCMAKE_BUILD_TYPE:String=Release -DDIST_INSTALLER:string=ON ..
             make
@@ -135,7 +130,6 @@ jobs:
             for f in VAPOR3-*.sh ; do mv "$f" "${f/Linux/CentOS7-Weekly}" ; done
             mkdir -p /tmp/workspace/installers
             mv *.sh /tmp/workspace/installers
-
 
       - persist_to_workspace:
           root: *workspace_root
@@ -395,7 +389,7 @@ workflows:
   nightly:
     triggers:
       - schedule:
-           cron: "0 9 * * 5" #weekly cron job on Friday @ 2:00am MST
+           cron: "30 13 * * 5" #weekly cron job on Friday @ 6:30am MST
            filters:
              branches:
                only:
@@ -404,10 +398,10 @@ workflows:
       - build_ubuntu18_installer
       - build_centos7_installer
       - build_osx_installer
-      - build_win10_installer
+#      - build_win10_installer
       - release_on_github:
           requires:
             - build_ubuntu18_installer
             - build_centos7_installer
             - build_osx_installer
-            - build_win10_installer
+#            - build_win10_installer
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -399,7 +399,7 @@ workflows:
            filters:
              branches:
                only:
-                 - nightly
+                 - main
     jobs:
       - build_ubuntu18_installer
       - build_centos7_installer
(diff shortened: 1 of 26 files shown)


sgpearse:
@stasj

> The changed signal is sometimes emitted multiple times for a single action with the same value.
Fixed, by adding a signal for the valueChanged and editingFinished events from the QSpinBox.

> As you type, it will emit a value changed signal for every key press.
This is still the behavior when you connect to the VIntSpinBox's ValueChanged signal, which can be used to change values interactively with the scroll-wheel.  I think the EditingFinished signal is what you may want, which will only emit on loss of focus, or return-press.

> Right-clicking on the widget causes the current value to be emitted as a changed value.
This was a weird one.  When right clicking on the widget, the QLineEdit within the QSpinBox would receive focus, then lose focus upon the drawing of the menu.  I've worked around it by having the VIntSpinBox track the current value and only emit EditingFinished if the value has changed.

StasJ:
>> The changed signal is sometimes emitted multiple times for a single action with the same value.
>
> Fixed, by adding a signal for the valueChanged and editingFinished events from the QSpinBox.
> 
>> Right-clicking on the widget causes the current value to be emitted as a changed value.
>
> This was a weird one. When right clicking on the widget, the QLineEdit within the QSpinBox would receive focus, then lose focus upon the drawing of the menu. I've worked around it by having the VIntSpinBox track the current value and only emit EditingFinished if the value has changed.

Great, thanks!

> This is still the behavior when you connect to the VIntSpinBox's ValueChanged signal, which can be used to change values interactively with the scroll-wheel. I think the EditingFinished signal is what you may want, which will only emit on loss of focus, or return-press.

Should `VIntSpinbox` have multiple signals? I believe we decided that for consistency, the `VWidget` would take care of figuring out when it was appropriate to register a change and it would do that with the `ValueChanged` signal.

sgpearse:
> Should VIntSpinbox have multiple signals? I believe we decided that for consistency, the VWidget would take care of figuring out when it was appropriate to register a change and it would do that with the ValueChanged signal.

This depends on whether we want to use a widget interactively.  ValueChanged will fire off as a user types a new number, or scrolls up/down on the spinbox.  EditingFinished will not fire as they type a new number - it will only fire when return is pressed, or the LineEdit loses focus.  I think we should anticipate the need to support both of these two use cases: the interactive and non-interactive modes.

StasJ:
> > Should VIntSpinbox have multiple signals? I believe we decided that for consistency, the VWidget would take care of figuring out when it was appropriate to register a change and it would do that with the ValueChanged signal.
> 
> This depends on whether we want to use a widget interactively. ValueChanged will fire off as a user types a new number, or scrolls up/down on the spinbox. EditingFinished will not fire as they type a new number - it will only fire when return is pressed, or the LineEdit loses focus. I think we should anticipate the need to support both of these two use cases: the interactive and non-interactive modes.

That sounds good to me. Currently other widgets that support this have a `ValueChanged` signal and a `ValueChangedIntermediate` signal to account for this.

sgpearse:
Got it.  I've changes the signal/slot names for consistency.

sgpearse (approved):

sgpearse:
@clyne this PR is a collaboration between Stas and myself.  I think it's correct, but I'm going to ask for stas' review just to make sure we're all on the same page.

clyne:
okey doke

sgpearse:
Github can't find @StasJ in its reviewer selector right now.  Who knows why.  Stas, I think we're good here but could you let us know?

Screen Shot 2021-02-24 at 8 21 45 AM

StasJ:
> Github can't find @StasJ in its reviewer selector right now. Who knows why. Stas, I think we're good here but could you let us know?

@sgpearse GitHub won't let me review my own PR but I approve your changes.

clyne (commented):
The spin box introduced in the Flow/Seeding tab doesn't seem to work. Shouldn't it? If I load DUKU, and create a Flow renderer, changing the integration steps has no effect.

sgpearse:
@clyne thanks for the catch John.  

There was a recent change regarding the ValueChangedIntermediate signal that introduced this bug.  I've been testing this on a different branch (#2622) and I missed this case.  It's fixed, and this widget now emits intermediate changes from the user.

interactiveStreamlines

clyne (approved):

PR #2615, opened 2021-02-03
Fix #2530 Setting constant colors for glyphs is non-intuitive

### LINKED ISSUE #2530: Setting constant colors for glyphs is non-intuitive

StasJ:
Fix #2530

Adds an explicit "Use Constant Color" checkbox and color selector to all the mentioned renderers.

Copy of #2602

### DIFF
--- a/apps/vaporgui/BarbEventRouter.cpp
+++ b/apps/vaporgui/BarbEventRouter.cpp
@@ -25,6 +25,7 @@ BarbEventRouter::BarbEventRouter(QWidget *parent, ControlExec *ce) : RenderEvent
     }));
     
     AddSubtab("Appearance", new PGroup({
+        new PColormapTFEditor,
         new PSection("Barbs", {
             (new PIntegerSliderEdit(BP::_xBarbsCountTag, "X Barbs"))->SetRange(1, 50),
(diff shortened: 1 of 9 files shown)


clyne:
Why was #2602 closed and this PR submitted? I know we discussed issues with some of the PRs having not handled the migration from master to main correctly, but #2602 doesn't appear to be one of them that I can see.

StasJ:
It was one of the PRs that were messed up from the migration to main and it
has a comment from you asking me to fix it. After I made the changes that
were supposed to fix it, GitHub still showed the wrong diff so I just
recreated it. It appears that PR now has the correct diff so GitHub
probably was just delayed in updating its cache, however I have encountered
situations where the diff is never fixed by GitHub which is why I just
recreated the PR.

StasJ:
I closed this one and reopened the original.

PR #2616, opened 2021-02-04
Introduce clang-format checks on CircleCI, and client tools to perform local linting

sgpearse:
This PR addresses #2309.  There are two parts:

1) CircleCI configuration to test that code has been linted with our style

2) Tools so that we as developers can automatically lint the code when pushing, through a pre-push hook

When this gets merged merged developers will need to do the following:

- Install clang-format, version 11+
- Run the following script, which will install the pre-push hook in the .git/hooks directory of your local repository
`VAPOR/share/gitHooks/setupHooks.sh`

### DIFF
--- a/share/gitHooks/pre-push
+++ b/share/gitHooks/pre-push
@@ -1,0 +1,11 @@
+#!/bin/sh
+#
+
+# Run clang-format on all .cpp and .h files to be pushed
+for FILE in $(git diff --name-only origin/master |grep -E "\.h|\.cpp")
+do
+    echo "    Running clang format on  $FILE"
+    clang-format -i $FILE
+done
+
+git commit -a -m "clang-format pre-receive hook"
--- a/share/gitHooks/pre-push
+++ b/share/gitHooks/pre-push
@@ -2,7 +2,7 @@
 #
 
 # Run clang-format on all .cpp and .h files to be pushed
-for FILE in $(git diff --name-only origin/master |grep -E "\.h|\.cpp")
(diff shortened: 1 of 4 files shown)


clyne:
It looks like developers will have to run setupHooks.sh every time then clone the repo, correct? Can we add this as a Makefile target so that it's automated?

share/gitHooks/pre-push
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+#
+
+# Run clang-format on all .cpp and .h files to be pushed
+for FILE in $(git diff --name-only origin/master |grep -E "\.h|\.cpp")
```
StasJ, line 5: This is referencing master instead of main.

StasJ (changes requested):

sgpearse:
> It looks like developers will have to run setupHooks.sh every time then clone the repo, correct? Can we add this as a Makefile target so that it's automated?

@clyne we can, and it should be a simple 1 liner.  

However there is a general feeling that these hooks should be explicitly added by developers for two reasons, 

1. Some developers don't want scripts running without their knowledge and explicit approval
2. Some consider having scripts running like this to be a security threat

I don't feel strongly about this, so if you'd like it in this PR I can do it in short order.

share/gitHooks/pre-push
sgpearse: Thanks for the catch.

StasJ (commented):
Do we want to clang-format the entire file vs just the lines changed?

StasJ (approved):

Issue #1659, opened 2019-06-10
VDC that is missing higher resolution data still reports having it


StasJ:
1. Create a VDC file
2. Remove the higher resolution data files (*.nc{1,2,3})
3. Open the dataset in vapor
4. Add a grid renderer

The `Data Fidelity` widget shows the higher resolutions as available. Selecting them results in the lower resolution being resampled onto a higher grid which can result in lower quality images than the lowest setting. For example:

![Screen Shot 2019-06-10 at 5 29 49 PM](https://user-images.githubusercontent.com/2772687/59233364-5d23c200-8ba5-11e9-95bd-c7383cd1e79e.png)

PR #2618, opened 2021-02-06
Fix #2617 VariableExists results incorrect

### LINKED ISSUE #2617: VariableExists results incorrect
1. Delete the higher refinement levels for a VDC variable
2. Query if the higher refinement levels exist from the data manager

It will return true even though they have been deleted.

StasJ:
Opened a separate issue and PR for this one because it appears this bug has become a feature. If you open a dataset with missing higher resolution data, it will just load the highest available resolution. This is not perfect as it will also select a too high CRatio which creates artifacting. After this PR, VariableExists correctly returns false for higher resolutions when they do not exists, but now if you try to select a higher resolution it will throw an error. My other PR for #1659 greys out options to select resolutions which do not exist, however.

Fix #2617

### DIFF
--- a/lib/vdc/VDCNetCDF.cpp
+++ b/lib/vdc/VDCNetCDF.cpp
@@ -1047,7 +1047,7 @@ bool VDCNetCDF::variableExists(size_t ts, string varname, int level, int lod) co
 
     for (int i = 0; i <= lod; i++) {
         struct stat statbuf;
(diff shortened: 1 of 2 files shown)


clyne (approved):

clyne:
@StasJ this PR has been approved but has conflicts that need to be resolved.

sgpearse:
@clyne @stasj I accidentally committed an extra file during the merge.  It's been resolved.

Issue #904, opened 2018-06-16
Ambiguous language in VDC documentation


StasJ:
Every time I run into this I have to do a double take:

```
[in] lod | A value of -1 indicates the maximum approximation level defined for the VDC
```

`maximum approximation` is an ambiguous phrase. It could technically refer to both the lowest and highest level of detail. My natural inclination is that it means the lowest level of detail, although I think it is supposed to mean the latter.

Also since we're already on the topic, the two main compression related parameter names are `level` and `level of detail`, both referring to completely different things.

PR #2619, opened 2021-02-06
Fix #1659 VDC that is missing higher resolution data still reports having it

### LINKED ISSUE #1659: VDC that is missing higher resolution data still reports having it

StasJ:
Fix #1659

Adds the ability to automatically grey out missing resolutions but without merging #2618, all resolutions are always returned as existing by the data mgr.

### DIFF
--- a/apps/vaporgui/PFidelitySection.cpp
+++ b/apps/vaporgui/PFidelitySection.cpp
@@ -50,26 +50,25 @@ void PQuickFidelitySelector::updateGUI() const
     int lod = rp->GetCompressionLevel();
     int ref = rp->GetRefinementLevel();
     int n = items.size();
-    
+
     _vComboBox->SetIndex(paramsToSimple(n, nLod, nRef, lod, ref));
-    
+
     long ts = rp->GetCurrentTimestep();
     for (int i = 0; i < n; i++) {
         simpleToParams(n, nLod, nRef, i, &lod, &ref);
         bool exists = dm->VariableExists(ts, vn, 0, ref);
-        if (!exists)
-            _vComboBox->SetItemEnabled(i, false);
+        if (!exists) _vComboBox->SetItemEnabled(i, false);
     }
 }
 
 void PQuickFidelitySelector::dropdownTextChanged(std::string)
 {
     auto *rp = (RenderParams *)getParams();
-    auto dm = getDataMgr();
-    auto vn = rp->GetFirstVariableName();
-    int  nLod = dm->GetCRatios(vn).size();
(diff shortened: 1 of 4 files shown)


shaomeng:
In my test, removing a file (e.g. ANGLE.nc3) doesn't trigger higher resolutions being grayed out?

sgpearse (approved):

clyne:
@shaomeng needs to re-examine. Note, it's the levels-of-detail that may be missing, not resolution.

shaomeng:
In my test, after removing a file (e.g., ANGLE.nc3), the highest level of detail, which is 1:1, is still shown as black and select-able. I believe it should be grayed out.

StasJ:
> In my test, after removing a file (e.g., ANGLE.nc3), the highest level of detail, which is 1:1, is still shown as black and select-able. I believe it should be grayed out.

Main with the fix from #2618 had not been merged into this PR. It should work now.

shaomeng:
It seems with the latest change, removing the highest numbered nc file (e.g. ANGLE.nc3) will gray out the highest refinement level, which is represented as a dimension, but not level of detail, which is represented as a ratio. Maybe @clyne can clarify which one is the correct value to gray out?

clyne:
> It seems with the latest change, removing the highest numbered nc file (e.g. ANGLE.nc3) will gray out the highest refinement level, which is represented as a dimension, but not level of detail, which is represented as a ratio. Maybe @clyne can clarify which one is the correct value to gray out?

You are correct @shaomeng . It should be the highest LOD that is grayed out, not the Refinement Level.

StasJ:
> > It seems with the latest change, removing the highest numbered nc file (e.g. ANGLE.nc3) will gray out the highest refinement level, which is represented as a dimension, but not level of detail, which is represented as a ratio. Maybe @clyne can clarify which one is the correct value to gray out?
> 
> You are correct @shaomeng . It should be the highest LOD that is grayed out, not the Refinement Level.

Good catch, thanks. I always mix up "Level of Detail" and "Refinement Level" or as they're named in the API: `lod` and `level` (#904).

It is now using the correct one.

shaomeng (approved):

Issue #1509, opened 2019-04-22
VFileSlector: more intelligent space and layout management 


shaomeng:
**Describe the bug**
The current composite widget `VFileSelector` doesn't seem to be very intelligent in managing spaces/layouts. This is not ideal, since a key promise of these rebranded `VaporWidgets` is to manage layout/spaces intelligently. Without such capability, the benefit of `VaporWidgets` is in doubt. 

**To Reproduce**
1. Checkout a branch that makes use of `VFileSelector`. In this report I'd use `flow`.
2. Create a flow renderer.
3. Switch to `Seeding` tab, and select a file at the bottom.
4. Notice the layout of the `VFileSelector` in the screenshot. There's much white space between the label and the button, while not enough space to display the full name of the file in the text box.
5. A more intelligent layout manager would give enough space for the text box to display the full filename when there's still white space left. 

**Screenshots**
![Screenshot 2019-04-22 17 25 38](https://user-images.githubusercontent.com/814885/56539641-b7988e80-6523-11e9-879d-4cb01bedbcfe.png)

clyne:
Related to #265

PR #2622, opened 2021-02-12
New Preferences Menu

### LINKED ISSUE #2257: Data-independent settings (e.g. Cache size) can't be changed unless a data set is open
The VAPOR GUI Settings/Cache Size cannot be changed unless  the application is open AND a data set is loaded. This is problematic because a user may not be able to open a data set until AFTER the cache size is changed.

sgpearse:
Fixes #2257 

This draft PR duplicates our "Settings" tab into a PWidget based "Preferences" dialog that is accessible in the top-level "vapor" menu on OSX.  On other systems, it's in the "File" menu.  The aim is to remove these options from the Settings tab, in favor of a Preferences menu.
Screen Shot 2021-02-12 at 1 19 09 PM

Please disregard the following flies, as they are changes that I needed from another PR, or are workarounds for bugs.  I can share more details if asked.
`FlowEventRouter.cpp, VIntSpinBox.cpp, VIntSpinBox.h, VPushButton.cpp`

ToDo:
1. Remove the Settings tab
2. Rename "Settings/AppSettings" to "Preferences"
2. Add the Flow, Transfer Function, and Python directory options that were previously disabled.  This is assuming that we still want to keep these settings.

My questions:
1. Older params classes use the _init() function to restore default values (as I use in this PR).  Newer classes use a SetDefaults() function.  Which one do we want to use, and should it be a pure virtual member of ParamsBase?
2. Some of our SettingsParams setters have hard coded default values.  We also have hard coded default values in the _init/SetDefault functions.  How do we make these consistent?  Would default value tags for each parameter work?


sgpearse:
@clyne @shaomeng @StasJ - This is a draft PR because I want to review the usability of this change.

Please check out this branch and see the menu as shown in the comment above.

clyne:
> This draft PR duplicates our "Settings" tab into a PWidget based "Preferences" dialog that is accessible in the top-level "vapor" menu on OSX. On other systems, it's in the "File" menu. The aim is to remove these options from the Settings tab, in favor of a Preferences menu.
> Screen Shot 2021-02-12 at 1 19 09 PM
> 
> Please disregard the following flies, as they are changes that I needed from another PR, or are workarounds for bugs. I can share more details if asked.
> `FlowEventRouter.cpp, VIntSpinBox.cpp, VIntSpinBox.h, VPushButton.cpp`
> 
> ToDo:
> 
> 1. Remove the Settings tab
> 2. Rename "Settings/AppSettings" to "Preferences"
> 3. Add the Flow, Transfer Function, and Python directory options that were previously disabled.  This is assuming that we still want to keep these settings.
> 
> My questions:
> 
> 1. Older params classes use the _init() function to restore default values (as I use in this PR).  Newer classes use a SetDefaults() function.  Which one do we want to use, and should it be a pure virtual member of ParamsBase?
> 2. Some of our SettingsParams setters have hard coded default values.  We also have hard coded default values in the _init/SetDefault functions.  How do we make these consistent?  Would default value tags for each parameter work?

Each params class has two constructors: one that initializes the params class from scratch, and one that initializes the params class from an existing params database (XML tree). The private _init function is a convention that is used in the case of the latter for many of the params classes, nothing more. With one exception (RenderParams), and I don't know the reasons, none of the params classes provide public methods for changing the defaults. Is there a reason for providing this capability now?

clyne:
> @clyne @shaomeng @StasJ - This is a draft PR because I want to review the usability of this change.
> 
> Please check out this branch and see the menu as shown in the comment above.

From a usability standpoint it looks good to me. I'd say move forward.

clyne:
@shaomeng please review for usability only

shaomeng:
Looking good scott. Only one GUI element to notice: the file path widget doesn't seem to be using the available real estate while chopping the full file path. I believe there was a fix already, and it'd be good to apply that fix here! 

Screenshot from 2021-02-23 21-05-38

sgpearse:
@shaomeng Hey sam, this is using the file selector that's currently on our main branch.  If it's still not looking right to you, I think we should re-open #1509 instead of addressing it here.

sgpearse:
> The private _init function is a convention that is used in the case of the latter for many of the params classes, nothing more. With one exception (RenderParams), and I don't know the reasons, none of the params classes provide public methods for changing the defaults. Is there a reason for providing this capability now?

@clyne Yes, I'm using the public Init() function (formally _init()) so that the GUI can restore default values with a button-press.

clyne:
> > The private _init function is a convention that is used in the case of the latter for many of the params classes, nothing more. With one exception (RenderParams), and I don't know the reasons, none of the params classes provide public methods for changing the defaults. Is there a reason for providing this capability now?
> 
> @clyne Yes, I'm using the public Init() function (formally _init()) so that the GUI can restore default values with a button-press.

Got it. To your original question, the Params class base does not provide API for reinitializing the defaults. You could either add an Init (or SetDefaults) function only to your preferences params class, or create a virtual method for all params classes. I guess I'd steer away from the latter because of the work that is involved and it's not clear that there is a need.

apps/vaporgui/SettingsParams.cpp
```diff
@@ -150,6 +151,7 @@ void SettingsParams::SetCacheMB(long val)
 {
     if (val < 0) val = 8000;
     SetValueLong(_cacheMBTag, "Set cache size", val);
+    SaveSettings();
```
clyne, line 154: SaveSettings() can fail and return an error code. You're ignoring the return value and the user will never see the error message. Is there a reason that SaveSettings has been added to every setter and not simply called once higher up the stack as was done before?

apps/vaporgui/AppSettingsMenu.cpp
```diff
@@ -0,0 +1,50 @@
...
+                //(new PIntegerInputHLI<SettingsParams>("Number of threads (0 for available num. of CPU cores)", &SettingsParams::GetNumThreads, &SettingsParams::SetNumThreads)->SetRange(0,1024)),
+                new PIntegerInputHLI<SettingsParams>("Number of threads (0 for available num. of CPU cores)", &SettingsParams::GetNumThreads, &SettingsParams::SetNumThreads),
+                new PIntegerInputHLI<SettingsParams>("Cache size (Megabytes)", &SettingsParams::GetCacheMB, &SettingsParams::SetCacheMB),
+                new PCheckboxHLI<SettingsParams>("Lock window dimensions", &SettingsParams::GetWinSizeLock, &SettingsParams::SetWinSizeLock),
+                (new PIntegerInputHLI<SettingsParams>("Width", &SettingsParams::GetWinWidth, &SettingsParams::SetWinWidth))->EnableBasedOnParam(SettingsParams::_winSizeLockTag),
+                (new PIntegerInputHLI<SettingsParams>("Height", &SettingsParams::GetWinHeight, &SettingsParams::SetWinHeight))->EnableBasedOnParam(SettingsParams::_winSizeLockTag),
+                new PStringDisplay("", "*Vapor must be restarted for these settings to take effect"),
```
StasJ, line 31: There is a PLabel class for this

apps/vaporgui/AppSettingsMenu.h
```diff
@@ -0,0 +1,27 @@
...
+//! \ingroup Public_GUI
+//! \brief Menu for global application preferences
+//! \author Scott Pearse
+
+class AppSettingsMenu : public QDialog {
+    Q_OBJECT
+
+public:
+    AppSettingsMenu(QWidget *parent);
+    void Update(SettingsParams *sp);
```
StasJ, line 23: There is an Updatable interface to standardize this.

shaomeng:
When "Automatically Save Session" is turned off, the two auto save related options should be grayed out, just the same as the behavior of the "Width" and "Height" options in the same screen. 

Screenshot from 2021-03-09 19-32-33

apps/vaporgui/AppSettingsMenu.h
sgpearse: I'm not sure what you're referring to.  We don't currently have a QDialog based PWidget, and our current dialogs (ParamsWidgetDemo, PythonVariables, Statistics, RenderHolder, etc) implement the Update() function it on their own.

Do you mean that this should be inheriting from PWidget, and use its Update() function?

apps/vaporgui/AppSettingsMenu.cpp
sgpearse: Got it.  Fixed.

apps/vaporgui/AppSettingsMenu.h
StasJ: There is an `Updatable` class. All the other dialogs you mention were implemented before.

shaomeng:
Oops, I thought I was posting a new comment but it ended up changing my old comment. Anyway, my new comment is supposed to be this:  https://github.com/NCAR/VAPOR/pull/2622#issuecomment-794765021

apps/vaporgui/SettingsParams.cpp
sgpearse: SaveSettings is now performed in Update(), and the user gets notified upon a bad error code.

apps/vaporgui/AppSettingsMenu.h
sgpearse: Fixed.

sgpearse:
@shaomeng The Close button is now described as "Save and Close"

clyne (commented):
At least on the Mac there appears to be no way to close the preferences popup without saving the changes; I believe you need a cancel button.

sgpearse:
@clyne This would be new functionality that didn't exist before.  Previously and currently, all changes were written when a value is changed.

We'd need to talk about how this would work.  Right now, every time a PWidget changes, Update() gets triggered and we save the file.

One thought is to save the Params state during QDialog::show().  It could be restored on QDialog::reject() when the cancel button is pushed.  But I'm not sure how restoration should happen.  ParamsMgr::LoadState() doesn't look like the right tool.

We could also try to count the number of changes that happen while the dialog is open, and issue that number of Undo's upon Cancel.  That feels a little hacky to me though.

clyne:
> @clyne This would be new functionality that didn't exist before. Previously and currently, all changes were written when a value is changed.
> 
> We'd need to talk about how this would work. Right now, every time a PWidget changes, Update() gets triggered and we save the file.
> 
> One thought is to save the Params state during QDialog::show(). It could be restored on QDialog::reject() when the cancel button is pushed. But I'm not sure how restoration should happen. ParamsMgr::LoadState() doesn't look like the right tool.
> 
> We could also try to count the number of changes that happen while the dialog is open, and issue that number of Undo's upon Cancel. That feels a little hacky to me though.

@sgpearse good point. I forgot that the "save" part of "save and close" is meaningless because stuff gets saved automatically. I'd suggest either:

1. revert the name of the button from "save and close" to "close". Otherwise people are going to want to know how to cancel. Not having to explicitly save is consistent with preferences for other applications. Plus "undo" can be used to revert a change.

or 

2. Do NOT perform the save automatically and have an explicit button for doing so (and one to cancel)

Either options works for me.

cc @shaomeng

sgpearse:
@clyne and @shaomeng 

I've pushed changes for option 1.

Option 2 is misleading because we would have two different ways of changing our settings: 
1) in the current session (user changes a parameter + close)
2) in current and future sessions (user changes a parameter, + save)

I think I mentioned yesterday that option 1 is how OSX does things.  Changes are simply registered as they're made, and there's no save or apply button.  For example, check out the Dock settings menu below.

Screen Shot 2021-03-18 at 12 35 34 PM

clyne (approved):

shaomeng:
> @clyne and @shaomeng
> 
> I've pushed changes for option 1.
> 
> Option 2 is misleading because we would have two different ways of changing our settings:
> 
>     1. in the current session (user changes a parameter + close)
> 
>     2. in current and future sessions (user changes a parameter, + save)
> 
> 
> I think I mentioned yesterday that option 1 is how OSX does things. Changes are simply registered as they're made, and there's no save or apply button. For example, check out the Dock settings menu below.
> 
> Screen Shot 2021-03-18 at 12 35 34 PM

I think this is a good example screenshot. I'd say we completely remove the "Close" button, just like this screenshot shows.

The whole purpose is to make things clear, i.e., avoid confusion. When a button named "Close" is present, my immediate question was that if my changes are saved or not, and the nature of the changes are that the changes don't have immediate effects visible.

How do @sgpearse and @clyne think of the GUI change to remove "Close" button completely?

shaomeng (approved):

StasJ (approved):

Issue #2178, opened 2020-01-29
Colorbar customization


phawbeck:
I would like to be able to customize the ticks for the colorbar in Vapor plots. As is, the colorbar tick numbers and decimals can be adjusted, but the numbers are still (as far as I can see) arbitrarily selected. It would be helpful to be able to specify exactly what ticks are to be labeled through a range (min tick, max tick, step) or a list (1,5, 10, 100, 1000, etc).

**Impact**
(*Please click check boxes AFTER submitting ticket*)
- [ ] High - User productivity significantly improved
- [ ] Medium - User productivity partially improved
- [ ] Low - User productivity trivially improved

PR #2625, opened 2021-02-24
New Colorbar

### LINKED ISSUE #717: Color bar: moves out of visualizer
To reproduce: 
1. Open Katrina, create a Barb renderer, go to Appearance tab.
2. Enable colorbar.
3. Type "1" in the "X Position" textbox.
4. The colorbar moves out of the visualizer.

### LINKED ISSUE #808: Colorbar title: why isn't it in the colorbar box?

### LINKED ISSUE #898: ColorBar: values not able to read.
Platform: cisl-vapor
Issue: colorbar not able to read, as the screenshot shows.
To reproduce:
1. Open VAPOR, Import NetCDF-CF data: `/little_glade/data/Source/MPAS/MPAS-O/hist.mpas-o.0001-01-01_00.00.00.nc`
1. Create a Barb Renderer, enable it.
1. Enable the color bar. Notice the numbers are not able to read.

screenshot from 2018-06-11 12-17-49

### LINKED ISSUE #1308: Windows: Undo/Redo modifies the X size of the colorbar
This problem does not exhibit itself on OSX.

To reproduce:
1) Import any Katrina WRF file
2) Create a 2D renderer, enable it
3) Enable its colorbar
4) Modify the renderer geometry
5) Undo, and the colorbar will expand along the X axis

### LINKED ISSUE #1376: Disagreement between TF widget and Colorbar
In this case, I created a slice renderer for `dbz` for Lee Orf's tornado dataset with the default TF and added a colorbar.

screen shot 2019-01-24 at 10 20 05 pm

### LINKED ISSUE #1387: Barb colorbar jumps
The color bar inside of Barb renderer is able to jump from an initial width to a final width. 

peek 2019-01-25 16-36

To reproduce:
1. Open VAPOR with DUKU.
2. Create a barb renderer. 
3. Enable color bar. It jumps.

### LINKED ISSUE #1599: Contour: colorbar X Width jump
### Describe the bug
The width of colorbar increases as font size increases. More interestingly, it also increases as the font size decrease. 

To reproduce:
1. Open VAPOR, load VDC data: `/little_glade/data/VDC/vapor-3.1.0/CAM/AdamPhillips/F2000C5_new120_025/F2000C5_new120_025-3.1.0.nc`
2. Enable colorbar.
3. Increase font size. Notice that the color bar width also increases.
4. Also notice that the height of the color bar does not increase, so it deesn't help readability. 
5. When the font is at a large size, decrease the font size. Notice that the color bar still jumps wider for one time. (The following animation captures this behavior.)
5. Also notice that when the font size is decreased, the color bar width does not decrease. 

Peek 2019-06-03 16-40

### LINKED ISSUE #1605: Colorbar gradient, ticks, and numbers don't align
If you configure the `Colorbar Settings` like below, nothing lines up:

colorbar

### LINKED ISSUE #1688: Colorbar title is rendered outside of colorbar
It does not satisfy its function as a title. Also, while the colorbar can be used with a white background on a white rendering because it has a border, the title is rendered on the outside in white so it is not visible.

StasJ:
### Automatic font sizing
ticks

### Auto Scaling
Automatically scales for all resolutions.

### Significant Figures
Rather than requiring the user to manually enter the number of decimals for every variable, it now uses significant figures so the default of 4 will round off to the first 4 digits and only include decimals if needed. This also solves the issue that the old color bar would show garbage values that are outside of the precision of floats. For example, below we have a color bar that goes from 42 to 1170000. On the left, is the old color bar and on the right the new one. The old shows garbage values that are outside of the float precision range.

old new

### Scientific notation
sci_notation

### Pixel-Perfect Borders, Ticks, Padding, etc.
All graphical elements are corrected for aspect ratio and aligned to the pixel grid. Old on left, new on right.

old tall new tall

### Position Quick Select Widget
Custom widget which allows for quick and easy placement in commonly used annotation locations.

pos

### All settings play well together
All settings give a valid result. With the previous color bar it was very easy to end up with something like below. Often as you would fix one parameter, it would break other parameters, creating a frustrating feedback loop.

bad

### Correct tick placement and start/stop
It was not possible to have ticks at the min/max in the old color bar so it would never match up with the transfer function. The old ticks would also not actually point to the correct datapoint that the number shows. As you can see in the screenshot above the ticks are somehow outside of the color bar entirely. Both transfer functions below go from 42 to 1170000.

old new

### Titles
Auto-adjusts the size for titles. The old widget just rendered it outside the annotation window.

title title_long

### Simple intuitive use
- Position works independent of size. (before if you adjusted one, you had to adjust the other. Also if position + size > 1 it would clip off the screen)
- Height parameter adjusts height, while scale scales the entire widget. Both parameters seamlessly work together.

use

### Issues Fixed
Fix #898
Fix #1376
Fix #1688
Fix #1605
Fix #1599
Fix #1387
Fix #717
Fix #808
Fix #1308


clyne (commented):
Looks great, Stas. It will nice to finally have this thing cleaned up. A couple of things I've noticed.

No colorbar is displayed for the wireframe renderer. 

You indicate that this PR closes #2178. I don't believe that is the case. I think what Pat is asking for is the ability to explicitly place the tick marks by setting, for example, a minimum value and an increment (similar to what is done for specifying contour lines). This allows you to put the tick marks at round numbers.

StasJ:
> Looks great, Stas. It will nice to finally have this thing cleaned up. A couple of things I've noticed.
> 
> No colorbar is displayed for the wireframe renderer.

I think the wireframe renderer is breaking something in the OpenGL state. I am still looking into it.

> You indicate that this PR closes #2178. I don't believe that is the case. I think what Pat is asking for is the ability to explicitly place the tick marks by setting, for example, a minimum value and an increment (similar to what is done for specifying contour lines). This allows you to put the tick marks at round numbers.

Here is why I did that:
- The requests are:
  - Set a minimum value
  - Set an increment
  - Place ticks at round numbers
  - Use tick values from a list
- With the old color bar, it was not at all possible to do any of those things.
  - Set a minimum value: **No** - All tick values are arbitrary<sup>[1]</sup>
  - Set an increment: **No** - All tick values are arbitrary<sup>[1]</sup>
  - Place ticks at round numbers: **No** - The tick values are always displayed with more precision than they actually have
  - Use tick values from a list: **No**
- With the new color bar
  - Set a minimum value: **Yes** - The minimum value will be exactly the lower bound of the transfer function
  - Set an increment: **Yes** - Since you have the exact lower and upper bounds, you can set the number of ticks to 1 + (upper-lower) / increment
  - Place ticks at round numbers: **Yes** - With the above you can already get round numbers theoretically but floating point error would prevent this. If you lower the number of significant figures, it will round off the numbers.
  - Use tick values from a list: **No** but this seems like the least important use case but requires by far the most work and Pat seems to suggest it as an alternative to a minimum with an increment so I don't think it is worth the effort to add it.

For example if you have a TF with a range of 0 to 100, here is the default color bar:



Decrease the sig figs by 2:



If you want an increment of 10, increase the ticks from 6 to 11:



<sup> [1] Technically they're not arbitrary, but as far as the end user is concerned they are.</sup>

clyne:
I understand your point, but the implications are that if you want have tick marks at nice round numbers you have to define your transfer function accordingly, which may be a limitation that users do not want. @phawbeck what are your thoughts on this?

phawbeck:
Hi Stas and John, I hope everything is well! I think this is a great improvement on the colorbars and am happy to see the new adjustments! I see what John is saying in that there are a lot of times I'll want the lowest rendered value to be whatever the minimum is (let's just say that's 0.34) and the maximum (e.g., 52), but would want ticks to be at some round numbers to make it clean (e.g., 10, 30, 50). Is there anyway to disassociate the minimum tick labels from the actual minimum of the TF?
Thanks a lot for looking into this!

StasJ:
> Hi Stas and John, I hope everything is well! I think this is a great improvement on the colorbars and am happy to see the new adjustments! I see what John is saying in that there are a lot of times I'll want the lowest rendered value to be whatever the minimum is (let's just say that's 0.34) and the maximum (e.g., 52), but would want ticks to be at some round numbers to make it clean (e.g., 10, 30, 50). Is there anyway to disassociate the minimum tick labels from the actual minimum of the TF?
> Thanks a lot for looking into this!

@phawbeck Thanks for your feedback and I see your point. For arguments sake and to explain my rational, I have recreated your scenario. Leaving the range as 0.34 to 52, below is the default Colorbar you would get. If you decrease the sig figs, you will get reasonable clean ticks.

sig

In this case where you just decrease the sig figs until you round to the tens place, you have a maximum error of 4% in color gradient. Here is what 4% error looks like:

     


Another option is to change the range to 0 to 50. It will still render 0.34 to 52: 50-52 will not be clipped and the 0-0.34 will be 99.6% blue rather than 100% blue which would be indiscernible. In this case the last 50-52 will have the same color. Vapor's default colormap has had its right color control point not quite at 100% for years so the top of the TF range will have the same color and nobody's ever noticed.

map

I guess in conclusion @clyne maybe we need to have further discussion on what the abilities should be. Since there weren't any deliverables for this, my goal was to fix all the issues with the old Colorbar and make it do one thing well with a very simple interface. 

Also I fixed the issue with the Colorbar not rendering for the wireframe renderer.

shaomeng:
Not sure if it's intended, but introducing a while rectangle in the rendering doesn't seem cool. 

Screenshot from 2021-03-01 15-26-00

clyne:
@StasJ changing the significant figures as you've demonstrated will work in some cases, but is not a general solution. If you want to get an idea of the way scientists like to lay out tick marks take a look at the geocat gallery: https://geocat-examples.readthedocs.io/en/latest/

In particular, for example:

https://geocat-examples.readthedocs.io/en/latest/gallery/Masking/NCL_mask_2.html#sphx-glr-gallery-masking-ncl-mask-2-py
https://geocat-examples.readthedocs.io/en/latest/gallery/Masking/NCL_mask_5.html#sphx-glr-gallery-masking-ncl-mask-5-py

I'd suggest either providing the missing functionality, or simply leaving #2178 open so that we can address it later.

shaomeng:
Following the steps below you'll get a buggy colorbar:

1) Compile the code in "Release" mode
2) Open VAPOR with one of Leigh Orf's `cm1.nc` files.
3) Create a Slice renderer. Enable it with all default parameters.
4) Go to "Annotation" tab and click on "Show Colorbar"
5) You'll get the image blow:

Screenshot from 2021-03-02 17-30-39

StasJ:
@clyne I removed issue from the fixed list. I think we should see if the new functionality is sufficient and we can go back and add the addition features if needed.
@shaomeng I fixed the issue, thanks. It was due to a non-deterministic memory error which for some reason I could never reproduce on my computer but I was able to reproduce it on a different system.

clyne:
Sounds good, @StasJ

clyne (approved):

shaomeng (approved):
That's pretty slick colorbar, good job Stas!

sgpearse (approved):

clyne:
@stasj this PR is good to go as soon as you resolve the conflicts. Thanks.

StasJ:
@clyne Done.

PR #2628, opened 2021-02-26
Retain python editor's script after closing the dialog

### LINKED ISSUE #1225: Need to re-open python script when opening script window
1. Create a script
2. Return to vapor
3. Open script editor

The script editor will be blank and you have to manually re-open. Can be confusing

sgpearse:
Fixes #1225 - the python editor keeps its script after closing the dialog.

### DIFF
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -2253,9 +2253,10 @@ void MainForm::launchPythonVariables()
 {
     if (!_pythonVariables) {
         _pythonVariables = new PythonVariables(this);
-        connect(_pythonVariables, &QDialog::finished, this, &MainForm::_pythonClosed);
     }
-    if (_controlExec) { _pythonVariables->InitControlExec(_controlExec); }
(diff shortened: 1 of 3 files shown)


apps/vaporgui/MainForm.cpp
```diff
@@ -2253,7 +2253,6 @@ void MainForm::launchPythonVariables()
 {
     if (!_pythonVariables) {
         _pythonVariables = new PythonVariables(this);
-        connect(_pythonVariables, &QDialog::finished, this, &MainForm::_pythonClosed);
```
StasJ, line 2256: It seems `MainForm::_pythonClosed` is now dead code so it may make sense to delete it?

StasJ (changes requested):

apps/vaporgui/MainForm.cpp
sgpearse: Makes sense.  Fixed.

StasJ (approved):

PR #2631, opened 2021-03-01
Track active tabs and automatically activate appropriate region selector

### LINKED ISSUE #991: Need Manipulators to be activated by the Geometry tab
If our mouse mode Manipulators were activated whenever the current renderer's Geometry tab was active, we may be able to remove the "Modes" combo box from the main title bar.

screen shot 2018-09-06 at 2 28 19 pm

### LINKED ISSUE #1695: Naviage and Navigation do completely different things and are right next to eachother
Throughout the tutorial we often directed users to click on the `Navigation` tab. In most cases users ended up clicking the `Navigate` dropdown which is right next to it.

StasJ:
Fix #1695
Fix #991

### DIFF
--- a/apps/vaporgui/RenderEventRouterGUI.cpp
+++ b/apps/vaporgui/RenderEventRouterGUI.cpp
@@ -3,9 +3,9 @@
 #include <QScrollArea>
 #include "VContainer.h"
 
-const std::string RenderEventRouterGUI::VariablesTabName  = "Variables";
+const std::string RenderEventRouterGUI::VariablesTabName = "Variables";
 const std::string RenderEventRouterGUI::AppearanceTabName = "Appearance";
-const std::string RenderEventRouterGUI::GeometryTabName   = "Geometry";
+const std::string RenderEventRouterGUI::GeometryTabName = "Geometry";
 const std::string RenderEventRouterGUI::AnnotationTabName = "Annotation";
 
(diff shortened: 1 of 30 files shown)


apps/vaporgui/RenderEventRouterGUI.cpp
```diff
@@ -0,0 +1,85 @@
...
+    blockSignals(false);
+
+    auto pm = _controlExec->GetParamsMgr();
+    bool undoEnabled = pm->GetSaveStateUndoEnabled();
+    pm->SetSaveStateUndoEnabled(false);
+    getGUIStateParams()->SetActiveTab(tabText(i).toStdString());
+    pm->SetSaveStateUndoEnabled(undoEnabled);
+}
+
+// clang-format off
```
clyne, line 75: I'm assuming that these comments are directives to the clang formatter. There seems to be several instances where your disabling the formatter for what looks like pretty basic code. Is there a reason for this?

StasJ: After running the formatter I found that my original formatting was clearer, however I can undo that if you'd like.

clyne: It would seem that we're defeating the purpose of using a formatter if we're continuously disabling it :-)

StasJ: Ok done.

clyne (approved):
Looks great, @StasJ

clyne (approved):

sgpearse (approved):

PR #2633, opened 2021-03-02
Adjust random variable bias strength

### LINKED ISSUE #2480: Random bias seeding option for flow is not strong enough
The random seeding with bias option for the flow renderer does not give the desired results; it is not possible to constrain seeding points to regions of the data with large values. 

To reproduce:
1. Load the CM1 May 24 control run data.
2. Enable an isosurface renderer with the 'vortmag' variable and set the isosurface value to 0.5. Note that all of the structures are found in a small region around the vortex core.
3. Enable a flow renderer. 
4. Set the integration steps to 10.
5. Select the "random with bias" seeding option and select 'vortmag' as the bias variable
6. Change the bias value up to the maximum, 10, and observe the placement of seed points. While there are certainly more seed locations around the the visible isosurface structures there are plenty more found in regions with lower values.  We should be more quantitative about this, making use of the histogram to define how a bias weight impacts the distribution of seed points

shaomeng:
The random seed bias is adjusted to be in range (-100, 100).

Fixes  #2480

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -8,7 +8,7 @@ AlignConsecutiveBitFields: true
 AlignConsecutiveDeclarations: true
 AlignConsecutiveMacros: true
 AlignEscapedNewlines: Left
-AlignOperands: true
+AlignOperands: AlignAfterOperator
 AlignTrailingComments: true
 AllowAllArgumentsOnNextLine: true
 AllowAllConstructorInitializersOnNextLine: true
@@ -49,7 +49,7 @@ ForEachMacros:
 IncludeBlocks:   Preserve
 IndentCaseBlocks: false
 IndentCaseLabels: false
-# IndentExternBlock: NoIndent
+IndentExternBlock: NoIndent
 IndentGotoLabels: true
 IndentPPDirectives: BeforeHash
 IndentWidth: 4
--- a/.clang-format
+++ b/.clang-format
@@ -49,7 +49,7 @@ ForEachMacros:
 IncludeBlocks:   Preserve
 IndentCaseBlocks: false
 IndentCaseLabels: false
-IndentExternBlock: NoIndent
+# IndentExternBlock: NoIndent
 IndentGotoLabels: true
 IndentPPDirectives: BeforeHash
 IndentWidth: 4
--- a/.clang-format
+++ b/.clang-format
@@ -8,7 +8,7 @@ AlignConsecutiveBitFields: true
 AlignConsecutiveDeclarations: true
 AlignConsecutiveMacros: true
 AlignEscapedNewlines: Left
-AlignOperands: AlignAfterOperator
+AlignOperands: true
 AlignTrailingComments: true
 AllowAllArgumentsOnNextLine: true
 AllowAllConstructorInitializersOnNextLine: true
--- a/apps/vaporgui/FlowEventRouter.cpp
+++ b/apps/vaporgui/FlowEventRouter.cpp
@@ -53,7 +53,7 @@ FlowEventRouter::FlowEventRouter(QWidget *parent, ControlExec *ce) : RenderEvent
             }),
             (new PShowIf(FP::_seedGenModeTag))->Equals((int)FlowSeedMode::RANDOM_BIAS)->Then({
                 (new PIntegerSliderEdit(FP::_randomNumOfSeedsTag, "Seed count"))->SetRange(1, 500),
-                (new PDoubleSliderEdit(FP::_rakeBiasStrength, "Bias weight"))->SetRange(-10, 10),
+                (new PIntegerSliderEdit(FP::_rakeBiasStrength, "Bias weight"))->SetRange(-100, 100),
                 new PVariableSelector(FP::_rakeBiasVariable)
             }),
             (new PShowIf(FP::_seedGenModeTag))->Equals((int)FlowSeedMode::LIST)->Then({
--- a/include/vapor/FlowParams.h
+++ b/include/vapor/FlowParams.h
@@ -105,8 +105,8 @@ public:
     std::string GetRakeBiasVariable() const;
     void        SetRakeBiasVariable(const std::string &);
 
-    float GetRakeBiasStrength() const;
-    void  SetRakeBiasStrength(float);
+    long GetRakeBiasStrength() const;
+    void SetRakeBiasStrength(long);
 
     int  GetPastNumOfTimeSteps() const;
     void SetPastNumOfTimeSteps(int);
--- a/include/vapor/FlowRenderer.h
+++ b/include/vapor/FlowRenderer.h
@@ -72,7 +72,7 @@ private:
     long               _cache_randNumOfSeeds = 5;
     int                _cache_seedInjInterval = 0;
     int                _cache_pastNumOfTimeSteps = 0;
-    float              _cache_rakeBiasStrength = 0.0f;
+    long               _cache_rakeBiasStrength = 0;
     double             _cache_deltaT = 0.05;
     FlowSeedMode       _cache_seedGenMode = FlowSeedMode::UNIFORM;
     FlowDir            _cache_flowDir = FlowDir::FORWARD;
--- a/lib/params/FlowParams.cpp
+++ b/lib/params/FlowParams.cpp
@@ -78,7 +78,7 @@ FlowParams::FlowParams(DataMgr *dataManager, ParamsBase::StateSave *stateSave) :
     SetValueDouble(PhongSpecularTag, "", 0);
     SetValueDouble(PhongShininessTag, "", 2);
 
-    // Give the random bias variable the same as color mapping variable
+    // Give the random bias variable the same as color mapping variable.
     auto colorvar = GetColorMapVariableName();
     SetValueString(_rakeBiasVariable, "which variable to bias with", colorvar); 
 }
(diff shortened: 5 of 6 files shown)


clyne:
@shaomeng I'm not sure how your weighting algorithm works, but I'm still not able get the desired results. See the attached images generated with VAPOR2 and VAPOR3, both of which have set the bias to the maximum.

VAPOR2:

vapor2_flow

VAPOR3:

vapor3_flow

shaomeng:
> @shaomeng I'm not sure how your weighting algorithm works, but I'm still not able get the desired results. See the attached images generated with VAPOR2 and VAPOR3, both of which have set the bias to the maximum.

@clyne  It works this way:
1) when you specify a number of seeds to generate (say 50) and a bias strength (say 100), then there will be 50 x 100 = 5000 random locations generated in the entire domain. Note that these 5K locations are uniform in the domain.
2) Then each of these 5K locations are assigned a weight based on the selected weight variable, and they're sorted based on their weight.
3) In the sorted location list, either the first 50 or the last 50 locations are used as seeds to generate flow lines. 
> 
> VAPOR2:
> 
> vapor2_flow
> 
> VAPOR3:
> 
> vapor3_flow

shaomeng:
> 
> VAPOR2:
> 
> VAPOR3:

I used sphere rendering to only render the seeds but not actual stream lines, and then got the following images. It's a Leigh Orf tornado data.

Screenshot from 2021-03-02 17-20-09
Screenshot from 2021-03-02 17-24-05

clyne:
@shaomeng I'd forgotten that the range of the slider could be adjusted by right clicking on it. Once I cranked it up to 100,000 I got results similar to VAPOR2. The good news is that it can be made to work. The bad news is that from a usability standpoint having to fiddle around so much with the slider control is pretty onerous for such a common use case. I'd suggest applying a multiplier so that small changes with the slider have a more visible impact (as opposed to say changing the default maximum value of the slider to 100,000 :-)

shaomeng:
> @shaomeng I'd forgotten that the range of the slider could be adjusted by right clicking on it. Once I cranked it up to 100,000 I got results similar to VAPOR2. The good news is that it can be made to work. The bad news is that from a usability standpoint having to fiddle around so much with the slider control is pretty onerous for such a common use case. I'd suggest applying a multiplier so that small changes with the slider have a more visible impact (as opposed to say changing the default maximum value of the slider to 100,000 :-)

Well, I was thinking that if you know that well where you want your seeds, why don't you use the regular rake or list of seeds mode? 

Nevertheless, I changed the internal knob to be 10X stronger: 2acbac856

clyne:
> > @shaomeng I'd forgotten that the range of the slider could be adjusted by right clicking on it. Once I cranked it up to 100,000 I got results similar to VAPOR2. The good news is that it can be made to work. The bad news is that from a usability standpoint having to fiddle around so much with the slider control is pretty onerous for such a common use case. I'd suggest applying a multiplier so that small changes with the slider have a more visible impact (as opposed to say changing the default maximum value of the slider to 100,000 :-)
> 
> Well, I was thinking that if you know that well where you want your seeds, why don't you use the regular rake or list of seeds mode?

The example I gave you is contrived: there is single structure. There could be multiple regions with extreme values. 

> 
> Nevertheless, I changed the internal knob to be 10X stronger: [2acbac8](https://github.com/NCAR/VAPOR/commit/2acbac85689297de4d8ed7443b360d13b792c126)

Hmm. That doesn't seem like enough since I had to crank up the value to 100,000. Did you try it for yourself and verify you can get results as shown with VAPOR2?

shaomeng:
> 
> Hmm. That doesn't seem like enough since I had to crank up the value to 100,000. Did you try it for yourself and verify you can get results as shown with VAPOR2?

Honestly the image you showed looks like a core in the middle and nothing surrounding it. I'd definitely use a small rake to specify where that core is. And in case of multiple cores in a data set, then I'll use a list of seeds to specify all those cores. 

I imagine the use case of "Random w/ Bias" is to generate images like I showed [here](https://github.com/NCAR/VAPOR/pull/2633#issuecomment-789323382). You're getting more seeds in certain areas but not excluding other areas, as the "Random" word in the name suggests.

That being said, I'm happy to keep cranking up the internal multiplier while keeping the GUI to display a (-100, 100) range. Right now it's varying in a range of (-1000, 1000) internally.

clyne:
> > Hmm. That doesn't seem like enough since I had to crank up the value to 100,000. Did you try it for yourself and verify you can get results as shown with VAPOR2?
> 
> @clyne
> Honestly the image you showed looks like a core in the middle and nothing surrounding it. I'd definitely use a small rake to specify where that core is. And in case of multiple cores in a data set, then I'll use a list of seeds to specify all those cores.
> 

The whole point of these changes is usability (i.e. ease of use). How would you suggest a user identify multiple cores, then get the coordinates for the regions surrounding, and then randomly inject points into those regions? Does that sound easy to you?

> I imagine the use case of "Random w/ Bias" is to generate images like I showed [here](https://github.com/NCAR/VAPOR/pull/2633#issuecomment-789323382). You're getting more seeds in certain areas but not excluding other areas, as the "Random" word in the name suggests.

That's one use case. I've given you another.

> 
> That being said, I'm happy to keep cranking up the internal multiplier while keeping the GUI to display a (-100, 100) range. Right now it's varying in a range of (-1000, 1000) internally.

Keep cranking until you can reproduce the image I showed you.

shaomeng:
@clyne 
It seems a hack to me to skew a "random" process to an extreme degree in order to produces a highly predictable and very well defined outcome.

Commit ec7b01a cranks the internal multiplier to 1000 though, so the combined effect is 100,000 which was reported to be big enough.

clyne:
> @clyne
> It seems a hack to me to skew a "random" process to an extreme degree in order to produce a highly predictable and very well defined outcome.
> 

I think you are assuming that all random process are uniform. They are not. See https://en.wikipedia.org/wiki/Probability_distribution#Common_probability_distributions_and_their_applications


> Commit [ec7b01a](https://github.com/NCAR/VAPOR/commit/ec7b01a26b22d79421b07fd1df4c96cc9700dbb3) cranks the internal multiplier to 1000 though, so the combined effect is 100,000 which was reported to be big enough.

shaomeng:
> > @clyne
> > It seems a hack to me to skew a "random" process to an extreme degree in order to produce a highly predictable and very well defined outcome.
> 
> I think you are assuming that all random process are uniform. They are not. See https://en.wikipedia.org/wiki/Probability_distribution#Common_probability_distributions_and_their_applications
> 
I guess I meant that if you want a step function, then use a step function. If you want a step function, but use a probability process to simulate, and insist tweaking the parameters so the probability process looks quite sharp on the edge but still not as sharp as a step function, well, it's a hack after all and why not simply use a step function?

clyne:
Think of it this way: we want a smooth, parameterized probability function that maps field magnitude to probability that has the property that it can vary smoothly from a uniform distribution to a dirac delta function (a spike). I.e. if parameterized by lambda, a value of 0.0 gives a uniform distribution, and lamdba=1.0 gives a spike at the extreme value of the data. If it were me I would consider modeling this as a half gaussian with the gaussian "mean" at the extreme magnitude and the tail of the gaussian over the min, and lambda proportional to sigma. (if lamba is negative, reverse min and max).

shaomeng:
> Think of it this way: we want a smooth, parameterized probability function that maps field magnitude to probability that has the property that it can vary smoothly from a uniform distribution to a dirac delta function (a spike). I.e. if parameterized by lambda, a value of 0.0 gives a uniform distribution, and lambda=1.0 gives a spike at the extreme value of the data (the use case I identified above needs lambda=1.0, but other use cases will want different values). If it were me I would consider modeling this as a half gaussian with the gaussian "mean" at the extreme magnitude and the tail of the gaussian over the min, and lambda proportional to sigma. (if lambda is negative, reverse min and max).

I see. I think your use case is valid, and the process you described shouldn't have "random" in it's name. Something like "Probability Process" sounds more appropriate to me ;)

clyne:
From wikipedia: "A random variable has a probability distribution, which specifies the probability of Borel subsets of its range". Later, "The realizations of a random variable, that is, the results of randomly choosing values according to the variable's probability distribution function, are called random variates." 

I think "random" is indeed the correct terminology :-)

shaomeng:
That's interesting. In that case, let's keep the random name there!

clyne (approved):

sgpearse (changes requested):
My one concern is that I get an error when I enable Flow, then change the distribution type to "Random w/ Bias".  Once I select a bias variable, things work as expected.  I'm not sure if it makes sense to have the "none" option when selecting the bias variable.

Minor nit - can we change the bias variable from "RakeBiasVariable" to "Rake Bias Variable"?

clyne:
> My one concern is that I get an error when I enable Flow, then change the distribution type to "Random w/ Bias". Once I select a bias variable, things work as expected. I'm not sure if it makes sense to have the "none" option when selecting the bias variable.
> 
> Minor nit - can we change the bias variable from "RakeBiasVariable" to "Rake Bias Variable"?

Agreed we should avoid the error message (e.g. either pick a valid default variable, or not try to open the variable when it's "none").

shaomeng:
Hi @StasJ , a quick question about the variable selector: 

Some instances of the variable selector, for example the color mapping variable selector, automatically chooses the 1st variable in the variable list to be it's value. Is it possible to achieve the same for a generic variable selector that's created by calling `new PVariableSelector()`?

Thanks!

StasJ:
> Hi @StasJ , a quick question about the variable selector:
> 
> Some instances of the variable selector, for example the color mapping variable selector, automatically chooses the 1st variable in the variable list to be it's value. Is it possible to achieve the same for a generic variable selector that's created by calling `new PVariableSelector()`?
> 
> Thanks!

@shaomeng no, that is handled by the params rather than the variable selector GUI.

shaomeng:
> @shaomeng no, that is handled by the params rather than the variable selector GUI.

I got what you mean, thanks!

shaomeng:
@sgpearse your concern is addressed in 6dd7d84.

sgpearse (approved):

PR #2634, opened 2021-03-04
Fix #2626 by having gitHook only push added files

### LINKED ISSUE #2626: Pre-Commit hook adds unrelated files to commit
1. Modify files A and B
2. Commit file A
3. Push

Notice both files A and B have been pushed.

sgpearse:
Fix #2626 by having gitHook only push added files


StasJ (approved):
This fixes the original issue and can be accepted, however, we still run into #2635

Issue #2197, opened 2020-01-31
Slice: Render along "logical" plane


sgpearse:
With VDC data, sometimes our compressed data will have a Z dimensionality of 1 for 3D variables.  This causes the Slice renderer to miss during its sampling, unless the sample lies perfectly where coordinate that the data exists.

This feature would allow the Slice renderer to sample the logical data in a user-defined dimension.

clyne:
Moreover, there are times when users want to see data along a slice of the grid, not a mathematical plane.

Xenotech81:
Hello,
in wind energy we are often interested in a "**terrain following slice**", for example wind speed magnitude at a constant vertical offset from the ground (usually at hub height).

Also, making the regular slices rotatable about the z axis (#2637) makes absolutely sense, as we usually want to slice along the local flow direction upstream of the wind turbine, which is not necessarily aligned with any of the geometrical axes.

PR #2641, opened 2021-03-12
Fix smokeTest artifact directory

sgpearse:
smokeTest results were not being posted onto CircleCI's artifact listing.  The tests were being run, but results were not being posted.


clyne:
@sgpearse the driver script, smokeTests.py, would be much more helpful if it echoed all of the commands that it invoked so that you can easily identify and rerun the executable with the appropriate arguments.

sgpearse:
@clyne This PR fixes a bug that was preventing our tests on different grid shapes (1x1x1, 8x8x1, etc).  The test for 8x8x1 is currently failing due to an assertion in LayeredGrid.cpp:160, shown below.  I took a look at it but I'm not sure there is an obvious fix.  I'm not sure if this relates to #2197.

`VAssert(indices[2] < nz - 1) failed in LayeredGrid.cpp, line 160`

clyne:
@sgpearse will remove problematic tests (dimension length = 1), convert PR from draft to read-for-review, and open a separate issue to address problem cases, which may be already fixed by #2439

test_apps/smokeTests/testResults/cf_baseline.txt
```diff
@@ -60,4 +60,4 @@ Time Coordinates:
 
 Time coordinate variable name: time
 Number of time steps: 1
-Elapsed time: 0.131529
+Elapsed time: 0.318985
```
clyne, line 63: Should all of these *baseline* files contain the run time? Aren't they used for  comparison to test for regressions? You could add an option to disable them. I added an execution timer because I wanted to do performance evaluation, but I wasn't aware that the outputs were being used in this way.

sgpearse: The final line in the baseline files that shows "Elapsed time" does not get compared in the diff.  I still need to scrub the new "Time" entries that you've added, but I think they're currently benign as all of our tests report a value of 0.  Still, these should be scrubbed in case they ever report non-zero.

PR #2642, opened 2021-03-12
Time annotation coordinates should be normalized #1250

### LINKED ISSUE #1250: Time annotation coordinates should be normalized
Screen space Coordinates for placement of Time Annotation (under Navigation/Annotation) are specified in an unknown coordinate system. They should be in normalized coordinates (0..1.0) that get mapped into the viewport coordinates as is done with color bars.

StasJ:
Fix #1250

### DIFF
--- a/apps/vaporgui/AnnotationEventRouter.cpp
+++ b/apps/vaporgui/AnnotationEventRouter.cpp
@@ -93,6 +93,16 @@ AnnotationEventRouter::AnnotationEventRouter(QWidget *parent, ControlExec *ce) :
         })
     });
     layout()->addWidget(_axisArrowGroup);
+    
+    
+    _timeSlidersGroup = new PGroup({
+        new PLabel("Lower-left coordinates:"),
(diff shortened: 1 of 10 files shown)


clyne (commented):
We're using sliders to position color bars, and and text entry boxes for time annotation. Shouldn't we use sliders for both to be consistent? Also, since the coordinates are normalized from 0.0 to 1.0, and values outside of that range probably don't make any sense, a slider with limits 0.0 to 1.0 would help prevent the user from fumbling around with nonsense values.

StasJ:
> We're using sliders to position color bars, and and text entry boxes for time annotation. Shouldn't we use sliders for both to be consistent? Also, since the coordinates are normalized from 0.0 to 1.0, and values outside of that range probably don't make any sense, a slider with limits 0.0 to 1.0 would help prevent the user from fumbling around with nonsense values.

I agree but I did not think that was within the scope of the issue, especially since that entire GUI needs to be replaced.

I've replaced the widgets with sliders going from 0-1.

clyne (approved):
Awesome!

sgpearse (approved):

PR #2662, opened 2021-03-22
Remove "~" references to home-directory in SettingsParams

### LINKED ISSUE #2657: Default path for auto session saving invalid
The default path used for automatically saving session state appears to be invalid on the Mac, and maybe elsewhere. It is currently set to "~/VaporAutoSave.vs3" and I do not believe that the "~" is being expanded to the user's home directory.

### To Reproduce

1. Load any data set
2. Open the preferences menu and click "restore defaults"
3. Remove any existing auto session file (~/VaporAutoSave.vs3)
4. Interact with the application (change view points, etc.) A new auto save file should be written, but it is not.
5. Use the file browser to select a valid file path. VAPOR will immediately begin saving state to the file selected via the file browser.

### Desktop
Mac OSX 10.15

### LINKED ISSUE #2676: SettingsParams: ~ / homeDir replacement problem
SettingsParams tries to scrub the "~" character from file paths, and replace it with the home directory.  However it only does so if the file path is only the single character, "~".  

The file path "~/foo" will not have ~ replaced with the home directory.

sgpearse:
Fixes #2657 by removing "~" references to the user's home directory, and adding error checking for file-writes in MainForm.

The ofstream method is picky about the file path its given.   While it writes to existing files with "~" in its path, it will not create them.

Note: You may need to "restore defaults" to overwrite an existing .vapor3_prefs file in your home directory, which still may refer to "home" as "~"

Also fixes #2676

### DIFF
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -1930,9 +1930,7 @@ void MainForm::_performSessionAutoSave()
     if (_eventsSinceLastSave >= eventCountForAutoSave) {
         string autoSaveFile = sParams->GetAutoSaveSessionFile();
         int    rc = _paramsMgr->SaveToFile(autoSaveFile);
-        if (rc < 0) { 
-            MSG_ERR( "Unable to write settings file " + autoSaveFile );
-        }
+        if (rc < 0) { MSG_ERR("Unable to write settings file " + autoSaveFile); }
         _eventsSinceLastSave = 0;
     }
 }
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -1930,7 +1930,9 @@ void MainForm::_performSessionAutoSave()
     if (_eventsSinceLastSave >= eventCountForAutoSave) {
         string autoSaveFile = sParams->GetAutoSaveSessionFile();
         int    rc = _paramsMgr->SaveToFile(autoSaveFile);
-        if (rc < 0) { MSG_ERR(); }
+        if (rc < 0) { 
+            MSG_ERR( "Unable to write settings file " + autoSaveFile );
+        }
         _eventsSinceLastSave = 0;
     }
 }
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -1929,12 +1929,11 @@ void MainForm::_performSessionAutoSave()
 
     if (_eventsSinceLastSave >= eventCountForAutoSave) {
         string autoSaveFile = sParams->GetAutoSaveSessionFile();
-        int rc = _paramsMgr->SaveToFile(autoSaveFile);
-        if ( rc == -1 ) {
-            MSG_ERR( "Unable to open settings file " + autoSaveFile );
-        }
-        else if ( rc < -1 ) {
-            MSG_ERR( "Unable to write settings file " + autoSaveFile );
+        int    rc = _paramsMgr->SaveToFile(autoSaveFile);
+        if (rc == -1) {
+            MSG_ERR("Unable to open settings file " + autoSaveFile);
+        } else if (rc < -1) {
+            MSG_ERR("Unable to write settings file " + autoSaveFile);
         }
         _eventsSinceLastSave = 0;
     }
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -1930,9 +1930,7 @@ void MainForm::_performSessionAutoSave()
     if (_eventsSinceLastSave >= eventCountForAutoSave) {
         string autoSaveFile = sParams->GetAutoSaveSessionFile();
         int    rc = _paramsMgr->SaveToFile(autoSaveFile);
-        if (rc < 0) {
-            MSG_ERR();
-        } 
+        if (rc < 0) { MSG_ERR(); }
         _eventsSinceLastSave = 0;
     }
 }
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -1929,7 +1929,13 @@ void MainForm::_performSessionAutoSave()
 
     if (_eventsSinceLastSave >= eventCountForAutoSave) {
         string autoSaveFile = sParams->GetAutoSaveSessionFile();
-        _paramsMgr->SaveToFile(autoSaveFile);
+        int rc = _paramsMgr->SaveToFile(autoSaveFile);
+        if ( rc == -1 ) {
+            MSG_ERR( "Unable to open settings file " + autoSaveFile );
+        }
+        else if ( rc < -1 ) {
+            MSG_ERR( "Unable to write settings file " + autoSaveFile );
+        }
         _eventsSinceLastSave = 0;
     }
 }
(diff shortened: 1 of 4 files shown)


apps/vaporgui/MainForm.cpp
```diff
@@ -1929,7 +1929,12 @@ void MainForm::_performSessionAutoSave()
 
     if (_eventsSinceLastSave >= eventCountForAutoSave) {
         string autoSaveFile = sParams->GetAutoSaveSessionFile();
-        _paramsMgr->SaveToFile(autoSaveFile);
+        int    rc = _paramsMgr->SaveToFile(autoSaveFile);
+        if (rc == -1) {
+            MSG_ERR("Unable to open settings file " + autoSaveFile);
```
StasJ, line 1934: Wouldn’t this be the session file rather than the settings file? Also we may want to make this a warning that can be disabled rather than an error.

apps/vaporgui/SettingsParams.cpp
```diff
@@ -434,7 +434,8 @@ void SettingsParams::Init()
 {
     SetSessionAutoSaveEnabled(true);
     SetChangesPerAutoSave(5);
-    SetAutoSaveSessionFile("~/VaporAutoSave.vs3");
+    std::string homeDir = QDir::homePath().toStdString();
+    SetAutoSaveSessionFile(homeDir + "/VaporAutoSave.vs3");
```
StasJ, line 438: This may be outside the scope of this PR, but I think it is customary to make this a hidden file and provide a menu option to restore it rather than placing a file in the user’s home directory.

StasJ (commented):
Also I think the proper solution would be to resolve the tilde when accessing the file rather than replacing the default with a hard path:
- a user may manually enter a path relative to their home directory
- A user may share a session file with someone else and vapor will try to use the original user’s config which may be undesirable, if it is even accessible.

apps/vaporgui/MainForm.cpp
sgpearse: You're correct.  I removed all directory references (not just the settings) that used "~" because ofstream will not create a new file without a full file path.

sgpearse:
@stasj If you check out SettingsParams, the ~ is still handled in the getters and setters, so the user can still use a ~.  But this is moot because your file selectors do not allow for manual path entry.

apps/vaporgui/SettingsParams.cpp
clyne: Agreed, but people are already accustomed to it not being a hidden file. Changing it won't get us much and may cause some confusion.

clyne:
> Also I think the proper solution would be to resolve the tilde when accessing the file rather than replacing the default with a hard path:
> 
> * a user may manually enter a path relative to their home directory
> * A user may share a session file with someone else and vapor will try to use the original user’s config which may be undesirable, if it is even accessible.

@stasj please suggest the code changes to support this suggestion.

StasJ:
> > Also I think the proper solution would be to resolve the tilde when accessing the file rather than replacing the default with a hard path:
> > 
> > * a user may manually enter a path relative to their home directory
> > * A user may share a session file with someone else and vapor will try to use the original user’s config which may be undesirable, if it is even accessible.
> 
> @StasJ please suggest the code changes to support this suggestion.

Doing the above would require refactoring either the getters to replace tildes or refactoring our IO calls to handle them. That is assuming that our current IO calls cannot handle them, however, I still cannot reproduce #2657.

clyne:
> > > Also I think the proper solution would be to resolve the tilde when accessing the file rather than replacing the default with a hard path:
> > > 
> > > * a user may manually enter a path relative to their home directory
> > > * A user may share a session file with someone else and vapor will try to use the original user’s config which may be undesirable, if it is even accessible.
> > 
> > 
> > @StasJ please suggest the code changes to support this suggestion.
> 
> Doing the above would require refactoring either the getters to replace tildes or refactoring our IO calls to handle them. That is assuming that our current IO calls cannot handle them, however, I still cannot reproduce #2657.

@stasj I can no longer reproduce this either, but the branch has been updated with a fix that is SettingsParams::Init(). Wasn't your argument that a more robust fix would be to handle the "~" in the getter? With the changes currently proposed don't we still have the problem that if a "~" finds its way into the database it may not be handled by whoever is doing I/O?

clyne:
@stasj, the bug can still be reproduced if you have an existing .vapor3_settings file that has the old AutoSaveFileLocation default value of "~/VaporAutoSave.vs3".

However, in looking into this I noted that every single getter in SettingsParams that returns a path contains:

`if (dir == "~") { dir = QDir::homePath().toStdString(); }`

except for GetAutoSaveSessionFile(). I suggest we change GetAutoSaveSessionFile() to be consistent with the other getters and be done with this.

sgpearse:
> I suggest we change GetAutoSaveSessionFile() to be consistent with the other getters and be done with this.

@clyne I've added added logic that does this, but the way it was being done previously (shown below) was problematic.

`if (dir == "~") { dir = QDir::homePath().toStdString(); }`

This misses the case where dir=="~/foo".  Instead, I'm replacing all occurrences of ~ with the home directory.  For consistency, this logic has been given its own function, and is now being used by all Getters in SettingsParams instead of the code shown above.

apps/vaporgui/SettingsParams.cpp
```diff
@@ -139,6 +139,15 @@ void SettingsParams::Reinit() { Init(); }
 
 SettingsParams::~SettingsParams() {}
 
+void SettingsParams::_swapTildeWithHome(std::string &file) const
+{
+    size_t pos = 0;
+    while ((pos = file.find("~", pos)) != std::string::npos) {
+        file.replace(pos, 1, QDir::homePath().toStdString());
+        pos += 1;
```
clyne, line 147: This will work, but shouldn't 'pos' be incremented by QDir::homePath().toStdString().length()?  Not likely, but if QDir:: homePath() returns an empty string I'm not sure what would happen.

apps/vaporgui/MainForm.cpp
```diff
@@ -1929,7 +1929,12 @@ void MainForm::_performSessionAutoSave()
 
     if (_eventsSinceLastSave >= eventCountForAutoSave) {
         string autoSaveFile = sParams->GetAutoSaveSessionFile();
-        _paramsMgr->SaveToFile(autoSaveFile);
+        int    rc = _paramsMgr->SaveToFile(autoSaveFile);
+        if (rc == -1) {
```
clyne, line 1933: The error handling here seems redundant and more complex than needed. ParamsMgr::SaveToFile() already distinguishes between a file write vs file access error and prints an appropriate message for each case. The message generated here replicates the message and doesn't add any additional information. Why not simply leave SaveToFile() unchanged (returning a single error code), and not have the code here detect between the two error conditions and generate what is essentially the same message?

clyne (commented):
Getting close. Just a couple of in-line comments.

apps/vaporgui/MainForm.cpp
clyne: MSG_ERR will include any messages that have been reported by MyBase::SetErrMsg

apps/vaporgui/SettingsParams.cpp
sgpearse: Infinite loop problem understood.  Increment changed to 

`pos += QDir::homePath().toStdString().length();`

apps/vaporgui/MainForm.cpp
sgpearse: Got it.  Fixed.

StasJ (approved):

sgpearse:
@clyne this should be all set.

clyne (approved):

PR #2663, opened 2021-03-22
Remove window size locking from Vapor's "Settings"

### LINKED ISSUE #2656: Window size feature should apply to visualization window, not entire application
There is a setting on the Preferences menu to set (and lock) the window size. This is particularly useful for people who want to capture a number of outputs images, often over numerous session, and want the images to be a consistent size. However, the size specification applies to the overall application window, not the renderer window, limiting the usability of the feature.

sgpearse:
Removes window size locking from the GUI and SettingsParams database.  Fixes #2656.


clyne:
I requested @stasj review this, but let's discuss and make sure this is really what we want.

StasJ (approved):

clyne (approved):

PR #2664, opened 2021-03-23
Disable dynamic update on flow integration steps

### LINKED ISSUE #2658: text input box commits changes before "enter" is pressed.
### Describe the bug
At least one text input box, namely the one that controls "Integration Steps" in the Flow renderer, accepts and applies changes before the "enter" key is pressed by the user. 

### Expected behavior
When a user inputs a value to this particular input box, and really most of the text input boxes in vapor, it is supposed to only *display* the value, and only *commits* the change when the user hits "enter" key. The benefit will be:
1. For expensive operations, VAPOR doesn't perform the expensive operation prematurely, i.e., before the user finishes typing.
2. It will result in one "undo/redo" step instead of many.

### To Reproduce
1. Open VAPOR, create a flow renderer.
2. Enable the flow renderer. Start typing a new "Integration Step" and notice that the rendering changes with every key stroke before you hit "enter."

sgpearse:
Fixes #2658 by disabling dynamic updates to the Flow Integration Steps spinbox.

### DIFF
--- a/apps/vaporgui/FlowEventRouter.cpp
+++ b/apps/vaporgui/FlowEventRouter.cpp
@@ -32,7 +32,7 @@ FlowEventRouter::FlowEventRouter(QWidget *parent, ControlExec *ce) : RenderEvent
             new PEnumDropdown(FP::_isSteadyTag, {"Streamlines", "Pathlines"}, {true, false}, "Flow Type"),
             (new PShowIf(FP::_isSteadyTag))->Equals(true)->Then({
                 new PEnumDropdown(FP::_flowDirectionTag, {"Forward", "Backward", "Bi-Directional"}, {(int)FlowDir::FORWARD, (int)FlowDir::BACKWARD, (int)FlowDir::BI_DIR}, "Flow Direction"),
-                (new PIntegerInput(FP::_steadyNumOfStepsTag, "Integration Steps"))->SetRange(0, 10000)->EnableDynamicUpdate(),
+                (new PIntegerInput(FP::_steadyNumOfStepsTag, "Integration Steps"))->SetRange(0, 10000),
             })->Else({
                 _pathlineLengthSlider = new PIntegerSliderEdit(FP::_pastNumOfTimeSteps, "Pathline Length"),
                 _pathlineInjectionSlider = new PIntegerSliderEdit(FP::_seedInjInterval, "Injection Interval"),


StasJ (approved):

shaomeng (commented):
Well, I think the "press enter to commit" model should *only* apply to text boxes. I.e., when user clicks on the up/down arrow, the change is immediately committed. 

This way of operation would be consistent with other widgets, for example, the slider+text box combo. In that widget, when you drag the slider, the change is immediately committed, while when you type a number in the text box, the changes are committed when "enter" is hit. 

Note, if such behavior is impossible on this particular widget, then this PR addresses the original concern.

sgpearse:
> when you drag the slider, the change is immediately committed, while when you type a number in the text box, the changes are committed when "enter" is hit.

This isn't true for all sliders.  Some have intermediate changes enabled, and others do not.  

Previously, this spinbox allowed for intermediate changes.  This PR only applies parameter changes once enter is pressed, or the spinbox loses context.

StasJ:
> > when you drag the slider, the change is immediately committed, while when you type a number in the text box, the changes are committed when "enter" is hit.
> 
> This isn't true for all sliders. Some have intermediate changes enabled, and others do not.
> 
> Previously, this spinbox allowed for intermediate changes. This PR only applies parameter changes once enter is pressed, or the spinbox loses context.

I agree with Sam that the widget should emit a change when a user clicks the arrows.

shaomeng:
Yeah I think this would be a desirable behavior, but if it's hard to
implement, then I can live with a click on arrow + press enter 2-step
process.

clyne:
@StasJ , @shaomeng I'm a little confused here. My understanding is that the spin box supports two modes: continuous update and update on completion. According to @sgpearse the latter occurs "once enter is pressed, or the spinbox loses context." It would seem that for an expensive event you want the latter behavior. Stas, and Sam, are you advocating for using the first mode (continuous update), or the second mode, but changing the definition of the second mode (i.e. having some other event trigger update completion)?

shaomeng:
> @StasJ , @shaomeng I'm a little confused here. My understanding is that the spin box supports two modes: continuous update and update on completion. According to @sgpearse the latter occurs "once enter is pressed, or the spinbox loses context." It would seem that for an expensive event you want the latter behavior. Stas, and Sam, are you advocating for using the first mode (continuous update), or the second mode, but changing the definition of the second mode (i.e. having some other event trigger update completion)?

I'm thinking that a spin box has 2 ways to interact with a user. 1) to manually type in a value, and 2) to use a motion to change its value, which is to click up/down arrows. The widget should respond differently to these 2 types of interactions. W.r.t 1), it should commit changes when a user hits "enter." W.r.t 2), it should commit changes whenever the arrows are clicked. 

Think about the slider + text box combo that we have in other parts of the GUI. I'm advocating that the spin box behaves consistently with the slider + text box combo.

StasJ:
> @StasJ , @shaomeng I'm a little confused here. My understanding is that the spin box supports two modes: continuous update and update on completion. According to @sgpearse the latter occurs "once enter is pressed, or the spinbox loses context." It would seem that for an expensive event you want the latter behavior. Stas, and Sam, are you advocating for using the first mode (continuous update), or the second mode, but changing the definition of the second mode (i.e. having some other event trigger update completion)?

I think the correct behavior would be to update when the user finishes entering text or when they click an arrow.

clyne:
> > @StasJ , @shaomeng I'm a little confused here. My understanding is that the spin box supports two modes: continuous update and update on completion. According to @sgpearse the latter occurs "once enter is pressed, or the spinbox loses context." It would seem that for an expensive event you want the latter behavior. Stas, and Sam, are you advocating for using the first mode (continuous update), or the second mode, but changing the definition of the second mode (i.e. having some other event trigger update completion)?
> 
> I think the correct behavior would be to update when the user finishes entering text or when they click an arrow.

@StasJ , @shaomeng, sorry but I'm still not sure that we're all on the same page.

I *think* we're all in agreement that the text+slider combo should behave consistently with text+spinbox. I.e. the text box portion of the combos should behave identically, as should the slider or spinbox portion of the combo. In the latter case dragging a slider maps to holding a spinbox arrow, and releasing (unclicking) a slider maps to releasing the spinbox arrow. 

Agreed?

At present, it does not look like the text boxes are consistent with these two combos. With the release candidate, changes made to the text box in a text+spinbox combo are made immediately (before hitting enter), while changes made to the text box in a text+slider combo are not triggered until enter is hit. This is what I've observed with the "Integration Steps" and "{X,Y,Z} axis seed" controls on the flow tab. These should be changed to be made consistent; an event should not be generated until enter is hit or focus leaves the text box.

Agreed?

Both combos support a EnableDynamicUpdate() configuration method. When this is set spinbox/slider events are generated immediately (whenever the user drags the slider or holds down the arrows of the spinbox). When it is not set, events are not generated until interaction is completed (releasing the slider or arrow as the case may be).

Agreed?

Because changing Integration Steps on the Flow tab is a potentially expensive event we do not want Dynamic Updates enabled. 

Agreed?

shaomeng:
maybe it's easier to make sure we're on the same page when we meet this
afternoon. But one thing I think is worth making clear is that the
terminology "spinbox" is a single widget that QT provides; it resembles a
combo of textbox+arrows.
https://doc.qt.io/qt-5/qtwidgets-widgets-spinboxes-example.html

On the other hand the combo of a textbox+slider is something VAPOR
invented.

clyne:
@stasj, please see if there is a better choice of Qt event option for the spinbox (one that triggers when the arrow is released).

sgpearse:
The QSpinBox provides two signals:

```
void textChanged(const QString &text)
void valueChanged(int i)
```

https://doc.qt.io/qt-5/qspinbox.html#signals

shaomeng (approved):

PR #2671, opened 2021-03-24
Set default num threads to 0

### LINKED ISSUE #2653: Default setting for number of threads should be zero
The default setting for the number of execution threads is currently set to 4, which is probably not an ideal choice. If set to zero the application will try to use as many threads as available to it as returned by sysconf(_SC_NPROCESSORS_ONLN).

To reproduce:
1 Open the preferences menu.
2. Click "restore default". Observe the number of threads.

StasJ:
Fix #2653


clyne (commented):
@stasj this certainly addresses the problem, but how difficult would it be to add a checkbox to the preferences menu that says something like "set the number of threads to the number of cores", and if not checked allowing the user to specify a number?

StasJ:
> @StasJ this certainly addresses the problem, but how difficult would it be to add a checkbox to the preferences menu that says something like "set the number of threads to the number of cores", and if not checked allowing the user to specify a number?

Done

clyne (approved):

clyne:
@StasJ please resolve check failures and I will merge. Thanks.

StasJ:
> @StasJ please resolve check failures and I will merge. Thanks.

Done

PR #2682, opened 2021-03-30
Issue 2678 : The VAPOR cache size is unintuitive and should be made unlimited by default

### LINKED ISSUE #2678: The VAPOR cache size is unintuitive and should be made unlimited by default
Similar argument to the default number of CPU cores.

clyne:
Fixes #2678

### DIFF
--- a/apps/vaporgui/SettingsParams.cpp
+++ b/apps/vaporgui/SettingsParams.cpp
@@ -141,14 +141,14 @@ SettingsParams::~SettingsParams() {}
 
 long SettingsParams::GetCacheMB() const
 {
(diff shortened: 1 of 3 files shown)


shaomeng:
@clyne  It seems this PR handles the dealing of `0`, but doesn't set the default value to be 0. I.e., when clicking on that "Restore Default" button, the value goes back to 8000.

clyne:
Good catch, @shaomeng . Fixed.

shaomeng (approved):

StasJ (approved):

PR #2683, opened 2021-03-31
Issue2645 Fix poor barb performance with large datasets

### LINKED ISSUE #2645: Performance of Barbs unusable with larger data
### Describe the bug
The Barb renderer is unusable with larger data. It appears to be performing unnecessary calculations with every update, including scene navigation.

### To Reproduce
1. Load PALM/cbl_KIT_ug00_dx10_masked_M01.000.0119.split.nc from glade
2. Enable the barb renderer
3. Try to change the viewpoint. It's impossible to do this interactively.

StasJ:
Fixes the problem by caching the generated barbs.

Fix #2645

### DIFF
--- a/include/vapor/BarbRenderer.h
+++ b/include/vapor/BarbRenderer.h
@@ -71,7 +71,7 @@ private:
 
     //! \copydoc Renderer::_paintGL()
     virtual int _paintGL(bool fast);
-    
+
     int _generateBarbs();
 
     int _getVectorVarGrids(int ts, int refLevel, int lod, std::vector<double> minExts, std::vector<double> maxExts, std::vector<VAPoR::Grid *> &varData);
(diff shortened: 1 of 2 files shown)


sgpearse:
@stasj could we please remove the commented code in BarbRenderer::_isCacheDirty()?

sgpearse:
> That code was left intentionally but if you insist.

I don't want to fuss, but this is in our [convention doc](https://github.com/NCAR/VAPOR/wiki/Notions-And-Actions-on-Code-Quality).  We can change the doc too.  I'd just like to be on the same page if we use a standard.

clyne (approved):
Nice work! But please remove commented out code.

shaomeng (approved):

sgpearse (approved):
Thank you stas.

PR #2695, opened 2021-04-05
Issue 2694 DataMgrUtils::GetGrid needs refactoring to avoid oversubscribed cache

### LINKED ISSUE #2694: DataMgrUtils::GetGrid needs refactoring to avoid oversubscribed cache
The DataMgrUtils::GetGrid() convenience function does not support a "lock" option as does DataMgr::GetVariable(); all grids returned by GetGrids() are locked and have to be explicitly freed using DataMgrUtils::UnlockGrid(). Since the most typical use case is to not required locked grids, this often leads to oversubscribing the cache, and developers forget to call UnlockGrid(). GetGrids() should be refactored to include a lock option. 

Relatedly, DataMgr::UnlockGrid() has special case handling for each grid type (and is currently missing UnstructuredGrids). Thus, coordinate arrays are often not freed. Also, StretchedGrids are another problematic case because this class copies coordinate data to internal memory and does not retain a pointer to the cached coordinate values. 

Other problems noted: 

- The lock parameter is not mentioned in the documentation for any of the GetVariable or GetGrids methods
- The default value for the lock parameter is false for all of the GetVariable methods
- The default value for the lock parameter is true for all of the GetGrids methods and this is hidden from the interface
- It should be possible to have the destructor of Grid automatically reduce its reference count, solving this issue

Relates to #2686

clyne:
Fixes #2694 

This PR does the following:

1. A "lock" flag has been added (and documented) to the  DataMgrUtils::GetGrids(), making it consistent with the DataMgr::GetVariable method. A default value of False is set. I.e. Grid(s) returned by GetGrids are unlocked by default. N.B. if multiple Grids are fetched by GetGrids() they are locked internally to ensure all the requested Grids can be returned. If the lock flag is not set, the Grids are unlocked prior to returning from GetGrids().
2. Removes the now superfluous explicit calls to DataMgrUtils::UnlockGrids() that were found throughout the application. 
3. Addresses an unreported memory leak in the DataMgr::UnlockGrid() that resulted in some of the data held by a Grid not being freed.

### DIFF
--- a/include/vapor/DataMgr.h
+++ b/include/vapor/DataMgr.h
@@ -683,8 +683,8 @@ private:
 
     std::map<string, BlkExts> _blkExtsCache;
 
-    std::map <const Grid *, vector <float *> > _lockedFloatBlks;
-    std::map <const Grid *, vector <int *> > _lockedIntBlks;
+    std::map<const Grid *, vector<float *>> _lockedFloatBlks;
+    std::map<const Grid *, vector<int *>>   _lockedIntBlks;
 
     // Get the immediate variable dependencies of a variable
     //
--- a/include/vapor/DataMgr.h
+++ b/include/vapor/DataMgr.h
@@ -683,6 +683,9 @@ private:
 
     std::map<string, BlkExts> _blkExtsCache;
 
+    std::map <const Grid *, vector <float *> > _lockedFloatBlks;
+    std::map <const Grid *, vector <int *> > _lockedIntBlks;
+
     // Get the immediate variable dependencies of a variable
     //
     std::vector<string> _get_var_dependencies_1(string varname) const;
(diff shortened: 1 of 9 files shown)


shaomeng:
My view on this attempt to perform memory management inside of VAPOR is summarized in #2700. Given that there were severe bugs in this area not detected for years, one thing I'm certain is that cache size management is something pretty rarely used.

clyne:
> My view on this attempt to perform memory management inside of VAPOR is summarized in #2700. Given that there were severe bugs in this area not detected for years, one thing I'm certain is that cache size management is something pretty rarely used.

I don't see any input ("summary") from you in #2700....

shaomeng:
> > My view on this attempt to perform memory management inside of VAPOR is summarized in #2700. Given that there were severe bugs in this area not detected for years, one thing I'm certain is that cache size management is something pretty rarely used.
> 
> I don't see any input ("summary") from you in #2700....

My bad, it should be #2702 ...

StasJ (approved):

shaomeng (approved):

PR #2699, opened 2021-04-09
Bookmarks

### LINKED ISSUE #2607: Add support for bookmarking
Allow the user to "book mark" the current state of the application in a manner similar to web browser bookmarks. The user could then quickly toggle between different states. This might be particularly helpful when, for example, editing a colormap. The book mark browser could contain a snapshot of the visualizer window along with optional text

StasJ:
Adds support for bookmarking as described in #2607.

The current snapshot size is the maximum natively supported by MacOS menu bars.

s

s2

Fix #2607

### DIFF
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -913,14 +913,14 @@ void MainForm::_createEditMenu()
 
     connect(_editUndoAction, SIGNAL(triggered()), this, SLOT(undo()));
     connect(_editRedoAction, SIGNAL(triggered()), this, SLOT(redo()));
-    
-    
+
+
     _Edit->addSeparator();
     _Edit->addAction("Create Bookmark", this, &MainForm::createBookmark);
-    
+
     _loadBookmarkMenu = new QMenu("Load Bookmark");
     _Edit->addMenu(_loadBookmarkMenu);
-    
+
     _deleteBookmarkMenu = new QMenu("Delete Bookmark");
     _Edit->addMenu(_deleteBookmarkMenu);
 }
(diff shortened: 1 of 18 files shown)


apps/vaporgui/MainForm.cpp
```diff
@@ -1060,6 +1074,141 @@ void MainForm::_createDeveloperMenu()
...
+    for (auto &b : bookmarks) p->AddBookmark(&b);
+
+    string           activeVizWin = p->GetActiveVizName();
+    ViewpointParams *vpp = _paramsMgr->GetViewpointParams(activeVizWin);
+
+    bool useCustomViewport = vpp->GetValueLong(vpp->UseCustomFramebufferTag, 0);
+    int  customViewportWidth = vpp->GetValueLong(vpp->CustomFramebufferWidthTag, 0);
+    int  customViewportHeight = vpp->GetValueLong(vpp->CustomFramebufferHeightTag, 0);
+
+    int           iconSize = 19;
```
clyne, line 1106: What does this encoded string do? Needs some documentation.

include/vapor/Base16StringStream.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include <streambuf>
+#include <ostream>
+#include <istream>
+#include <memory>
+
+
+
```
clyne, line 9: Please document this public class.

lib/render/Visualizer.cpp
```diff
@@ -491,28 +491,39 @@ RegionParams *Visualizer::getActiveRegionParams() const { return _paramsMgr->Get
...
     if (!_getPixelData(framebuffer))
         ;    // goto captureImageEnd;
 
+    if (STLUtils::BeginsWith(path, ":RAM:")) {
+        void *ptr;
+        sscanf(path.c_str(), ":RAM:%p", &ptr);
+
+        memcpy(ptr, framebuffer, sizeof(*framebuffer) * 3 * width * height);
+
+        delete[] framebuffer;
```
clyne, line 519: Any reason to not use a C++ string and avoid the explicit memory management? There is already a goto commented out above that if were uncommented would result in a memory leak.

clyne (commented):
Stas, this looks great! Can we make the icons a little bigger? Pretty hard to see. thanks!

apps/vaporgui/MainForm.cpp
StasJ: Done.

include/vapor/Base16StringStream.h
StasJ: Done.

lib/render/Visualizer.cpp
StasJ: I don't see how uncommenting the goto statement would result in a memory leak or how using a c++ string would help.

StasJ:
> Stas, this looks great! Can we make the icons a little bigger? Pretty hard to see. thanks!

The current snapshot icon size is the maximum natively supported by MacOS menu bars. The only way around this is to add another non-native menu somewhere else in the GUI which uses custom Qt widgets.

lib/render/Visualizer.cpp
clyne: I said "string" but I should have said C++ vector or array. I.e. a smart pointer that will clean up after itself when it goes out of scope.

StasJ: Done.

clyne (approved):

shaomeng (approved):

sgpearse (approved):

PR #2700, opened 2021-04-09
Fix #2681 Vapor prefs being overridden

### LINKED ISSUE #2681: Vapor prefs being overridden
1. Open a session
2. Check that the ram cache limit is 8000. If not, delete vapor prefs and create a new session, save it and repeat.
3. Change the ram cache limit to 12000

Notice the vapor prefs file now has a limit of 12000

4. New session
4. Import a dataset
5. Close vapor
6. Open the original session

Notice the vapor prefs file now has a limit of 8000

StasJ:
Fix #2681

The constructors are "no-fail" since if the settings file cannot be read, it just initializes with the default settings.

### DIFF
--- a/apps/vaporgui/SettingsParams.cpp
+++ b/apps/vaporgui/SettingsParams.cpp
@@ -88,7 +88,7 @@ SettingsParams::SettingsParams(ParamsBase::StateSave *ssave, bool loadFromFile) 
     // Try to get settings params from .settings file
     //
     if (loadFromFile) {
-        bool ok = _loadFromSettingsFile();
+        bool ok = LoadFromSettingsFile() >= 0;
         if (ok) return;
     }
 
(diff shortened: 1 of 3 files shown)


apps/vaporgui/SettingsParams.cpp
```diff
@@ -402,7 +402,7 @@ void SettingsParams::SetFidelityDefault2D(long lodDef, long refDef)
     SetValueLongVec(_fidelityDefault2DTag, "Set fidelity 2D default", val);
 }
 
-bool SettingsParams::_loadFromSettingsFile()
+int SettingsParams::LoadFromSettingsFile()
```
clyne, line 405: Why was the return type change from bool to int? Our error policy states that if a function logs an error message (via MyBase::SetErrMsg) it should return an int (or NULL pointer) to indicate failure. If it does not log a message it should return a boolean.

StasJ: Fixed.

sgpearse (approved):

clyne (approved):

Issue #1359, opened 2019-01-23
Control panel needs better resizing capabilities


sgpearse:
From @leigh.orf

See Leigh's "Orf Notes For Vapor Team" for context

So, I can't get that window to get any bigger. The "Data Set" text is pretty useless as you can see. I can drag the handle to the right of the smaller white box but that nukes the "New Delete Duplicate In:" buttons, and still doesn't display enough text in the Data Set region. Not sure what to suggest here, using the file name as the Data Set clearly is an issue with this configuration. Since this window tears off nicely, perhaps the easiest solution is to make it such that the window can get bigger than its current locked in horizontal geometry.

clyne:
@sgpearse please see if you can find Leigh's notes

sgpearse:
Leigh is referring to the renderer table, and how we name our data sets as their corresponding file-name.  The file-name is usually long, and gets clipped.  Expanding the left panel has been improved since this report, so we can display more text, but I don't think anybody wants to use Vapor with the control panel taking up half of the window.  

While this problem has a workaround, we might be able to do better by removing one of the columns in the renderer table.  Since all of the renderers are named after their type, ihe "Type" column is redundant, imo.

![Screen Shot 2021-01-14 at 8 32 25 AM](https://user-images.githubusercontent.com/9522770/104612105-13ce0b00-5643-11eb-832f-19caa883c76a.png)

PR #2707, opened 2021-04-20
Fix Annotation and Transform tables in the Scene tab

### LINKED ISSUE #2705: Regression: VaporTable not honoring inputs
The Scene tab has two VaporTables that do not honor user input: 1) in the Annotation tab, where the # Tics, Size, Min, Max, and Origin of the axis annotations are defined, and 2) in the Viewpoint tab, where our dataset transforms are defined.

This may warrant a 3.4.1 hot-fix release.

sgpearse:
This fixes #2705, which reports that the AnnotationTable and TransformTable in the Scene tab would not honor user input.

Prior to 3.4.0, each cell in a VaporTable contained a QWidget, typically a QLineEdit or QCheckBox.  The recent fix for #1359 involved replacing QLineEdits with QTableWidgetItems to improve resizing of the RenderHolder (the table that lists renderer instances) and to simplify the class.

Previously, the QLineEdits in table cells would emit signals when their values changed.  Now that they have been removed, we need something to replace their signal emission.  This PR emits the signal when the a table's QTableWidget::cellChanged fires.  See VaporTable.cpp:48

Note - this branch is based off of 987b575, where we bumped Vapor's version number to 3.4.1.  It is not derived from the head of our main branch.  This was done in case we want to apply this fix as a patch that only fixes this bug, and nothing else.

This has been tested on all platforms, documented in [this spreadsheet](https://docs.google.com/spreadsheets/d/1MEwLRPQXk-KUqlvQ9x4yjv6IfYj4zch_CFs3Us3ACPc/edit?usp=sharing).

### DIFF
--- a/apps/vaporgui/AnnotationEventRouter.cpp
+++ b/apps/vaporgui/AnnotationEventRouter.cpp
@@ -117,7 +117,7 @@ void AnnotationEventRouter::connectAnnotationWidgets()
     connect(_ticWidthCombo, SIGNAL(valueChanged(double)), this, SLOT(setAxisTicWidth(double)));
     connect(axisColorButton, SIGNAL(pressed()), this, SLOT(setAxisColor()));
     connect(axisBackgroundColorButton, SIGNAL(pressed()), this, SLOT(setAxisBackgroundColor()));
-    connect(_annotationVaporTable, SIGNAL(returnPressed()), this, SLOT(axisAnnotationTableChanged()));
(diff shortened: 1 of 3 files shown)


StasJ (approved):
The issue only appears to manifest in new sessions, old session files work correctly. Either way, this fixes it.

clyne:
@sgpearse merging this PR. If you want to put out a 3.4.1 it is ready.

PR #2721, opened 2021-05-10
Update checker

### LINKED ISSUE #2597: Restore version checker from vapor2
Restore the VAPOR VersionChecker capability from VAPOR2 that checks to ensure the user is using the most current version of vapor and warns them if they are not.

StasJ:
It uses the GitHub API to query our releases and parses them to find the newest non-prerelease version. By default it will check for new versions upon startup in the background (i.e. the application won't hang over a slow connection). If a newer version is available, it prompts the user to download it, providing a button that takes the user directly to the download page of that release, as well as a checkbox to prevent further notifications about new versions. The version can be manually checked in the settings panel as well. Automatic version checking is disabled when running Vapor in batch mode, as well as in debug builds.

### Popup
popup

### Settings Panel
settings

### Settings New Version Available
settings new

### Settings Error Fetching Version
settings error

Fix #2597

### DIFF
--- a/apps/vaporgui/AppSettingsMenu.cpp
+++ b/apps/vaporgui/AppSettingsMenu.cpp
@@ -8,6 +8,80 @@
 #include "PCheckboxHLI.h"
 #include "VPushButton.h"
 #include "ErrorReporter.h"
+#include "CheckForUpdate.h"
+
+class PUpdateChecker : public PWidget
+{
+    VGroup *_group;
+    QLabel *_label;
+    VPushButton *_checkButton;
+    VPushButton *_getButton;
+    UpdateInfo _updateInfo;
+    
+    void updateGUI() const override {}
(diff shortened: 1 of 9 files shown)


shaomeng (approved):

sgpearse (approved):

clyne (commented):
Is there a way to demonstrate what happens when a newer version is available? I'm assuming nothing will be triggered until we have a newer release posted to the web site?

StasJ:
> Is there a way to demonstrate what happens when a newer version is available? I'm assuming nothing will be triggered until we have a newer release posted to the web site?

If you replace line 49 in CheckForUpdate.cpp with `updateAvailable = true;`, it will treat the first release it finds as a new update.

clyne (approved):
Nice!

PR #2729, opened 2021-05-14
Fix #2723 Color bar shows constant value and color on non-constant variables #2723

### LINKED ISSUE #2723: Color bar shows constant value and color on non-constant variables
Steps to reproduce:
1. Open the following session file on Casper. The data file will be loaded automatically.
2. Notice that the flow lines show different colors, however, the color bar has only one color and one value.
3. Examine the color mapping variable using the histogram. Notice that this variable has a non-constant data field.

[CircularRibbonFlare.vs3.tar.gz](https://github.com/NCAR/VAPOR/files/6461193/CircularRibbonFlare.vs3.tar.gz)

StasJ:
I would like to note that the bug this PR fixes is unrelated to the new color bar renderer.

Fix #2723

### DIFF
--- a/include/vapor/VolumeParams.h
+++ b/include/vapor/VolumeParams.h
@@ -100,7 +100,7 @@ public:
     static const std::string OSPSampleRateScalar;
     static const std::string OSPAmbientLightIntensity;
     static const std::string OSPDirectionalLightIntensity;
-    
+
     static const std::string OSPVolmeAlgorithmName;
 };
 
--- a/include/vapor/VolumeParams.h
+++ b/include/vapor/VolumeParams.h
@@ -58,7 +58,7 @@ public:
     //! \copydoc RenderParams::GetActualColorMapVariableName()
     virtual string GetActualColorMapVariableName() const override
     {
-        if (STLUtils::Contains(STLUtils::ToLower(GetAlgorithm()), "ospray"))
+        if (GetAlgorithm() == OSPVolmeAlgorithmName)
             return GetVariableName();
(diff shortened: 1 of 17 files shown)


include/vapor/VolumeParams.h
```diff
@@ -54,6 +55,17 @@ class PARAMS_API VolumeParams : public RenderParams {
     //
     virtual size_t GetRenderDim() const override { return (3); }
 
+    //! \copydoc RenderParams::GetActualColorMapVariableName()
+    virtual string GetActualColorMapVariableName() const override
+    {
+        if (STLUtils::Contains(STLUtils::ToLower(GetAlgorithm()), "ospray"))
```
clyne, line 61: Can we do something more robust here? Checking for a hard-coded string name is pretty ugly

StasJ: Done.

shaomeng:
The PR explains the root of the user experience bug as the following:
```
    //! Due to legacy code, the "ColorMapVariableName" is not  the name of the variable by which
    //! the renderer is colormapping, but rather the variable it should use if it is not already using its
    //! primary variable or a constant color. This function should return the name of the variable that
    //! is currently being used for colormapping.
```

And the current code seems pretty confusing with a few different functions seemingly doing the same thing based on their names. My question is, should we aim for a fix of the root cause or a bandaid to resolve this particular bug?

shaomeng (approved):

clyne (approved):

PR #2734, opened 2021-05-20
Particles

### LINKED ISSUE #2349: Add support for particle visualization
Add support for visualizing Lagrangian particle data

StasJ:
Fix #2349

You can generate test data with `vapor/share/examples/generateExampleDCP.py`

# DCP Particle Data Format

**D**ata **C**ollection **P**articles is a simple data format built upon NetCDF for importing particle data into VAPoR.

### P
P is on only required dimension. The size of the dimension P represents the number of particles at this timestep. If the number of particles varies between timesteps, every timestep must be in its own file since each NetCDF file can only define a single length for a dimension.

### T
T is the time dimension and it is required for time-varying data. It should be set to *unlimited*.
### axis
This optional dimension is must have a length of 3 and it is used to pack 3 component values such as position or velocity into a single variable for convenience.

## Variables

The only required variable is Position. Non-time-varying data should have a single dimension of `P`. If the data is time-varying, it should have the dimensions `(T, P)`. 3D particles will often have 3-component data associated with them, for example their velocity would have 3 components, one for each dimension. Since this is a common use case, variables can have 3 components per particle and this is done by setting the fastest-varying dimension to `axis` which as a size of 3. For an example, look at the position variable description below.

### Position

This variable contains the position data for the particles. The data can be stored in either of the below methods:
```
float Position(T, P, axis) ;
```
```
float Position_x(T, P) ;
float Position_y(T, P) ;
float Position_z(T, P) ;
```

### Python Script

`vapor/share/examples/generateExampleDCP.py` shows how to generate a DCP dataset from a particle simulation. 

### Basic NetCDF File

This is an example of a NetCDF file containing 2 particles with time-varying position and particle speed data.

```
netcdf particles_000 {
dimensions:
	P = 2 ;
	T = UNLIMITED ; // (1 currently)
	axis = 3 ;
variables:
	double T(T) ;
		T:units = "seconds" ;
	float Position(T, P, axis) ;
	float speed(T, P) ;
data:

 T = 0 ;

 Position =
  9.762701, 43.03787, 20.55268,
  8.976637, -15.26904, 29.17882 ;

 speed =
  26.02757, 18.87516 ;
}
```


StasJ:
@sgpearse I tried to run the clang-format script but it is not doing anything.

clyne:
This isn't building for me:

/Users/clyne/src/vapor/apps/vdcdump/vdcdump.cpp:302:95: error: too few arguments to function call,
      expected 3, have 2
        for (int i = 0; i < dimNames.size(); i++) vdc.GetDimension(dimNames[i], _dimensions[i]);
                                                  ~~~~~~~~~~~~~~~~                            ^

apps/vaporgui/ParticleEventRouter.h
```diff
@@ -0,0 +1,23 @@
...
+//! \ingroup Public_GUI
+//! \brief Particle renderer GUI
+//! \author Stas Jaroszynski
+
+class ParticleEventRouter : public RenderEventRouterGUI {
+public:
+    ParticleEventRouter(QWidget *parent, VAPoR::ControlExec *ce);
+    static string GetClassType() { return VAPoR::ParticleRenderer::GetClassType(); }
+    string        GetType() const { return GetClassType(); }
+    bool          Supports2DVariables() const { return true; }
```
clyne, line 16: Should Supports{2D,3D} be returning true? This methods refers to a variable's **topolological** (Euclidean) dimension: E.g. zero for points, one for lines, two for planes, and so on.

include/vapor/DC.h
```diff
@@ -2107,6 +2116,10 @@ class VDF_API DC : public Wasp::MyBase {
     //
     virtual bool getDimension(string dimname, DC::Dimension &dimension) const = 0;
 
+    //! \copydoc GetDimension()
+    //
+    virtual bool getDimension(string dimname, DC::Dimension &dimension, long ts) const { return getDimension(dimname, dimension); };
```
clyne, line 2121: ts in unused. How come?

include/vapor/DataMgr.h
```diff
@@ -155,10 +155,10 @@ class VDF_API DataMgr : public Wasp::MyBase {
 
     //! \copydoc DC::GetDimension()
     //
-    bool GetDimension(string dimname, DC::Dimension &dimension) const
+    bool GetDimension(string dimname, DC::Dimension &dimension, long ts) const
```
clyne, line 158: Wouldn't it make more sense to give ts a default value (-1) so that we don't have to change all of the users of this class? Ditto for some of the other classes for which at 'ts' argument was added?

StasJ: For most of the functions I changed, I did make the ts argument optional (although by overloading rather than a default value, e.g. your comment above about ts being unused). In the case of this function, I determined it was important for the callee to be explicitly aware of whether they pass in the correct timestep since this function is often used to present information to the user/get information about the datasets outside of the internal systems which would already be aware of whether they need the timestep information. It would therefore make it very easy for end users to call this function without being aware that they might need timestep information which is often the case, now that we've added this capability.

include/vapor/DC.h
StasJ: DCs do not support time-varying dimensions by default.

apps/vaporgui/ParticleEventRouter.h
StasJ: Fixed.

StasJ:
> This isn't building for me:
> 
> /Users/clyne/src/vapor/apps/vdcdump/vdcdump.cpp:302:95: error: too few arguments to function call,
> expected 3, have 2
> for (int i = 0; i < dimNames.size(); i++) vdc.GetDimension(dimNames[i], _dimensions[i]);

Fixed.

include/vapor/DataMgr.h
clyne: okey doke

lib/vdc/UnstructuredGrid3D.cpp
```diff
@@ -0,0 +1,278 @@
...
+    VAssert(zug.GetDimensions().size() == 1);
+
+    VAssert(location == NODE);
+}
+
+
+vector<size_t> UnstructuredGrid3D::GetCoordDimensions(size_t dim) const
+{
+    if (dim == 0) {
+        return _xug.GetDimensions();
```
clyne, line 41: These should all be the same size, correct? Just a sanity check.

lib/vdc/UnstructuredGrid3D.cpp
```diff
@@ -0,0 +1,278 @@
...
+    float               zwgt[2];
+
+    return (_insideGrid(coords, indices, nodes2D, lambda, zwgt));
+}
+
+
+bool UnstructuredGrid3D::_insideGrid(const DblArr3 &coords, Size_tArr3 &cindices, std::vector<size_t> &nodes2D, std::vector<double> &lambda, float zwgt[2]) const
+{
+    VAssert(_location == NODE);
+    VAssert(!"Not implemented");
```
clyne, line 143: How about deleting the rest of the body of this function since it's not being used? If it's a WIP than maybe just ifdef it out.

share/examples/generateExampleDCP.py
```diff
@@ -0,0 +1,75 @@
+# Run this file to create an example DCP dataset
+
+
+from netCDF4 import Dataset
+import numpy as np
+
+
+N = 200
```
clyne, line 8: Can we add command line arguments to control the number of points and time steps? Optionally outputting velocities would also be useful.

lib/render/ParticleRenderer.cpp
```diff
@@ -0,0 +1,163 @@
...
+            _dataMgr->GetVarCoordVars(var, true, varCoords);
+
+            if (mainVarCoords != varCoords) {
+                if (grid) delete grid;
+                for (auto g : vecGrids) delete g;
+                SetErrMsg("Variable \"%s\" on different grid from main variable", var.c_str());
+                return -1;
+            }
+
+            Grid *ng = _dataMgr->GetVariable(p->GetCurrentTimestep(), var, p->GetRefinementLevel(), p->GetCompressionLevel(), minExt, maxExt);
```
clyne, line 105: Better to use DataMgrUtils::GetGrids which will fetch multiple variables at once and temporarily "lock" them so that fetching one variable doesn't purge another one from the cache. Alternatively, you can use DataMgr::GetVariable() with lock=true, but then you'll need to explicitly unlock them.

lib/vdc/DataMgr.cpp
```diff
@@ -635,6 +644,40 @@ int DataMgr::Initialize(const vector<string> &files, const std::vector<string> &
         SetErrMsg("Failed to get time coordinates");
         return (-1);
     }
+
+#if DCP_ENABLE_PARTICLE_DENSITY
```
clyne, line 648: Is this just test code? Are we planning on enabling it? Need to handle error returns below if so. Also, it looks like the position_x variable is hardcoded. Is that reasonable?

lib/vdc/DerivedParticleDensity.cpp
```diff
@@ -0,0 +1,355 @@
...
+    int    xd = dims[0], yd = dims[1], zd = dims[2];
+    float *data = new float[xd * yd * zd];
+    compute(g, data, xd, yd, zd, realNP);
+    delete g;
+
+    //    assert(max[0]-min[0]+1 == xd);
+    //    assert(max[1]-min[1]+1 == yd);
+    //    assert(max[2]-min[2]+1 == zd);
+
+    //    for (int z = 0; z < zd; z++)
```
clyne, line 94: Can we remove this?

lib/vdc/DerivedParticleDensity.cpp
```diff
@@ -0,0 +1,355 @@
...
+            assert(0);
+            return -1;
+        }
+        size_t nCoords = dim.GetLength();
+        float *coords = new float[nCoords];
+        ts = f->GetTS();
+
+        _dc->ReadRegion(fd, {0}, {nCoords - 1}, coords);
+        float min = FLT_MAX;
+        float max = -FLT_MAX;
```
clyne, line 311: Will this work as anticipated? I.e. does it give the same result as std::numeric_limits<float>::lowest()?

lib/vdc/DerivedVar.cpp
```diff
@@ -865,21 +864,19 @@ int DerivedCoordVar_CF1D::GetDimLensAtLevel(int level, std::vector<size_t> &dims
 
 int DerivedCoordVar_CF1D::OpenVariableRead(size_t ts, int level, int lod)
 {
-    if (level != 0) {
-        SetErrMsg("Invalid parameter");
-        return (-1);
-    }
+    //    if (level != 0) {
```
clyne, line 867: REmove commented out code

lib/vdc/GridHelper.cpp
```diff
@@ -541,6 +549,117 @@ UnstructuredGridLayered *GridHelper::_make_grid_unstructured_layered(size_t ts,
...
+    // a shared pointer so that we can cache it for use by other Grid
+    // classes. This a peformance optimization, necessary be creating
+    // a QuadTreeRectangle is expensive.
+    //
+    std::shared_ptr<const QuadTreeRectangle<float, size_t>> qtr = _qtrCache.get(qtr_key);
+
+    UnstructuredGrid3D *g = new UnstructuredGrid3D(vertexDims, faceDims, edgeDims, bs, blkptrs, vertexOnFace, faceOnVertex, faceOnFace, location, maxVertexPerFace, maxFacePerVertex, vertexOffset,
+                                                   faceOffset, xug, yug, zug, qtr);
+
+    // No QuadTreeRectangle in cache. So get shared pointer for one created
```
clyne, line 650: Remove commented out code. A quadtree wouldn't work for 3D data anyway.

clyne (commented):
Stas, I've finished reviewing the code and left a number of minor in-line comments. A few more general things that may need some discussion:

1. We use the term "dimension" throughout the application pretty sloppily. It can refer to the Topological (Euclidean) dimension of the individual elements making up a mesh (e.g. points, lines, polygons, polyhedra), or the number of indices needed to address a node or cell in a mesh, or the number of coordinate variables. Sometimes these are all the same, but not always. In the case of particle data the Topological dimension is 0; we have a collection of points with no "cells". You've chosen to represent them internally as a 3D unstructured grid that has no connectivity information. This is reasonable, but maybe leads to confusion. Perhaps it should be a 0D unstructured grid (maybe just call it "Points", but still derive from the UnstructuredGrid class). The real issue is that DC::Mesh::GetTopologyDim() returns 3 for the new class, which is not really correct. One implication of this confusion is that all of the renderers that support 3D data (e.g. slice, volume) are activated for particle data (and will crash if  enabled).
2. A minor thing, but the NetCDF particle format is not recognized if you pass the files to the application from the command line.
3. What are you thinking for the "Melanie" data reader? Currently its a hidden feature. Maybe just ifdef it out?
4. Once we resolve (1) we should add the new particle grid class to Scott's 'smoke tests' (gridTools.cpp and testGrid.cpp). This may flush out some bugs.
5. I see there is a .md file for the DCP data format. We need to make sure this is accessible from the documentation on our web site.

lib/vdc/GridHelper.cpp
StasJ: Done.

lib/vdc/DerivedVar.cpp
StasJ: Done.

lib/vdc/DerivedParticleDensity.cpp
StasJ: Yes.

lib/vdc/DerivedParticleDensity.cpp
StasJ: This is part of an in-progress prototype so it has some commented out code that I was hoping to leave but I can remove it if you insist.

lib/vdc/DataMgr.cpp
StasJ: This is prototype code for automatically generating particle density fields when users load particle data. It is in a stage where if you set DCP_ENABLE_PARTICLE_DENSITY it will work but it will have a fixed grid size and it uses a sub-optimal sampling algorithm.

lib/render/ParticleRenderer.cpp
StasJ: Done

share/examples/generateExampleDCP.py
StasJ: This script is designed to show users how to output DCP data. It is very heavily documented and designed to be edited by the end user. I think it would be reasonable that if someone, for some reason, wants to customize the example dataset they just edit the file. Also, this script is designed to be as simple and self-explanatory as possible and adding a bunch of argument-handling will, I believe, complicate it unnecessarily. If you insist, however, I will add them.

lib/vdc/UnstructuredGrid3D.cpp
StasJ: Done

lib/vdc/UnstructuredGrid3D.cpp
StasJ: Correct

lib/vdc/DataMgr.cpp
clyne: Could you add a comment to that effect?

StasJ:
> Stas, I've finished reviewing the code and left a number of minor in-line comments. A few more general things that may need some discussion:
> 
> 1. We use the term "dimension" throughout the application pretty sloppily. It can refer to the Topological (Euclidean) dimension of the individual elements making up a mesh (e.g. points, lines, polygons, polyhedra), or the number of indices needed to address a node or cell in a mesh, or the number of coordinate variables. Sometimes these are all the same, but not always. In the case of particle data the Topological dimension is 0; we have a collection of points with no "cells". You've chosen to represent them internally as a 3D unstructured grid that has no connectivity information. This is reasonable, but maybe leads to confusion. Perhaps it should be a 0D unstructured grid (maybe just call it "Points", but still derive from the UnstructuredGrid class). The real issue is that DC::Mesh::GetTopologyDim() returns 3 for the new class, which is not really correct. One implication of this confusion is that all of the renderers that support 3D data (e.g. slice, volume) are activated for particle data (and will crash if  enabled).

All of your points here have been addressed. Particle data now has a topological dimension of 0. There is now a distinction between particle data and regular data. The particle renderer is now only available for datasets with particle data and the other renderers for datasets with regular data. The variable selectors also now only show particle or regular data, depending on the renderer. A few spots where the topological and geometric dimensions were used incorrectly have also been fixed since having a 0D dataset caused bugs to surface.

> 2. A minor thing, but the NetCDF particle format is not recognized if you pass the files to the application from the command line.

This has been fixed.

> 3. What are you thinking for the "Melanie" data reader? Currently its a hidden feature. Maybe just ifdef it out?

I ifdef'd out the Melanie data reader.

> 4. Once we resolve (1) we should add the new particle grid class to Scott's 'smoke tests' (gridTools.cpp and testGrid.cpp). This may flush out some bugs.

A large part of the "smoke tests" is testing sampling the grid at user coordinates which the particles grid does not support. We can implement the iterator tests but the node iterators are pretty much no-ops for particle data and the coord iterators are just 3 UnstructuredGridCoordless iterators.

> 5. I see there is a .md file for the DCP data format. We need to make sure this is accessible from the documentation on our web site.

I think this should probably wait until we merge the PR or maybe even until the next release since the feature will be unavailable to most users.

lib/vdc/DataMgr.cpp
StasJ: Done.

include/vapor/NetCDFCollection.h
```diff
@@ -780,6 +784,8 @@ class VDF_API NetCDFCollection : public Wasp::MyBase {
     std::vector<string>              _staggeredDims;
     std::vector<string>              _dimNames;    // Names of all dimensions
     std::vector<size_t>              _dimLens;     // Names of all dimensions
+    std::vector<bool>                _dimIsTimeVarying;
+    //    std::vector<
```
shaomeng, line 788: do you want to delete this line?

include/vapor/UnstructuredGrid3D.h
```diff
@@ -0,0 +1,110 @@
...
+    UnstructuredGrid3D(const std::vector<size_t> &vertexDims, const std::vector<size_t> &faceDims, const std::vector<size_t> &edgeDims, const std::vector<size_t> &bs, const std::vector<float *> &blks,
+                       const int *vertexOnFace, const int *faceOnVertex, const int *faceOnFace,
+                       Location location,    // node,face, edge
+                       size_t maxVertexPerFace, size_t maxFacePerVertex, long nodeOffset, long cellOffset, const UnstructuredGridCoordless &xug, const UnstructuredGridCoordless &yug,
+                       const UnstructuredGridCoordless &zug);
+
+    UnstructuredGrid3D() = default;
+    virtual ~UnstructuredGrid3D() = default;
+
+    // std::shared_ptr<const QuadTreeRectangle<float, size_t> >GetQuadTreeRectangle() const {
```
shaomeng, line 35: do you want to delete this unused function ?

lib/render/ParticleRenderer.cpp
```diff
@@ -0,0 +1,179 @@
...
+            const glm::vec3 n(*(dirs[0]), *(dirs[1]), *(dirs[2]));
+            const glm::vec3 p2 = p + n * dirScale;
+            lgl->Vertex3fv((float *)&p2);
+
+            for (auto &it : dirs) ++it;
+        }
+    }
+    lgl->End();
+
+    //    printf("Rendered %li particles\n", renderedParticles);
```
shaomeng, line 168: do you want to delete the debug code?

lib/vdc/DerivedParticleDensity.cpp
```diff
@@ -0,0 +1,355 @@
...
+    dvar.SetName(_derivedVarName);
+    dvar.SetMeshName(_meshName);
+    dvar.SetUnits("");
+    dvar.SetXType(DC::FLOAT);
+    dvar.SetWName("");
+    dvar.SetCRatios(vector<size_t>());
+    dvar.SetPeriodic(vector<bool>(3, false));
+
+    //    dvar.SetHasMissing(true);
+    //    dvar.SetMissingValue(-1);
```
shaomeng, line 188: remove?

lib/vdc/DerivedParticleDensity.cpp
```diff
@@ -0,0 +1,355 @@
...
+    dims_at_level.clear();
+    DC::Mesh mesh;
+    _dataMgr->GetMesh(_meshName, mesh);
+    DC::Dimension dim;
+    for (auto dimName : mesh.GetDimNames()) {
+        _dataMgr->GetDimension(dimName, dim, -1);
+        dims_at_level.push_back(dim.GetLength());
+    }
+
+    //    dims_at_level = {128, 128, 128};
```
shaomeng, line 160: remove this?

lib/vdc/NetCDFCollection.cpp
```diff
@@ -141,8 +144,9 @@ int NetCDFCollection::Initialize(const vector<string> &files, const vector<strin
             else if (find(l_time_dimnames.begin(), l_time_dimnames.end(), dimnames[j]) != l_time_dimnames.end()) {
                 _dimLens[itr - _dimNames.begin()] += dims[j];
             } else if (_dimLens[itr - _dimNames.begin()] != dims[j]) {
-                SetErrMsg("Spatial dimension %s changed size", dimnames[j].c_str());
-                return (-1);
+                _dimIsTimeVarying[itr - _dimNames.begin()] = true;
+                //                SetErrMsg("Spatial dimension %s changed size", dimnames[j].c_str());
+                //                return (-1);
```
shaomeng, line 149: remove debug code?

lib/vdc/UnstructuredGrid3D.cpp
```diff
@@ -0,0 +1,236 @@
...
+    minu[1] = range[0];
+    maxu[1] = range[1];
+
+    _zug.GetRange(min, max, range);
+    minu[2] = range[0];
+    maxu[2] = range[1];
+
+    //    printf("UnstructuredGrid3D::%s(-, -)\n", __func__);
+    //    printf("\t min = [%f, %f, %f]\n", minu[0], minu[1], minu[2]);
+    //    printf("\t max = [%f, %f, %f]\n", maxu[0], maxu[1], maxu[2]);
```
shaomeng, line 97: Remove debug code?

shaomeng (approved):
Minor changes required, otherwise things look good!

clyne:
@sgpearse will open an issue to add user documentation

lib/vdc/UnstructuredGrid3D.cpp
StasJ: I would like to leave this for later work.

lib/vdc/NetCDFCollection.cpp
StasJ: Done.

lib/vdc/DerivedParticleDensity.cpp
StasJ: Done.

lib/vdc/DerivedParticleDensity.cpp
StasJ: I would like to leave this for future reference.

lib/render/ParticleRenderer.cpp
StasJ: I would like to leave this for later work.

include/vapor/UnstructuredGrid3D.h
StasJ: Done.

include/vapor/NetCDFCollection.h
StasJ: Done.

include/vapor/DCP.h
```diff
@@ -0,0 +1,192 @@
...
+    string                          _proj4String;
+    std::map<string, DC::Dimension> _dimsMap;
+    std::map<string, DC::CoordVar>  _coordVarsMap;
+    std::map<string, DC::AuxVar>    _auxVarsMap;
+    std::map<string, DC::Mesh>      _meshMap;
+    std::map<string, DC::DataVar>   _dataVarsMap;
+    std::map<string, string>        _coordVarKeys;
+    std::map<string, string>        _sanitizedToOriginalMap;
+
+    const string          nodeFaceVar = "cellsOnVertex";
```
clyne, line 170: Please follow underscore convention for naming private variables.

clyne (commented):
Stas, this is really great! A few comments/questions

1. You implemented a novel particle visualization capability a while back that can be seen at about the 1'30" mark of this [video](https://www.youtube.com/watch?v=tZxH4-FcAqY). 

How much effort would be involved in adding support for your faded particle traces?

2. it would be really beneficial to be able to subsample the data somewhere in the pipeline. This reduces visual clutter, and improves performance. A simple, user-specifiable stride would help a lot. Some kind of random sampling would be even better. I'm not suggesting you implement these, but you might check out "smarter" sampling approaches in the literature such as in [here](https://dl.acm.org/doi/pdf/10.1145/3281464.3281467https://dl.acm.org/doi/pdf/10.1145/3281464.3281467) and references therein. 

3. Rendering performance is not so great. I generated 2M particles and the frame rates on my Mac were pretty slow. What would could be done to improve this? It looks like currently the entire render pipeline executes for each paint event (i.e. there is no caching). FYI some particle data sets can have billions of particles, but order millions is more common. Again, a down sampling scheme is probably critical.

StasJ:
> Stas, this is really great! A few comments/questions
> 
> 1. You implemented a novel particle visualization capability a while back that can be seen at about the 1'30" mark of this [video](https://www.youtube.com/watch?v=tZxH4-FcAqY).
> 
> How much effort would be involved in adding support for your faded particle traces?

There are two different techniques mentioned here:
- The method in the video would be doable, I don't know how much effort exactly but it would not be unreasonable.
- Adding support for faded particle traces would take a large amount of work. This would require tracking particles through time using their particle id which is currently not possible as Vapor doesn't support integer data grids, in addition to the work it would require to implement the rest of the feature.

> 
> 1. it would be really beneficial to be able to subsample the data somewhere in the pipeline. This reduces visual clutter, and improves performance. A simple, user-specifiable stride would help a lot. Some kind of random sampling would be even better. I'm not suggesting you implement these, but you might check out "smarter" sampling approaches in the literature such as in [here](https://dl.acm.org/doi/pdf/10.1145/3281464.3281467https://dl.acm.org/doi/pdf/10.1145/3281464.3281467) and references therein.

This is already in the PR. Under the data fidelity section there is a `Stride` parameter.

> 2. Rendering performance is not so great. I generated 2M particles and the frame rates on my Mac were pretty slow. What would could be done to improve this? It looks like currently the entire render pipeline executes for each paint event (i.e. there is no caching). FYI some particle data sets can have billions of particles, but order millions is more common. Again, a down sampling scheme is probably critical.

I agree, and there is already a downsampling option as mentioned above. I think this falls in the same situation as adding more rendering features. Any and all of these things can be added, it is just a matter of is this feature going to be used enough to warrant the effort of implementing these features. My understanding was that this PR was going to create a basic implementation that would be expanded upon if there is demand.
Of course, if there are any features you would like me to implement, just let me know and I will do so.

include/vapor/DCP.h
StasJ: Done

clyne (approved):

PR #2737, opened 2021-05-21
Issue 2438s5

sgpearse:
@clyne This PR does 2 things:

1) It updates our "baseline" files for the smoke tests, which have had minor format and information changes.  The 'diff' that we run between our test results and these baseline files is why the tests were reporting mismatches.

2) It applies clang-format to the files that were missed during the pre-push hook.  I have a separate PR that fixes the hook, which was applying clang-format to all commits, but not all merge-commits.  More details in #2736

### DIFF
--- a/test_apps/smokeTests/testResults/cf_baseline.txt
+++ b/test_apps/smokeTests/testResults/cf_baseline.txt
@@ -61,4 +61,4 @@ Time Coordinates:
 
 Time coordinate variable name: time
 Number of time steps: 1
-Elapsed time: 0.31866
+Elapsed time: 0.159113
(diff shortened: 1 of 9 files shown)


test_apps/smokeTests/testResults/cf_baseline.txt
```diff
@@ -60,4 +61,4 @@ Time Coordinates:
 
 Time coordinate variable name: time
 Number of time steps: 1
-Elapsed time: 0.318985
+Elapsed time: 0.159113
```
clyne, line 64: @sgpearse can we remove the output of elapsed time when creating these reference files so don't keep running into this?

sgpearse: This wasn't the issue - diff found that there was a new std::endl in one of the printouts, as well as another line reporting the number of threads used.  The elapsed time is not compared in the diff, noted in [smokeTests.py:173](https://github.com/NCAR/VAPOR/blob/b457c5c5f59a52f28e8d5f7d3fcb8820c6d6915c/test_apps/smokeTests/smokeTests.py#L173)

clyne (approved):

Issue #2718, opened 2021-05-03
AxisAnnotation: Refactor to use the Box class


sgpearse:
The AxisAnnotation class has records of its minimum and maximum extents, which could be replaced with the Box class.  This would also allow for the use of PCopyRegionWidget in the AnnotationEventRouter.

PR #2738, opened 2021-05-25
Issue2718 - Refactor AxisAnnotation to use Box, and add PCopyRegionAnnotationWidget

sgpearse:
Note: I'd like to merge #2713 into main before merging this to keep these PR's easier to review and test.  I'm leaving it as a draft for now, and if we decide it's ok to merge #2713 into main, we can re-target this PR to main after that.

Fixes issue #2718 by refactoring the AxisAnnotation class to use the Box instead of vectors for describing the extends of the axis annotation.

Also introduces the CopyRegionAnnotationWidget, which is a subclass of CopyRegionWidget.  Also introduces its associated the PWidget (PCopyRegionAnnotationWidget) into the AxisAnnotationEventRouter.


apps/vaporgui/PCopyRegionAnnotationWidget.h
```diff
@@ -0,0 +1,22 @@
...
+class ControlExec;
+}
+
+class CopyRegionAnnotationWidget;
+
+//! \class PPCopyRegionAnnotationWidget
+//! \brief PWidget wrapper for the PCopyRegionAnnotationWidget
+
+class PCopyRegionAnnotationWidget : public PWidget {
+    CopyRegionAnnotationWidget *_widget;
```
StasJ, line 15: This var is never used.

apps/vaporgui/PCopyRegionAnnotationWidget.cpp
```diff
@@ -0,0 +1,10 @@
+#include "PCopyRegionAnnotationWidget.h"
+#include "CopyRegionAnnotationWidget.h"
+#include <vapor/ControlExecutive.h>
+#include <vapor/RenderParams.h>
+
+using namespace VAPoR;
+
+PCopyRegionAnnotationWidget::PCopyRegionAnnotationWidget(ControlExec *ce) : PWidget("", _widget = new CopyRegionAnnotationWidget(ce)) {}
+
+void PCopyRegionAnnotationWidget::updateGUI() const { _widget->Update(getParamsMgr()); }
```
StasJ, line 10: Why would the update method need the ParamsMgr as a parameter if the control exec already has a copy?

apps/vaporgui/CopyRegionAnnotationWidget.h
```diff
@@ -0,0 +1,42 @@
...
+                  "region to asix annotations");
+    }
+    QString whatsThis() const
+    {
+        return tr("This widget contains all widgets "
+                  "necessary for copying renderer regions "
+                  "into a set of axis annotations");
+    }
+
+    virtual void Update(VAPoR::ParamsMgr *pm);
```
StasJ, line 33: I am assuming this is supposed to override `CopyRegionWidget::Update` since you made that function virtual. It has a different signature thought so it won't.

apps/vaporgui/CopyRegionAnnotationWidget.h
```diff
@@ -0,0 +1,42 @@
...
+public:
+    CopyRegionAnnotationWidget(VAPoR::ControlExec *ce);
+
+    QString name() const { return "CopyRegionAnnotationWidget"; }
+    QString includeFile() const { return "CopyRegionAnnotationWidget.h"; }
+    QString group() const { return tr("A widget for copying one renderer's region to axis annotations"); }
+    QString toolTip() const
+    {
+        return tr("A widget for copying one renderer's "
+                  "region to asix annotations");
```
StasJ, line 24: Typo

apps/vaporgui/CopyRegionAnnotationWidget.h
```diff
@@ -0,0 +1,42 @@
...
+}    // namespace VAPoR
+
+class CopyRegionAnnotationWidget : public CopyRegionWidget {
+    Q_OBJECT
+
+public:
+    CopyRegionAnnotationWidget(VAPoR::ControlExec *ce);
+
+    QString name() const { return "CopyRegionAnnotationWidget"; }
+    QString includeFile() const { return "CopyRegionAnnotationWidget.h"; }
```
StasJ, line 19: What are these for?

StasJ (commented):
Could we refactor axis annotations to use the box class and the generic CopyRegionWidget?

apps/vaporgui/PCopyRegionAnnotationWidget.h
StasJ: I misread the var type

apps/vaporgui/PCopyRegionAnnotationWidget.cpp
sgpearse: Fair.  It's now accessed through the controlExec.

apps/vaporgui/CopyRegionAnnotationWidget.h
sgpearse: These are legacy functions that should have been removed.  Fixed.

apps/vaporgui/CopyRegionAnnotationWidget.h
sgpearse: Thanks.  Fixed.

apps/vaporgui/CopyRegionAnnotationWidget.h
sgpearse: You're correct.  Fixed

StasJ (approved):

clyne (approved):
Nice work!

Issue #2704, opened 2021-04-20
Provide support for raw data (BOV format)


shaomeng:
It'd be good for VAPOR to support raw data, as raw data is often used when prototyping. Also, given the simple nature of raw data, it shouldn't be a big effort. 

Note that it's often helpful to have a header associated with a raw data file to provide meta data. I'd suggest to use the BOV format used by VisIt because of its simplicity.  (See Section 3.1 of VisIt manual.)
[GettingDataIntoVisIt2.0.0.pdf](https://github.com/NCAR/VAPOR/files/6341067/GettingDataIntoVisIt2.0.0.pdf)

PR #2740, opened 2021-05-27
Issue2704 - BOV Reader

sgpearse:
An operational BOV reader, addressing #2704.  Supports INT (32 bit), FLOAT, and DOUBLE data types.

Limitations:

1) The BOV tokens listed below are not supported.  Read more about these tokens on page 11 of [GettingDataIntoVisIt2.0.0.pdf](https://github.com/NCAR/VAPOR/files/6341067/GettingDataIntoVisIt2.0.0.pdf).

   - CENTERING
   - BYTE_OFFSET
   - DIVIDE_BRICK
   - DATA_BRICKLETS
   - DATA_COMPONENTS

2) The file specified by the DATA_FILE token must use an absolute path, not a relative path to your .bov file.

Attached is a zip file containing .bov descriptor and data files for ints, floats, and doubles.  You can generate your own with the following python code:

[bovFiles.zip](https://github.com/NCAR/VAPOR/files/6556674/bovFiles.zip)

```
import struct

outFile = open("bov.bin", "wb")

for i in range(0,10):
    for j in range(0,10):
        for k in range(0,10):
            #value = float(i%5)
            #value = int(i%2)
            value = float(i/5)
            
            #bytesObj = bytearray(struct.pack("i",value))
            #bytesObj = bytearray(struct.pack("f",value))
            bytesObj = bytearray(struct.pack("d",value))
            outFile.write(bytesObj)

outFile.close()
```


lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+    for (j = 0; j < n; j++) {
+        for (i = 0; i < size / 2; i++) {
+            uc = ucptr[i];
+            ucptr[i] = ucptr[size - i - 1];
+            ucptr[size - i - 1] = uc;
+        }
+        ucptr += size;
+    }
+}
```
shaomeng, line 295: Did you test that this function really change between big endian and small endian? (it doesn't look so for me.)

My suggestion is to not support the endian token at all, since it's actually pretty difficult to switch between the two.

include/vapor/DCBOV.h
```diff
@@ -0,0 +1,175 @@
...
+
+namespace VAPoR {
+
+class BOVCollection;
+
+//!
+//! \class DCBOV
+//! \ingroup Public_VDCBOV
+//!
+//! \brief Class for reading a "Brick of Values"
```
clyne, line 23: Also identify which elements are not supported by this class.

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,368 @@
...
+    // Read a "pencil" of data along the X axis, one row at a time
+    size_t count = max[0] - min[0] + 1;
+    for (int k = min[2]; k <= max[2]; k++) {
+        int zOffset = _dataSize[0] * _dataSize[1] * k;
+        for (int j = min[1]; j <= max[1]; j++) {
+            int xOffset = min[0];
+            int yOffset = _dataSize[0] * j;
+            int offset = formatSize * (xOffset + yOffset + zOffset);
+
+            unsigned char readBuffer[count * formatSize];
```
shaomeng, line 331: Please use a heap allocation instead of a stack allocation here.

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,368 @@
...
+        return (-1);
+    }
+    size_t numValues = _dataSize[0] * _dataSize[1] * _dataSize[2];
+
+    int  n = 1;
+    bool systemLittleEndian = *(char *)&n == 1 ? true : false;
+    bool dataLittleEndian = _dataEndian == "LITTLE" ? true : false;
+    bool needSwap = systemLittleEndian != dataLittleEndian ? true : false;
+
+    // Read a "pencil" of data along the X axis, one row at a time
```
shaomeng, line 322: I believe this way of reading a file will result in an unnecessarily big number of file operation requests. It might be OK on a personal computer, but on a HPC with parallel filesystems, this practice will result in poor performance.

I suggest you read in the whole file using one read request.

include/vapor/DCBOV.h
```diff
@@ -0,0 +1,175 @@
...
+    //!
+    virtual bool variableExists(size_t ts, string varname, int reflevel = 0, int lod = 0) const;
+
+private:
+    VAPoR::UDUnits      _udunits;
+
+    BOVCollection *_bovCollection;
+
+    string                                      _varname;
+    string                                      _proj4String;
```
clyne, line 157: BOV has no support for projection strings, correct? Suggest deleting this and just returning an empty string

include/vapor/DCBOV.h
```diff
@@ -0,0 +1,175 @@
...
+    std::map<string, DC::Dimension>             _dimsMap;
+    std::map<string, DC::CoordVar>              _coordVarsMap;
+    std::map<string, DC::Mesh>                  _meshMap;
+    std::map<string, DC::DataVar>               _dataVarsMap;
+    std::map<string, string>                    _coordVarKeys;
+    std::vector<NetCDFCollection::DerivedVar *> _derivedVars;
```
clyne, line 163: Is this needed? If not, delete.

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,368 @@
...
+            if (rc != count) {
+                if (ferror(fp) != 0) {
+                    MyBase::SetErrMsg("Error reading input file");
+                } else {
+                    MyBase::SetErrMsg("Short read on input file");
+                }
+                return -1;
+            }
+
+            if (_dataFormat == DC::XType::INT32) {
```
shaomeng, line 347: The prompt message claims that data format could be `BYTE` but it doesn't seem you support it here.

I suggest you limit the supported formats to be only float and double to reduce chance of bugs.

include/vapor/DCBOV.h
```diff
@@ -0,0 +1,175 @@
...
+private:
+    VAPoR::UDUnits      _udunits;
+
+    BOVCollection *_bovCollection;
+
+    string                                      _varname;
+    string                                      _proj4String;
+    std::map<string, DC::Dimension>             _dimsMap;
+    std::map<string, DC::CoordVar>              _coordVarsMap;
+    std::map<string, DC::Mesh>                  _meshMap;
```
clyne, line 160: There is no harm in using a Map for the mesh, but since there will only ever be a single mesh you might want to simply keep a copy of the one mesh and not use a map. Just a suggestion

lib/vdc/DCBOV.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+    return (0);
+}
+
+int DCBOV::_InitDimensions()
+{
+    _dimsMap.clear();
+    std::vector<std::string> dimnames = _bovCollection->GetSpatialDimensions();
+    std::vector<size_t>      dimlens = _bovCollection->GetDataSize();
+    VAssert(dimnames.size() == 3);
```
clyne, line 81: Does the BOV format not support 2D data?

shaomeng:
Scott, throwing a thought here: VisIt has BOV support for more than a decade, so the code must be pretty mature and stable. Why don't we just pull that code and plug in here? The benefit will be: 
1) Provide support of all tokens, instead of a subset;
2) Significantly less chance for bugs. Actually, after 10+ yrs, it could even be close to 0 bugs. 
3) It aligns with the spirit of open source development which encourages code reuse.
4) Overall less development time for us considering bug fixes and potentially support other tokens. 

Just something to think about.

lib/vdc/DCBOV.cpp
```diff
@@ -0,0 +1,398 @@
...
+std::vector<string> DCBOV::getCoordVarNames() const
+{
+    map<string, DC::CoordVar>::const_iterator itr;
+
+    vector<string> names;
+    for (itr = _coordVarsMap.begin(); itr != _coordVarsMap.end(); ++itr) { names.push_back(itr->first); }
+    return (names);
+}
+
+template<class T> bool DCBOV::_getAttTemplate(string varname, string attname, T &values) const
```
clyne, line 256: Does the BOV format support attributes? If not why have logic for it?

clyne (commented):
Code looks great, Scott. A few minor in-line nits. A couple things I noted when testing:

1. There does not seem to be support for time varying data or multiple variables. The BOV format supports this, and it is definitely a must for this to be useful. Note, this will entail scanning all of the .bov files, sorting based on TIME, and collecting variables from the same user time together. This is essentially the role of the NetCDFCollection class. 
2. Something is off with the coordinates. I specified a domain size of 0.0 to 10.0 for all dimensions of a 256^3 volume, but the maximum coordinate displayed in the geometry widget was ~9.96

clyne:
> Scott, throwing a thought here: VisIt has BOV support for more than a decade, so the code must be pretty mature and stable. Why don't we just pull that code and plug in here? The benefit will be:
> 
> 1. Provide support of all tokens, instead of a subset;
> 2. Significantly less chance for bugs. Actually, after 10+ yrs, it could even be close to 0 bugs.
> 3. It aligns with the spirit of open source development which encourages code reuse.
> 4. Overall less development time for us considering bug fixes and potentially support other tokens.
> 
> Just something to think about.

I agree with Sam 100% that you should always try to reuse code before you write code. But my experience with anything based on VTK is that the dependencies are often so deep it's really difficult to use their code unless you're embedding it in another VTK application. Still, it doesn't hurt to look. Also, keep in mind that part of the motivation for this reader was to provide something simple that could be used as an exemplar for other developers.

shaomeng:
> 
> I agree with Sam 100% that you should always try to reuse code before you write code. But my experience with anything based on VTK is that the dependencies are often so deep it's really difficult to use their code unless you're embedding it in another VTK application. Still, it doesn't hurt to look. Also, keep in mind that part of the motivation for this reader was to provide something simple that could be used as an exemplar for other developers.

My thought on the VTK thing without closely looking to the VisIT code: there must be some VisIt code that handles things like 1) parsing a text file, 2) reading/writing a raw binary file, 3) basic correctness checks, and so on, before the data is used to construct a VTK object. My suggestion was re-use the code before VTK to construct a Vapor Grid object instead of a VTK object.

lib/vdc/BOVCollection.cpp
shaomeng: Never mind, I got a wrong understanding of big/little endian.

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,368 @@
...
+        }
+
+        if (ss.bad()) {
+            std::string message = "Invalid value for " + token + " in BOV header file.";
+            SetErrMsg(message.c_str());
+            return -1;    // Bad token
+        }
+        return 1;    // we now have a token
+    }
+    return 0;    // we didn't find the token
```
StasJ, line 215: This function has a 3 state return so it would be clearer if the integer values were replaced with a success/error/not_found enum.

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+            if (_dataFormat == DC::XType::INT32) {
+                int *castBuffer = (int *)readBuffer;
+                for (int i = 0; i < count; i++) { *region++ = (typename std::remove_pointer<T>::type)castBuffer[i]; }
+            } else if (_dataFormat == DC::XType::FLOAT) {
+                float *castBuffer = (float *)readBuffer;
+                for (int i = 0; i < count; i++) { *region++ = (typename std::remove_pointer<T>::type)castBuffer[i]; }
+            } else if (_dataFormat == DC::XType::DOUBLE) {
+                double *castBuffer = (double *)readBuffer;
+                for (int i = 0; i < count; i++) { *region++ = (typename std::remove_pointer<T>::type)castBuffer[i]; }
```
StasJ, line 355: What is this code doing?

lib/vdc/BOVCollection.cpp
clyne: @shaomeng raises an interesting point, but the API needs to support the ability to read a **subset** of the data. Doing as Sam suggests would entail reading more data than necessary (in some cases a lot more data), and then copying the requested subset to the destination buffer. This might result in better performance, but modern operating systems are pretty smart about handling buffered I/O (and setvbuf() can be used for fine tuning if needed). I'd suggest convincing oneself that there is a problem before introducing an optimization that might not provide any benefit, at the cost of more complicated code.

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,368 @@
...
+            unsigned char readBuffer[count * formatSize];
+
+            if (needSwap) { _swapBytes(readBuffer, formatSize, numValues); }
+
+            fseek(fp, offset, SEEK_SET);
+            size_t rc = fread(readBuffer, formatSize, count, fp);
+
+            if (rc != count) {
+                if (ferror(fp) != 0) {
+                    MyBase::SetErrMsg("Error reading input file");
```
clyne, line 340: For C library calls that set errno use "%M" in SetErrMsg to get the error message associated with errno:

```suggestion
                    MyBase::SetErrMsg("Error reading input file: %M");
```

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,368 @@
...
+                } else {
+                    MyBase::SetErrMsg("Short read on input file");
```
clyne, line 342:
```suggestion
                    MyBase::SetErrMsg("Short read on input file: %M");
```

lib/vdc/BOVCollection.cpp
shaomeng: @clyne you're asking scott to do something (reading a subset of a data file) that the VisIT BOV specification doesn't support. Scott will need to come up with a scheme to augment the BOV specification to support it. I don't think we should go down that route. (I think we should stick to what BOV already supports which we are supporting a only subset at this point.)

clyne: @shaomeng I'm not sure I understand your comment. The BOV specification is for a file format, not an API. AFAIK it doesn't say anything about how the data should be accessed, only how the data are organized.  In any event, the VAPOR API requires the ability to read a subset of the data. Scott has already implemented this. However, you made the comment that his implementation was inefficient. I'm not sure that it is, and suggested that we not complicate the implementation prematurely for optimization purposes until we know there is a problem.

shaomeng: I see what you mean. I'm OK to deal with it later if there's a performance issue.

include/vapor/DCBOV.h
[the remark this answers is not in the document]
sgpearse: I've added a link to the .pdf file that Sam has added to our repository, https://github.com/NCAR/VAPOR/files/6341067/GettingDataIntoVisIt2.0.0.pdf.

Please let me know if there's more that's needed.

sgpearse: Fixed.

include/vapor/DCBOV.h
sgpearse: Fixed.

include/vapor/DCBOV.h
sgpearse: Fixed.

lib/vdc/BOVCollection.cpp
sgpearse: Fixed.

lib/vdc/BOVCollection.cpp
sgpearse: Interesting.  Fixed.

apps/vaporgui/MainForm.cpp
```diff
@@ -855,6 +858,11 @@ void MainForm::_createFileMenu()
     _dataImportMPAS_Action->setToolTip("Specify one or more MPAS output files to import into the "
                                        "current session");
 
+    _dataImportBOV_Action = new QAction(this);
+    _dataImportBOV_Action->setText(tr("Brick of Values (BOV)"));
+    _dataImportBOV_Action->setToolTip("Specify one BOV data file to import into the "
```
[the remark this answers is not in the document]
sgpearse: Got it.  Fixed.

lib/vdc/BOVCollection.cpp
sgpearse: This function takes a templated array and fills it.  I'm using std::remove_pointer to cast the 'unsigned char' values that are read from the file into the templated array.

I'm not sure if that's your specific question though.  If not, let me know.

lib/vdc/BOVCollection.cpp
sgpearse: Didn't know that.  Fixed.

lib/vdc/BOVCollection.cpp
sgpearse: Fixed.

lib/vdc/DCBOV.cpp
sgpearse: I assumed that the attributes were needed because DC has several pure-virtual functions regarding them.  (DC::getAtt, DC::getAttType, and DC::getAttNames)

Returning empty values (false, empty string, and empty vector) seems to work, so I'm inclined to call this fixed.  But I wonder if it would help other implementors in the future if these functions were not pure-virtual?

lib/vdc/BOVCollection.cpp
sgpearse: Thanks for that.  

Fixed, by describing the required formats as INT, FLOAT, or DOUBLE.  No more BYTE.

include/vapor/DCBOV.h
sgpearse: I'd like to keep the map for now if we want to support time varying BOV files.  Maybe my point is moot, but I want to follow the example given by the other DC classes.  I don't want to deviate from them.

lib/vdc/BOVCollection.cpp
shaomeng: Do you want to delete line 41 which still suggests `byte` ?

lib/vdc/DCBOV.cpp
sgpearse: 3 dimensions need to be described in each BOV file.  A 2D variable could be described by assigning the dimensionality of one axis as 1.  

IE:
   10x10x1

include/vapor/DCBOV.h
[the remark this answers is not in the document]
shaomeng: Do you want to have your own write up of what attributes are supported and what are not? There are almost half attributes in the PDF that are not supported right now.

lib/vdc/BOVCollection.cpp
shaomeng: Did you push new changes? Seems stack allocation still.

include/vapor/DCBOV.h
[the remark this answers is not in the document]
sgpearse: I noted the unsupported data types in another comment below, but I'll also specify which are mandatory, and which are optional.

lib/vdc/BOVCollection.cpp
sgpearse: I think we don't see it here because our comments are attached to the old file.  The permalink below shows the current version of BOVCollection.cpp where a WASP::SmartBuf is being used.

https://github.com/NCAR/VAPOR/blob/c65794fb6d0c2159f111b2cb99544b8a19efa670/lib/vdc/BOVCollection.cpp#L331

lib/vdc/BOVCollection.cpp
sgpearse: I don't think so.  I want to keep all of them for completeness, and have the logic designate BYTE and SHORT as XType::INVALID.  This helps me keep track of what data formats exist in BOV, even though we don't support all of them now.

include/vapor/DCBOV.h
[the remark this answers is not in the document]
clyne: Documentation looks good. Can we add support for BYTE_OFFSET? That's an important one for the reasons stated in the VisIt documentation (Fortran outputs) and is a one liner: simply do an fseek before you read data.

lib/vdc/BOVCollection.cpp
clyne: @sgpearse the SmartBuf pre-dates C++11 and should probably be deprecated. I typically just use a pre-allocated STL vector and access it via the 'data' member. E.g.

```
std::vector<float> buf (buf_size);
float *fptr = buf.data()
```

lib/vdc/DCBOV.cpp
clyne: They are needed, but if there are no attributes to return you can simply return false. There was a lot of logic here earlier that was eventually returning false. 

My thinking on virtual vs pure-virtual is that the latter forces the developer to at least think about. Clarifying the documentation could help I'm sure.

lib/vdc/BOVCollection.cpp
shaomeng: Ok.

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,369 @@
...
+const std::string BOVCollection::_dataBrickletsToken = "DATA_BRICKLETS";
+const std::string BOVCollection::_dataComponentsToken = "DATA_COMPONENTS";
+
+const std::string BOVCollection::_xDim = "x";
+const std::string BOVCollection::_yDim = "y";
+const std::string BOVCollection::_zDim = "z";
+const std::string BOVCollection::_timeDim = "t";
+
+const std::string BOVCollection::_byteFormatString = "BYTE";
+const std::string BOVCollection::_shortFormatString = "SHORT";
```
shaomeng, line 42: These 2 variables are declared, but not used anywhere else. Would it make sense to delete them?

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,369 @@
...
+template<class T> int BOVCollection::ReadRegion(const std::vector<size_t> &min, const std::vector<size_t> &max, T region)
+{
+    FILE *fp = fopen(_dataFile.c_str(), "rb");
+    if (!fp) {
+        SetErrMsg("Invalid file: %d", fp);
+        return -1;
+    }
+
+    size_t formatSize = _sizeOfFormat(_dataFormat);
+    if (formatSize < 0) {
```
shaomeng, line 306: With variable `formatSize` being declared to have type `size_t`, could you give me an example when its value is less than 0 ?

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,369 @@
...
+}
+
+template<class T> int BOVCollection::ReadRegion(const std::vector<size_t> &min, const std::vector<size_t> &max, T region)
+{
+    FILE *fp = fopen(_dataFile.c_str(), "rb");
+    if (!fp) {
+        SetErrMsg("Invalid file: %d", fp);
+        return -1;
+    }
+
```
shaomeng, line 304: There are numerous opportunities in the rest of this function where the function returns and the file is left open (i.e. not closed). Resource leak ain't good.

shaomeng (changes requested):

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,369 @@
...
+    size_t formatSize = _sizeOfFormat(_dataFormat);
+    if (formatSize < 0) {
+        SetErrMsg("Unspecified data format");
+        return (-1);
+    }
+
+    if (_dataSize.size() != 3) {
+        SetErrMsg("Invalid grid size (must be 3D)");
+        return (-1);
+    }
```
shaomeng, line 314: If this thing has to be size 3, why don't you declare it as a fixed sized structure, something like `std::array<float, 3>`? John has to put in a lot of effort to grid classes to deprecate variable length vectors, and I think we should avoid the same mistakes in new code.

shaomeng (changes requested):

lib/vdc/BOVCollection.cpp
sgpearse: Sure, see the following permalink.

https://github.com/NCAR/VAPOR/blob/08fc797d25c09feef2255ff5301755dc62d4b972/lib/vdc/BOVCollection.cpp#L296

lib/vdc/BOVCollection.cpp
sgpearse: For completeness, I'd like to keep all data type identifiers defined in the BOV specification.  This will save me from having to look up the question "what data types are supported in BOV, and how are they defined?" in the future.

lib/vdc/BOVCollection.cpp
sgpearse: That's a good point.  Fixed.

lib/vdc/BOVCollection.cpp
sgpearse: You're correct sir.  Fixed.

lib/vdc/BOVCollection.cpp
shaomeng: Did you push your changes?

lib/vdc/BOVCollection.cpp
sgpearse: No problem.  Fixed.

lib/vdc/BOVCollection.cpp
sgpearse: No problem.  Fixed.

sgpearse:
Comments have been addressed and the BOV reader now supports multiple variables and multiple time steps.  The multi variable/timestep functionality can be tested with the files in the attached archive.  Just import the *.bov files.

Note - you'll need to change the path of the binary data in each brick file.  You can do this with the following command after navigating to the location of the .bov files.

`find *.bov -type f -exec sed -i.bak "s/Users\/pearse/<myBOVDirectory>/g" {} \;`

Note - When rendering variable 'a', set your transfer function range from 1 to 4.

Note - When rendering variable 'b', set your transfer function range from -4 to -1.

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,509 @@
...
+        SetErrMsg("Unspecified data format");
+        fclose(fp);
+        return (-1);
+    }
+
+    if (_gridSize.size() != 3) {
+        SetErrMsg("Invalid grid size (must be 3D)");
+        fclose(fp);
+        return (-1);
+    }
```
shaomeng, line 452: Given the type of variable `_gridSize`, is it possible to not be 3? (Will this condition ever come true?)

shaomeng (changes requested):

lib/vdc/BOVCollection.cpp
sgpearse: An oversight.  Fixed.

include/vapor/DCBOV.h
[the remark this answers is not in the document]
sgpearse: Fixed.

shaomeng:
The resulting rendering appears different from the same data in vdc format. 

Specifically, using the bov file below:
```
TIME: 1.11
DATA_FILE: /glade/work/shaomeng/SPECK2020_Study/rotstrat/X
DATA_SIZE: 1024 1024 1024
DATA_FORMAT: FLOAT
VARIABLE: X
DATA_ENDIAN: LITTLE
CENTERING: zonal
BRICK_ORIGIN: 0. 0. 0.
BRICK_SIZE: 1024. 1024. 1024.
```
you get a volume rendering looking like:
Screenshot from 2021-06-12 12-26-24

While after converting it to VDC format, you get a volume rendering looking like:
Screenshot from 2021-06-12 12-24-43

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,500 @@
...
+}
+
+template<class T> int BOVCollection::ReadRegion(std::string varname, size_t ts, const std::vector<size_t> &min, const std::vector<size_t> &max, T region)
+{
+    float       time = _times[ts];
+    std::string dataFile = _dataFileMap[varname][time];
+
+    FILE *fp = fopen(dataFile.c_str(), "rb");
+    if (!fp) {
+        SetErrMsg("Invalid file: %d", fp);
```
clyne, line 433:
```suggestion
        SetErrMsg("Invalid file: %s : %M", dataFile.c_str());
```

Printing out the null pointer probably won't provide much useful information :-). The %M will pick up errno

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,500 @@
...
+    size_t count = max[0] - min[0] + 1;
+    // Re-use the same buffer for all the pencils.
+    std::vector<unsigned char> vReadBuffer(count * formatSize);
+    unsigned char *            readBuffer = vReadBuffer.data();
+    for (int k = min[2]; k <= max[2]; k++) {
+        int zOffset = _gridSize[0] * _gridSize[1] * k;
+        for (int j = min[1]; j <= max[1]; j++) {
+            int xOffset = min[0];
+            int yOffset = _gridSize[0] * j;
+            int offset = formatSize * (xOffset + yOffset + zOffset);
```
clyne, line 461: Can't you simply add _byteOffset to offset and have only a single seek with SEEK_SET? The code as is will work, but it's a little confusing.

lib/vdc/BOVCollection.cpp
clyne: Is all of this necessary? Can't you simply declare vReadBuffer as type 'T'?

clyne (commented):
Looks good, Scott. One thing that would improve the readability/robustness: the introduction of support for timesteps and variables made the parsing code really complex. The parser is now doing two things (parsing and validating previously parsed values), when it should probably only be doing one (parsing). Let _populateDataFileMap worry about validation across time steps and variables. It could call a validation method each time you add a new variable, or at the end after it has fully constructed _dataFileMap.

lib/vdc/BOVCollection.cpp
sgpearse: True.  Fixed.

lib/vdc/BOVCollection.cpp
sgpearse: Yep.  Fixed.

sgpearse:
@clyne 

> Is all of this necessary? Can't you simply declare vReadBuffer as type 'T'?

I've been spending some time trying to understand why we can't.

When I do as you suggest, double value reads work, but the float and integer readers break.

sgpearse:
@shaomeng thank you for spotting that bug - promoting int to size_t when iterating across data points fixed the problem.

lib/vdc/BOVCollection.cpp
clyne: It would be more helpful to include the file path per my earlier suggestion; the errno message won't have the file path:

SetErrMsg("Invalid file: %s : %M", dataFile.c_str());

clyne:
> @clyne
> 
> > Is all of this necessary? Can't you simply declare vReadBuffer as type 'T'?
> 
> I've been spending some time trying to understand why we can't.
> 
> When I do as you suggest, double value reads work, but the float and integer readers break.

OK. My bad. You're having to do with two types: the format that the data are stored in the file in, and the format requested by the template.

clyne (commented):
Just one more nit, and this [comment](https://github.com/NCAR/VAPOR/pull/2740#pullrequestreview-682434911) still needs to be addressed.

lib/vdc/BOVCollection.cpp
sgpearse: For completeness, I'm adding your quoted text regarding this issue from another comment below.

> OK. My bad. You're having to do with two types: the format that the data are stored in the file in, and the format requested by the template.

lib/vdc/BOVCollection.cpp
sgpearse: No problem.  Fixed.

sgpearse:
@clyne Nit has been fixed, and I broke out the validation code into the function [_validateParsedValues()](https://github.com/NCAR/VAPOR/blob/c7a058303f2fe268cc60718098b7d64ec6094571/lib/vdc/BOVCollection.cpp#L175) to make things easier to understand.

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,503 @@
...
+                    MyBase::SetErrMsg("Short read on input file: %M");
+                }
+                fclose(fp);
+                return -1;
+            }
+
+            if (needSwap) { _swapBytes(readBuffer, formatSize, numValues); }
+
+            // T castBuffer = (T)readBuffer;
+            // for (int i = 0; i < count; i++) { *region++ = (typename std::remove_pointer<T>::type)castBuffer[i]; }
```
shaomeng, line 480: Remove comments?

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,503 @@
...
+            size_t rc = fread(readBuffer, formatSize, count, fp);
+
+            if (rc != count) {
+                if (ferror(fp) != 0) {
+                    MyBase::SetErrMsg("Error reading input file: %M");
+                } else {
+                    MyBase::SetErrMsg("Short read on input file: %M");
+                }
+                fclose(fp);
+                return -1;
```
shaomeng, line 474: Throughout this code, sometime `return ERROR` is used, while some other times `return -1` is used. I suggest we strive for consistency within a module (i.e. BOV module).

lib/vdc/DCBOV.cpp
```diff
@@ -0,0 +1,329 @@
...
+    if (_bovCollection != nullptr) { delete _bovCollection; }
+}
+
+int DCBOV::initialize(const vector<string> &paths, const std::vector<string> &options)
+{
+    _bovCollection = new BOVCollection();
+    int rc = _bovCollection->Initialize(paths);
+    if (rc < 0) {
+        SetErrMsg("Failure reading .bov file");
+        return (-1);
```
shaomeng, line 43: Throughout this code, error is sometime detected by `if (rc == ERROR)`, while other time it's detected in the fashion above. I suggest we strive for consistency within the BOV module.

lib/vdc/DCBOV.cpp
```diff
@@ -0,0 +1,329 @@
...
+    // Set up the coordinate variables
+    //
+    rc = _InitCoordinates();
+    if (rc < 0) { return (-1); }
+
+    // Identify data and coordinate variables. Sets up members:
+    // Initializes members: _dataVarsMap, _meshMap
+    //
+    rc = _InitVars();
+    if (rc < 0) return (-1);
```
shaomeng, line 65: It doesn't look like these 3 functions have any opportunity to fail, is that right?

shaomeng (changes requested):

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,503 @@
...
+std::array<double, 3> BOVCollection::GetBrickSize() const { return _brickSize; }
+
+std::string BOVCollection::GetDataEndian() const { return _dataEndian; }
+
+// Template specialization for reading data of type DC::XType
+template<> int BOVCollection::_findToken<DC::XType>(const std::string &token, std::string &line, DC::XType &value, bool verbose)
+{
+    // Skip comments
+    if (line[0] == '#') { return NOT_FOUND; }
+
```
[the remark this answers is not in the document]
shaomeng: Also, what's the expected and actual behavior if a line with a token ends with a comment:
```
TIME 1.0  # testing a different time stamp
```

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,503 @@
...
+}
+
+int BOVCollection::Initialize(const std::vector<std::string> &paths)
+{
+    VAssert(paths.size() > 0);
+
+    int           rc;
+    std::ifstream header;
+    for (int i = 0; i < paths.size(); i++) {
+        header.open(paths[i]);
```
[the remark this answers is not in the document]
sgpearse: Thanks - fixed.

lib/vdc/BOVCollection.cpp
sgpearse: No problem.  TY.

lib/vdc/BOVCollection.cpp
sgpearse: For readability, Stas made the suggestion for using an enum when these functions have multiple return codes (FOUND, NOT_FOUND, and ERROR), instead of 0, -1, and -2.   These values don't really apply to other functions though.

lib/vdc/DCBOV.cpp
sgpearse: I know what you mean.  The enum is only for parsing though.  Not sure what to do here.

shaomeng: Think about it from a readability perspective. It's easy to under stand if you have all integers:
```c++
int ret = func1();
if( ret != 0 ) { // do something}
```

It's also easy to understand if you have all enum values:
```c++
CODE ret = func2();
if( ret != CODE::GOOD ) { // do something }
```
So a reviewer can know that one's comparing ints to ints, or enums to enums. However, it's quite confusing to mix them up:
```c++
int ret = func3();
if( ret != CODE::GOOD ) { // do something }
```

lib/vdc/BOVCollection.cpp
shaomeng: I think you might want to consider C++ 11 enum classes which provides better safety and readability. 
https://www.geeksforgeeks.org/enum-classes-in-c-and-their-advantage-over-enum-datatype/

lib/vdc/DCBOV.cpp
sgpearse: Hmm you're correct.  An artifact from copying DCCF's logic.  Fixed.

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
sgpearse: Thought I fixed this but nope.  Ty.

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
sgpearse: You are correct.  I've added new logic to handle these cases.  Thanks.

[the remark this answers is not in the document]
sgpearse: Here are my notes from my investigation.  We talked about this at our meeting two weeks ago you agreed not do to it.   Not sure if you remember.

src/databases/BOV/avtBOVFileFormat.C
- A lot of code for parallelized reads.  Would we prune them out?  Comment it?  Ifdef it?
- No initializer lists
- Hard coded strings
- no use of nullptr
- clang-format will mangle it, and I think we would want our copy to be as similar as possible to visit's

avtBOVFileFormat::PopulateDatabaseMetadata
- 150 line long function
- Hard coded strings
- Operates on VTK data objects like avtDatabaseMetadata, avtMeshMetadata
- avtBOVFileFormat::ReadTOC is a 665 line long function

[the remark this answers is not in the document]
shaomeng: yeah, I remember, and I thought it was about the entire BOV module from VisIt. What I'm thinking about now is a smaller module that parses the text file without doing any other things.

It's your ultimately your decision on what approach to take, and I'm only providing a suggestion. My work is to review functionalities of a BOV reader, so I'm simply having a lot of cases in mind (similar to the 2 I've pointed out) that yet to be tested.

lib/vdc/BOVCollection.cpp
sgpearse: Interesting read.  I have no problem with this.  Thanks, and fixed.

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
shaomeng: Good to know, thanks!

lib/vdc/DCBOV.cpp
sgpearse: Hmm.  I agree with you but I think this might be a bigger discussion.  

This is the same thing that's done everywhere else in the application.  If we had an enum class to define GOOD and BAD, wouldn't we want it up higher in the VAPoR namespace, not here?  I think we'd want to make a new rule for it in our coding conventions too.  Is it alright if I file a ticket for this that we can discuss later?

shaomeng: I think you're half way there to achieve a consistent style in the BOV module, so you should finish it. The motivation is that the code currently brings confusion when you compare an integer to an enum. I always need to go look at the enum definition and do a translation between the two values you're comparing in my mind. 

As far as what VAPOR should be doing outside of the scope of the BOV reader, we can certainly discuss later.

sgpearse: OK, but one question: where would you define the enum class?  I haven't investigated this yet.  DCBOV and BOVCollection both need to make use of it.  Should they both have their own identical copies of the enum class?

If we're doing down this route now, we need to consider that every return code in Vapor should be refactored to use it for consistency.  I feel like this is getting outside of the scope of this PR, but if you can make a suggestion I can follow through.

I need to go AFK for a while right now but if you can make a suggestion I can get to it when I return.

shaomeng: I think it should live in the VAPoR namespace so anyone who wants to use it can use it. BOV module is definitely one of the customers. 

Should we expand it to other customers? Probably should. Should it be now? Maybe not. 

This is something I have that you can use as a starting point. Of course you can add/remove items depending on your needs.
```c++
enum class RTNType { // Return Type
    Good,          
    WrongSize,
    IOError,
    InvalidParam,
    BitBudgetMet,
    VersionMismatch,
    ZSTDMismatch,
    ZSTDError,
    DimensionMismatch,
    Error
};
```

lib/vdc/BOVCollection.cpp
shaomeng: Update: Scott decides to wait for input from John before making any change.

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,529 @@
...
+                return -1;
+            }
+
+            // Ensure we have the required tokens in the header
+            //
+            if (_dataFile == _defaultFile) { return _missingValueError(_dataFileToken); }
+            if (_dataFormat == _defaultFormat) { return _missingValueError(_formatToken); }
+            if (_gridSize == _defaultGridSize) { return _missingValueError(_gridSizeToken); }
+
+            std::cout << _dataFile << std::endl;
```
shaomeng, line 106: Is this print statement necessary?

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,529 @@
...
+            return -1;
+        }
+        header.close();
+    }
+
+    return 0;
+}
+
+int BOVCollection::_parseHeader(std::ifstream &header)
+{
```
shaomeng, line 125: Is this parser supposed to correctly recognize scientific notations? If no, please state it in the documentation. If yes, does it?

lib/vdc/DCBOV.cpp
```diff
@@ -0,0 +1,317 @@
...
+    string    timeDim = _bovCollection->GetTimeDimension();
+    size_t    numTimes = _bovCollection->GetUserTimes().size();
+    Dimension dim(timeDim, numTimes);
+    _dimsMap[timeDim] = dim;
+}
+
+void DCBOV::_InitCoordinates()
+{
+    bool                       uniformHint = true;
+    vector<bool>               periodic(false, 4);
```
shaomeng, line 84: Can you 1) state the intend of this vector initialization, and 2) verify that the argument ordering is correct?

shaomeng:
Q to Scott: The VisIt BOV header supports multi-brick usage as described below:
```
For example, you could add “DATA_FILE: file%04d.dat” to the BOV header file to make
VisIt treat all files matching the file%04d.dat pattern as separate domains within the BOV dataset.
```
Is it something that VAPOR BOV parser also supports? If no, it should be stated clearly in a VAPOR BOV documentation.

lib/vdc/DCBOV.cpp
```diff
@@ -0,0 +1,317 @@
...
+    std::array<double, 3>      origin = _bovCollection->GetBrickOrigin();
+    std::array<double, 3>      brickSize = _bovCollection->GetBrickSize();
+
+    // Return spatial coordinate variable values
+    for (int dim = 0; dim < spatialDims.size(); dim++) {
+        if (varname == spatialDims[dim]) {
+            double increment = brickSize[dim] / (dataSize[dim] - 1);
+            double start = origin[dim] + min[0] * increment;
+            size_t steps = max[0] - min[0];
+            for (int i = 0; i <= steps; i++) { region[i] = start + i * increment; }
```
StasJ, line 292: Please make this code clearer by encapsulating the search for a coord variable and the returning of a linear ramp into two separate functions.

lib/vdc/BOVCollection.cpp
```diff
@@ -0,0 +1,529 @@
...
+const std::string BOVCollection::_formatToken = "DATA_FORMAT";
+const std::string BOVCollection::_variableToken = "VARIABLE";
+const std::string BOVCollection::_endianToken = "DATA_ENDIAN";
+const std::string BOVCollection::_centeringToken = "CENTERING";
+const std::string BOVCollection::_originToken = "BRICK_ORIGIN";
+const std::string BOVCollection::_brickSizeToken = "BRICK_SIZE";
+const std::string BOVCollection::_offsetToken = "BYTE_OFFSET";
+const std::string BOVCollection::_divideBrickToken = "DIVIDE_BRICK";
+const std::string BOVCollection::_dataBrickletsToken = "DATA_BRICKLETS";
+const std::string BOVCollection::_dataComponentsToken = "DATA_COMPONENTS";
```
StasJ, line 34: I believe making the token names follow variable naming conventions makes the code less readable. For example in `_findToken(_divideBrickToken, line, _divideBrick);` you cannot tell what is happening without knowing that the variables that end in token are the static tokens whereas `_findToken(TOKEN_DIVIDE_BRICK, line, _divideBrick);` is much clearer.

StasJ (changes requested):

lib/vdc/DCBOV.cpp
sgpearse: 1) to configure a non-periodic coordinate variable, 2) It's not, 3) Not sure.

Thanks for the catch.  Fixed.

shaomeng: If the intent is to create a vector of size 4, and in reality a vector of size 0 is created which is the case here, and the rest of the program behaves as if a vector of size 4 is created, then that almost certainly suggests that another bug exists somewhere. I encourage you to investigate thoroughly and rule out other potential bugs.

lib/vdc/DCBOV.cpp
sgpearse: No problem.  Fixed.

lib/vdc/BOVCollection.cpp
sgpearse: Fixed.

lib/vdc/BOVCollection.cpp
sgpearse: Yes, it does.

lib/vdc/BOVCollection.cpp
sgpearse: Fixed.

sgpearse:
@shaomeng It does not support wildcards.  Only one data file per .bov file as of now.  I've added it to the documentation.

shaomeng:
> @shaomeng It does not support wildcards. Only one data file per .bov file as of now. I've added it to the documentation.

Could you provide a write up of the specifications in PDF or ReadTheDocs? The idea is that when someone wants to write a BOV header, he/she needs to have something to refer to. Having the specification in the source code is not very user-friendly for someone to refer to.

sgpearse:
@shaomeng Yes, I'm working on that now but it's on a different branch.  Not a part of this PR, which is targeting main.

lib/vdc/DCBOV.cpp
sgpearse: Fixed in BOVCollection.

StasJ (approved):

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
shaomeng: Q to Scott, is the current code able to handle this case?
```TIME 1.0  # testing a different time stamp```

File: lib/vdc/BOVCollection.cpp (hunk 24 above)
Comment (shaomeng) [reply lib/vdc/BOVCollection.cpp]: Q to Scott, is this really fixed? I.e., is it consistent to use all integers or all enums?

File: lib/vdc/DCBOV.cpp [hunk 34]
```diff
@@ -0,0 +1,334 @@
...
+    string    timeDim = _bovCollection->GetTimeDimension();
+    size_t    numTimes = _bovCollection->GetUserTimes().size();
+    Dimension dim(timeDim, numTimes);
+    _dimsMap[timeDim] = dim;
+}
+
+void DCBOV::_InitCoordinates()
+{
+    bool                       uniformHint = true;
+    vector<bool>               periodic(false);
```
shaomeng, line 84: Suggest avoid confusion with the following line: 
```suggestion
    vector<bool>               periodic;
```

shaomeng (changes requested):

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
sgpearse: Yes, works for me.

lib/vdc/DCBOV.cpp
shaomeng: Scott, this suggestion isn't the same as what you wanted to achieve that you verbally described. Treat the suggestion with caution.

clyne: The suggested change should work fine for the way it is being used in the code. Though I might suggest:

`vector <bool> periodic(1,false)`

or 

`vector <bool> periodic = {false}
`
so that it is completely clear how the variable is being initialized (without having to consult the documentation). But no big deal

clyne (approved):

shaomeng (approved):

PR #2749, opened 2021-06-09
Refactor Animation and Viewpoint tab #2732

### LINKED ISSUE #2715: Camera controls do not update the scene
### Describe the bug
When I type values in any of the "camera control" fields nothing happens. I would expect that the view would update accordingly. 

- [ ] Did VAPOR crash?

**Impact**

### To Reproduce
1. Open a file 
2. type value in the camera control fields 
3. nothing happens 

*Additionally, please attach a session file to this ticket to easily reproduce (use File->Save Session).*

### Expected behavior
I would expect the visualizer view to be updated 

### Desktop
 - OS and version: Macos Mojave 
 - Version: [e.g. Vapor 3.4]

### LINKED ISSUE #2732: Refactor Animation and Viewpoint tab

StasJ:
Fix #2732
Fix #2715


clyne:
@StasJ please resolve conflicts and CI failures, and I will assign reviewers.

StasJ:
> @StasJ please resolve conflicts and CI failures, and I will assign reviewers.

Done!

clyne (commented):
Looks great, but we lost animation looping. Maybe put it on the title bar?

StasJ:
> Looks great, but we lost animation looping. Maybe put it on the title bar?

I added it back as an option in the animation tab. The title bar needs refactoring so I think we should do that before adding features to it in its current state.

apps/vaporgui/MainForm.cpp
```diff
@@ -731,7 +736,7 @@ void MainForm::_createVizToolBar()
     connect(_viewAllAction, SIGNAL(triggered()), _tabMgr, SLOT(ViewAll()));
     connect(_sethomeAction, SIGNAL(triggered()), _tabMgr, SLOT(SetHomeViewpoint()));
     connect(_alignViewCombo, SIGNAL(activated(int)), _tabMgr, SLOT(AlignView(int)));
-    connect(_viewRegionAction, SIGNAL(triggered()), _tabMgr, SLOT(CenterSubRegion()));
+    //    connect(_viewRegionAction, SIGNAL(triggered()), _tabMgr, SLOT(CenterSubRegion()));
```
sgpearse, line 739: Do we need to keep this?

apps/vaporgui/MainForm.cpp
```diff
@@ -642,12 +645,14 @@ void MainForm::_createAnimationToolBar()
...
-    connect(_stepBackAction, SIGNAL(triggered()), _tabMgr, SLOT(AnimationStepBackward()));
-    connect(_timeStepEdit, SIGNAL(returnPressed()), this, SLOT(_setTimeStep()));
+    // clang-format off
+    connect(_playForwardAction,  SIGNAL(triggered()),     _animationController, SLOT(AnimationPlayForward()));
+    connect(_playBackwardAction, SIGNAL(triggered()),     _animationController, SLOT(AnimationPlayReverse()));
+    connect(_pauseAction,        SIGNAL(triggered()),     _animationController, SLOT(AnimationPause()));
+    connect(_stepForwardAction,  SIGNAL(triggered()),     _animationController, SLOT(AnimationStepForward()));
+    connect(_stepBackAction,     SIGNAL(triggered()),     _animationController, SLOT(AnimationStepReverse()));
+    connect(_timeStepEdit,       SIGNAL(returnPressed()), this, SLOT(_setTimeStep()));
+    // clang-format on
```
sgpearse, line 655: It's a shame that clang-format doesn't seem to have an option that lets us align our code like this.  Just commenting.  If there's an option that I've missed, maybe we can consider using it.

apps/vaporgui/AnimationTab.cpp
```diff
@@ -0,0 +1,45 @@
...
+                         new PLineItem("Current Timestep", new PTimestepInput(_controlExec)),
+                         new PTotalTimestepsDisplay(_controlExec),
+                     }),
+        new PSection("Animation",
+                     {
+                         _startTS = new PIntegerInput(AnimationParams::_startTimestepTag, "Animation Start Timestep"),
+                         _stopTS = new PIntegerInput(AnimationParams::_endTimestepTag, "Animation Last Timestep"),
+                         new PCheckbox(AnimationParams::_repeatTag, "Loop Animation Playback"),
+                         (new PIntegerInput(AnimationParams::_stepSizeTag, "Animation Play Step Size"))->SetRange(1, 10),
+                         (new PDoubleSliderEdit(AnimationParams::_maxRateTag, "Max Animation Frames Per Second"))->SetRange(1, 60),
```
sgpearse, line 24: Is this a new option?  Also, shouldn't an FPS controller be in PIntSliderEdit?

apps/vaporgui/NavigationUtils.cpp
```diff
@@ -0,0 +1,403 @@
...
+            if (!rParams.size()) continue;
+
+            size_t local_ts = dataStatus->MapGlobalToLocalTimeStep(dataSetNames[j], ts);
+
+            for (int k = 0; k < rParams.size(); k++) {
+                RenderParams *rp = rParams[k];
+                rp->SetCurrentTimestep(local_ts);
+
+                // If variable was not initialzed but now is, reset the TF mapping range
+                std::string varNames[2] = {rp->GetVariableName(), rp->GetColorMapVariableName()};
```
sgpearse, line 249: Should this be using rp->GetActualColorMapVariableName()?

apps/vaporgui/NavigationUtils.cpp
```diff
@@ -0,0 +1,403 @@
...
+
+
+AnimationParams *NavigationUtils::GetAnimationParams(ControlExec *ce) { return ((AnimationParams *)ce->GetParamsMgr()->GetParams(AnimationParams::GetClassType())); }
+
+
+
+// This is some code that someone decided is worth keeping for future reference.
+// It allegedly worked at some point.
+// I don't want to have to write any camera code without refactoring the entire
+// system so I am leaving it here in case the functionality is needed again.
```
sgpearse, line 310: @clyne is it ok for us to remove VAPOR3_0_0_ALPHA code at this point?

sgpearse (commented):
Great work as always Stas.  Just a few questions/comments.

I think there may be a regression with undo/redo in the PLineEdit.  If it's not a regression, we can just file a bug report and fix it in a different issue.

To reproduce:
1) Load any dataset and modify the scale factor in the Viewpoint tab
2) Undo, and nothing happens
3) Move the scene, and do 2 undo's which will revert the move, and then revert the scale

apps/vaporgui/NavigationUtils.cpp
clyne: I'd say so. We still have lot of detritus like this throughout the code base that needs to be clean up. But it seems reasonable to remove this now.

clyne (approved):
Looks good!

apps/vaporgui/MainForm.cpp
StasJ: No and I removed it, thanks. This refactor involved thousands of changes so some things slipped through.

apps/vaporgui/AnimationTab.cpp
StasJ: No, it was in the old GUI too. The params database stores it as a double so that is why I used the double widget.

apps/vaporgui/NavigationUtils.cpp
StasJ: In this case it wouldn't really matter but I changed it to `GetActualColorMapVariableName`. Also, I did not write any of this code, it is legacy code that I moved around to clean up other parts of the GUI.

StasJ:
> Great work as always Stas. Just a few questions/comments.
> 
> I think there may be a regression with undo/redo in the PLineEdit. If it's not a regression, we can just file a bug report and fix it in a different issue.
> 
> To reproduce:
> 
> 1. Load any dataset and modify the scale factor in the Viewpoint tab
> 2. Undo, and nothing happens
> 3. Move the scene, and do 2 undo's which will revert the move, and then revert the scale

It is not a regression. This has to do with text edit boxes capturing ctrl-z when they are in focus.

apps/vaporgui/NavigationUtils.cpp
StasJ: I would like to leave this because, as the comment says, the camera code is quite a mess and I think we should keep it around until the camera backend as a whole is refactored. I can change the `ifdef` title if that makes its purpose more clear.

apps/vaporgui/MainForm.cpp
sgpearse: No problem.  I know how that goes.

apps/vaporgui/NavigationUtils.cpp
sgpearse: Ok, let's not worry about it now then.

sgpearse (approved):

PR #2752, opened 2021-06-10
Introduce new grid method: Grid::GetNumDimensions()

shaomeng:
This PR decouples two pieces of information that's contained by the return value of the old `grid::GetDimentions()` method. More specifically, the number of (useful) dimensions will be returned by `GetNumDimensions()` which will be an integer of 1, 2, 3, etc. The number of elements along each dimension will be returned by the new `GetDimensions()` method of a type `std::array<size_t, 3>`.

Partially addressses #2250


include/vapor/Grid.h
```diff
@@ -1195,6 +1199,8 @@ class VDF_API Grid {
         }
     }
 
+    double               _defaultZ = 0.0;
```
StasJ, line 1202: There is already a convention for the default Z and it is not always zero so this may break some things.

include/vapor/Grid.h
```diff
@@ -88,6 +88,10 @@ class VDF_API Grid {
     //!
     const std::vector<size_t> &GetDimensions() const { return (_dims); }
 
+    //! Get and set the default Z value for a grid.
+    auto GetDefaultZ() const -> double {return _defaultZ;};
+    void SetDefaultZ( double d)        {_defaultZ = d;};
```
StasJ, line 93: You have two different techniques of defining return values right next to each other

include/vapor/Grid.h
shaomeng: This PR is still in draft status, but feedback is welcome. Can you elaborate on the existing convention?

include/vapor/Grid.h
shaomeng: Not sure what you mean, but one of these 2 functions has no return value.

include/vapor/Grid.h
StasJ: John requested my review. Check out Renderer::GetDefaultZ

shaomeng: I see. Let me check it out.

clyne: Sorry. I didn't see that it was marked draft. We can pause the review.

shaomeng: There's no defaultZ in grids anymore.

clyne (commented):
Sam, this looks good. The one change that I'd like to see before completing the review is to add a method to the Grid class:

size_t Grid::GetNumDimensions() 

that returns the number of dimensions (basically, it returns Grid._dims.size()).

Then we can eliminate all of the new code that looks like:

```
auto tmp = GetDimensions();
auto tmp_size = std::count_if(tmp.begin(), tmp.end(), [](size_t v) { return v != 1; });
```

with

GetNumDimensions(). This would be more consistent with Grid::GetTopologyDim() and GetGeometryDim(). I guess if we wanted to be consistent the new function should be named "GetDimensionsDim()", but that may be confusing?

The other somewhat significant change we should probably make is to have Grid::GetCoordDimensions() return a std::array<size_t, 3> instead of a vector.

Thoughts?

shaomeng:
> Sam, this looks good. The one change that I'd like to see before completing the review is to add a method to the Grid class:
> 
> size_t Grid::GetNumDimensions()
> 
> that returns the number of dimensions (basically, it returns Grid._dims.size()).

I think this is a good idea, and I'll go ahead and make the changes.
> 
> 
> GetNumDimensions(). This would be more consistent with Grid::GetTopologyDim() and GetGeometryDim(). I guess if we wanted to be consistent the new function should be named "GetDimensionsDim()", but that may be confusing?

I agree that `GetDimensionsDim` sounds confusing, so I'm good with the previously suggested name.
> 
> The other somewhat significant change we should probably make is to have Grid::GetCoordDimensions() return a std::array<size_t, 3> instead of a vector.

There are numerous methods in Grid that returns a vector (including the one that you pointed out), and ideally they are all refactored to return a fixed sized array. But I don't know if they should be touched now. I consider a more appropriate refactoring effort is to touch the grid itself so the internal representations of a grid are in a fixed size, so the interface is naturally fixed sized too. 
> 
> Thoughts?

clyne (commented):
@shaomeng, everything looks good except for one performance related issue. There are a number of places in the code where statements like:

`const vector<size_t> &dims = g->GetDimensions();
`
have been replaced with:

```
auto tmp = g->GetDimensions();
auto dims = std::vector<size_t>{tmp[0], tmp[1], tmp[2]};
dims.resize(g->GetNumDimensions());
```

For code that is not performance sensitive this is fine. But for performance-sensitive code the replaced code introduces heap memory allocation and memory copies. An alternative that would have less performance impact would be something like:

```
const auto &dims = g->GetDimensions();
size_t nDims = g->GetNumDimensions()
```

and then replace the use of `dims.size()` with `nDims`

Note that the use of the reference above (&dims) should flag any code that is performance sensitive. 

Lastly, we should probably run a benchmark to compare the performance between the old and new code. Scott's `testGrid` should do the trick. Something like:

`./bin/testGrid -dims 256:256:256`

Be sure to compile in release mode.

StasJ (commented):
DefaultZ is still used in a few places while another comment states it is no longer used.

include/vapor/Grid.h
StasJ: void is a return type.

shaomeng:
> DefaultZ is still used in a few places while another comment states it is no longer used.

This PR doesn't use any DefaultZ. There are usage of it but not touched by this PR. I'm not sure if those use cases belong to the scope of this PR.

lib/vdc/StretchedGrid.cpp
```diff
@@ -188,7 +180,9 @@ StretchedGrid::ConstCoordItrSG::ConstCoordItrSG() : ConstCoordItrAbstract()
 
 void StretchedGrid::ConstCoordItrSG::next()
 {
-    const vector<size_t> &dims = _sg->GetDimensions();
+    auto tmp = _sg->GetDimensions();
```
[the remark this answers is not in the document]
shaomeng: This, and other similar occurrences, have been fixed in 372082e and eb97e89 .

include/vapor/Grid.h
shaomeng: I'm not sure if you're suggesting a change or not. It's just not valid syntax to write something like `auto func() -> void`.

StasJ: `auto func() -> void` is valid syntax, however the syntax is inconsistent with 99% of Vapor's code base so I am suggesting we don't use it.

shaomeng: 1) are you sure? 2) well, any new language feature will be 99% inconsistent with existing code base. 

It's a topic that can be discussed to whether add to the VAPOR style convention a certain syntax, but currently the style convention doesn't say one way or another.

clyne: @shaomeng will change and we will update the convention

shaomeng:
@clyne I have verified that the overhead does come from data copies and data copies alone. No additional dynamic allocations.

One potential improvement is to make `_dims` not private in `Grid` so that the following copies could be eliminated. They amount to the biggest overhead.
https://github.com/NCAR/VAPOR/blob/6ff6e111d4bac51065248211bdf7208488345337/include/vapor/StructuredGrid.h#L73-L80

clyne:
@shaomeng have you merged main with your branch lately?

shaomeng:
> @shaomeng have you merged main with your branch lately?

At the time you asked, it was 3 commits behind. At the time you see this message, it's up-to-date.

clyne:
> @clyne I have verified that the overhead does come from data copies and data copies alone. No additional dynamic allocations.
> 
> One potential improvement is to make `_dims` not private in `Grid` so that the following copies could be eliminated. They amount to the biggest overhead.
> https://github.com/NCAR/VAPOR/blob/6ff6e111d4bac51065248211bdf7208488345337/include/vapor/StructuredGrid.h#L73-L80

How about we simply declare _dims as:

`Size_tArr3           _dims = {{1, 1, 1}};
`
This would make it consistent with the members of the Grid class that have been changed from std::vector to std::array. It looks like only two lines of code would need to be changed in Grid.cpp. Note that the Grid iterators are also defined in Grid.cpp, and they have their own _dims member, but this is different from Grid::_dims. I.e. we would not have to change the Iterator _dims.

shaomeng:
> How about we simply declare _dims as:
> 
> `Size_tArr3 _dims = {{1, 1, 1}}; `
> This would make it consistent with the members of the Grid class that have been changed from std::vector to std::array. It looks like only two lines of code would need to be changed in Grid.cpp. Note that the Grid iterators are also defined in Grid.cpp, and they have their own _dims member, but this is different from Grid::_dims. I.e. we would not have to change the Iterator _dims.

Well, I have tried this tactic earlier, and there are way more road blocks than two lines of code. I had to back off. 
Further more, even this change is pushed through, it won't eliminate the copies that I identified above, as long as the function signature of GetNodeDimensions() remains the same. 
Further further more, I also attempted to change the function signature of GetNodeDimensions(), and there are also unexpected roadblocks. 

I think we should just accept the memory copy overhead for now---that's the cost of applying bandaids.

StasJ (approved):

clyne (approved):

PR #2767, opened 2021-07-01
Fixes #2784 - testGrid reports error yet it still passes CircleCI 

### LINKED ISSUE #2764: testGrid reports error yet it still passes CircleCI
While manually execute the following command yields the following error, CircleCI passes.

./bin/testGrid -dims 80:80:80

```
*** Error reported in Regular grid ***

  Interpolation order: linear          
    RMS error:                                           0.0000
    Missing value count:                                 0
    GetValueAtIndex() vs GetValue() disagreement count:  0
    Time:                                                0.0000

*** Error reported in Regular grid ***

  Interpolation order: nearestNeighbor 
    RMS error:                                           0.0000
    Missing value count:                                 0
    GetValueAtIndex() vs GetValue() disagreement count:  0
    Time:                                                0.0000
```

### LINKED ISSUE #2784: BOV reader doesn't understand relative path names
The BOV importer doesn't understand relative file path names specified by the DATA_FILE keyword in a .bov file. If the value of DATA_FILE is not an absolute path name it should be interpreted relative to the path to the .bov file that contains the DATA_FILE keyword.

To reproduce:
1. Copy data/Source/BOV/256f.{bin,bov} to another location
2. Edit 256f.bov, changing the value of DATA_FILE to a relative path, e.g.:
    DATA_FILE: ./256f.bin

or simply

    DATA_FILE: 256.bin

3. Start vapor, import the .bov file and try to render the data.

clyne:
Fixes #2764 

This draft PR partially address #2764; it replaces an exact floating point compare (which should not have been used) with an approximate floating point compare. This prevents the 'testGrid' from reporting an error. However, there is still a problem with the python script that calls testGrid, failing to detect the non-zero exit code.

### DIFF
--- a/test_apps/smokeTests/smokeTests.py
+++ b/test_apps/smokeTests/smokeTests.py
@@ -198,6 +198,8 @@ def main():
     print()
     makeBaseline = args.makeBaseline
 
+    rc = 0
+
     if ( args.makeBaseline == False and makeBaseline == True ): 
         print( "    Warning: Some or all baseline files for running DataMgr testswere missing.  "
                "    These files are needed as comparisons for the results of the current series "
@@ -212,7 +214,7 @@ def main():
         if ( testGrid(grid) != 0):
             print ("  See artifact file " + grid + ".txt or " + resultsDir + grid + ".txt for mismatches")
             print ("  Failed assertions, if any, are shown above.\n" )
-            sys.exit(-1)
+            rc = 1
  
     for dataType, dataFile in dataMgrs.items():
         baselineFile = resultsDir + dataType + "_baseline.txt"
@@ -223,10 +225,11 @@ def main():
     dataMgr = testDataMgrs( makeBaseline )
     if ( dataMgr != 0 ):
         print( "DataMgr tests failed.  Results are not identical to the baseline." )
-        exit(-1);
+        rc = 1
     else:
         print( "DataMgr tests passed" )
-        exit(0);
+
+    sys.exit(rc)
 
 if __name__ == "__main__":
     main()
--- a/test_apps/smokeTests/smokeTests.py
+++ b/test_apps/smokeTests/smokeTests.py
@@ -209,10 +209,10 @@ def main():
         os.mkdir( resultsDir )
 
     for grid in gridSizes:
-        if ( testGrid(grid) > 0):
+        if ( testGrid(grid) != 0):
             print ("  See artifact file " + grid + ".txt or " + resultsDir + grid + ".txt for mismatches")
             print ("  Failed assertions, if any, are shown above.\n" )
-            #sys.exit(1)
+            sys.exit(-1)
  
     for dataType, dataFile in dataMgrs.items():
         baselineFile = resultsDir + dataType + "_baseline.txt"
--- a/test_apps/smokeTests/gridTools.cpp
+++ b/test_apps/smokeTests/gridTools.cpp
@@ -191,7 +191,7 @@ bool CompareIndexToCoords(VAPoR::Grid *grid,
 
     rms = sqrt(sum / (x * y * z));
 
-    if ((! Wasp::NearlyEqual(rms, 0.0)) || disagreements > 0) rc = false;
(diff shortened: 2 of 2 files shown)


sgpearse:
@clyne the non-zero exit code is being tested on [line 125 in smokeTests.py](https://github.com/NCAR/VAPOR/blob/8f473e81a8a5c0f4401098adeb02ffc7609803e6/test_apps/smokeTests/smokeTests.py#L125).  I'm not able to reproduce what you're reporting.  Could you provide instructions on how to reproduce the failure you're seeing?

clyne:
You'll need to ask @shaomeng for clarification. He was the one that reported the problem.  But make sure testGrid is actually generated a non-zero exit code.

shaomeng:
Hi Scott, when you manually execute the test command reported in issue #2764 , do you see the reported output?

sgpearse:
@shaomeng The smoke tests are passing on circleCI because we only test the grids listed below.  We can't test every permutation, so your 80x80x80 example indicates a new bug in the DataMgr.

`2x2x2
4x2x2
8x2x2
7x7x7
8x8x8`

sgpearse:
@clyne EXIT_SUCCESS and EXIT_FAILURE are correctly being reported by testGrid, so I don't think there's another problem to fix here.

clyne:
> @clyne EXIT_SUCCESS and EXIT_FAILURE are correctly being reported by testGrid, so I don't think there's another problem to fix here.

OK. I think I see the disconnect here. @shaomeng thought that CircleCI was failing to report an error condition. However, @shaomeng was testing a configuration (-dims 80x80x80) that was not being tested by the test driver running on CircleCI. As @sgpearse points out, it's not practical to run all permutations of grid sizes. I do not believe that there is "another problem to fix here", only miscommunication. @shaomeng can you verify that I've summarized the events correctly? I think we can review this PR, which fixes the 80x80x80 grid problem, and move on.

sgpearse:
You guys were right.  We were not exiting -1 on failed grid tests.  This is fixed in the above commit.

sgpearse (approved):

shaomeng (approved):
Looking good, thanks for fixing it @clyne @sgpearse

test_apps/smokeTests/smokeTests.py
```diff
@@ -209,10 +209,10 @@ def main():
         os.mkdir( resultsDir )
 
     for grid in gridSizes:
-        if ( testGrid(grid) > 0):
+        if ( testGrid(grid) != 0):
             print ("  See artifact file " + grid + ".txt or " + resultsDir + grid + ".txt for mismatches")
             print ("  Failed assertions, if any, are shown above.\n" )
-            #sys.exit(1)
+            sys.exit(-1)
```
clyne, line 215: It's a small nit, but this should exit with a 1, not a negative 1. There is some discussion on exit codes here:

https://en.wikipedia.org/wiki/Exit_status

A little edification: the convention for unix system calls and the C standard library is  for functions to return 0 on success and a negative int (usually -1, but often not specified) on failure. This is the convention that the VAPOR API follows.

However, the *exit* codes of a Unix process are positive integers, and the convention is to return 0 on success, but a value >0 on failure. A '1' represents a generic error, other values have specified meanings.  Hence, we typically use '1' for non-normal process termination.

Lastly, there are two ways to gracefully exit a C/C++ program: calling exit() from anywhere, or returning from main(). Calling exit() will prevent C++ destructors from running, and should typically be avoided. Hence, main() should have a return statement that provides an appropriate exit code. Unfortunately, the convention when returning from main with a general error is to do 'return(1)', NOT return(-1), because exit codes are positive ints.

sgpearse: Interesting.  It looks like [Python follows roughly the same convention as C++](https://docs.python.org/3/library/sys.html#sys.exit).  It looks like it tries to do cleanup as well so it might be safer to use than in C++.

I've made the fix and will push it once my sanity test finishes.

sgpearse: Fixed.

Issue #2765, opened 2021-06-30
complete refactoring of Grid class


clyne:
The Grid class needs some minor refactoring to complete the conversion from the use of std::vector to std::array for the representation of fixed-size, arrays of length 3. In particular, a number of public getters on the grid class still return std::vector instead of std::array. This results in unnecessary, and sometimes expensive, conversions between the types.

shaomeng:
Well, in the proposed refactoring work, one should also move all the function implementations to the .cpp file, instead of leaving them in the .h file. That's because:
1) the current headers are often more than 1000 lines long, kinda ridiculous.
2) those function implementations in the header will be compiled every time the header is included, which means compiled hundreds of times throughout VAPOR, greatly increasing the compilation time and binary size.
3) separation between headers and implementations complies with the coding convention.

clyne:
This has nearly been completed as of 3.6. I think we should defer further action until we have a more extensive testing framework, enabled by the upcoming Python API

PR #2769, opened 2021-07-01
fix #2768

### LINKED ISSUE #2768: Regression introduced by #2752
There appears to be a regression introduced by #2752 when invoking the wireframe renderer. 

To reproduce:
1. Load /glade/p/cisl/vast/vapor/data/Source/DKRZ-Ocean/gr15/gr15_uvw_vis.nc
2. Enable a wireframe renderer. vapor will crash

shaomeng:
This PR fixes issue #2768.

The effort spent here, together with effort spent on PR #2752 , will turn out to be completely useless once issue #2765 is properly addressed. This exercise demonstrates the wasteful nature of addressing issues half-way by applying band-aids, and it's something we should avoid in the future.

### DIFF
--- a/include/vapor/ConstantGrid.h
+++ b/include/vapor/ConstantGrid.h
@@ -50,23 +50,23 @@ private:
     // They do nothing and return meaningless values.
     // Do not use!
     //
-    std::vector<size_t>        GetCoordDimensions(size_t) const override;
-    size_t                     GetGeometryDim() const override;
+    std::vector<size_t>         GetCoordDimensions(size_t) const override;
+    size_t                      GetGeometryDim() const override;
     const std::array<size_t, 3> GetNodeDimensions() const override;
-    const size_t               GetNumNodeDimensions() const override;
-    const std::vector<size_t> &GetCellDimensions() const override;
-    void                       GetBoundingBox(const Size_tArr3 &min, const Size_tArr3 &max, DblArr3 &minu, DblArr3 &maxu) const override {}
-    bool                       GetEnclosingRegion(const DblArr3 &minu, const DblArr3 &maxu, Size_tArr3 &min, Size_tArr3 &max) const override { return (false); }
-    virtual void               GetUserCoordinates(const Size_tArr3 &, DblArr3 &) const override {}
-    bool                       GetIndicesCell(const DblArr3 &coords, Size_tArr3 &indices) const override;
-    bool                       GetCellNodes(const Size_tArr3 &, std::vector<Size_tArr3> &) const override;
-    bool                       GetCellNeighbors(const Size_tArr3 &, std::vector<Size_tArr3> &) const override;
-    bool                       GetNodeCells(const Size_tArr3 &, std::vector<Size_tArr3> &) const override;
-    size_t                     GetMaxVertexPerFace() const override;
-    size_t                     GetMaxVertexPerCell() const override;
-    void                       ClampCoord(const DblArr3 &coords, DblArr3 &cCoords) const override { cCoords = coords; }
-    ConstCoordItr              ConstCoordBegin() const override;
-    ConstCoordItr              ConstCoordEnd() const override;
+    const size_t                GetNumNodeDimensions() const override;
+    const std::vector<size_t> & GetCellDimensions() const override;
+    void                        GetBoundingBox(const Size_tArr3 &min, const Size_tArr3 &max, DblArr3 &minu, DblArr3 &maxu) const override {}
+    bool                        GetEnclosingRegion(const DblArr3 &minu, const DblArr3 &maxu, Size_tArr3 &min, Size_tArr3 &max) const override { return (false); }
+    virtual void                GetUserCoordinates(const Size_tArr3 &, DblArr3 &) const override {}
+    bool                        GetIndicesCell(const DblArr3 &coords, Size_tArr3 &indices) const override;
+    bool                        GetCellNodes(const Size_tArr3 &, std::vector<Size_tArr3> &) const override;
+    bool                        GetCellNeighbors(const Size_tArr3 &, std::vector<Size_tArr3> &) const override;
+    bool                        GetNodeCells(const Size_tArr3 &, std::vector<Size_tArr3> &) const override;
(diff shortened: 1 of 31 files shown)


lib/vdc/ConstantGrid.cpp
```diff
@@ -39,14 +39,14 @@ std::vector<size_t> ConstantGrid::GetCoordDimensions(size_t) const
 
 size_t ConstantGrid::GetGeometryDim() const { return 3; }
 
-const std::vector<size_t> &ConstantGrid::GetNodeDimensions() const
+const std::array<size_t, 3> ConstantGrid::GetNodeDimensions() const
```
clyne, line 42: Please use Size_tArr3 for the type.

include/vapor/ConstantGrid.h
```diff
@@ -50,30 +50,29 @@ class VDF_API ConstantGrid : public Grid {
     // They do nothing and return meaningless values.
     // Do not use!
     //
-    std::vector<size_t>        GetCoordDimensions(size_t) const override;
```
clyne, line 53: Any idea why all of this code changed?

clyne (commented):
A couple of nits

include/vapor/Grid.h
```diff
@@ -1209,6 +1210,16 @@ class VDF_API Grid {
         }
     }
 
+    virtual void ClampIndex(const std::array<size_t, 3> &dims, size_t nNodeDims, const Size_tArr3 indices, Size_tArr3 &cIndices) const
+    {
+        cIndices = {0, 0, 0};
+
+        for (int i = 0; i < nNodeDims; i++) {
```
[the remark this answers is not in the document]
shaomeng: well, I don't understand the logic of the task performed here, so I wrote it in a way that's equivalent to the old code. If you can tell me that it will do the right thing even when the length is fixed to 3, then I can definitely remove the length parameter.

include/vapor/ConstantGrid.h
shaomeng: clang-format did it

lib/vdc/ConstantGrid.cpp
shaomeng: Well, this alias is so confusing---it *appears* to be using both camel case and underscore conventions simultaneously within 10 letters. How about we use the C++ standard name for this PR and wait the alias to be fixed to be less confusing later?

include/vapor/Grid.h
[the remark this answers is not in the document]
clyne: It will do the right thing, but you might want to  look at the code and the documentation to convince yourself.

lib/vdc/ConstantGrid.cpp
clyne: @shaomeng I'd like the code to be consistent. We have a developer-defined type for all of these 3-element array arguments, please use it. If you really have a problem with the naming of the type you are more than welcome to change it.

include/vapor/ConstantGrid.h
clyne: OK. Any idea why? I.e. our entire code base was supposed to be consistently formatted with clang-format. Do you have any idea why clang-format decided to re-format this unchanged code now?

include/vapor/Grid.h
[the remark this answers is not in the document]
shaomeng: Got it, fixed in 52081bac1

lib/vdc/ConstantGrid.cpp
shaomeng: Ok, I've renamed the alias in 52081bac1 with two of them: `DimsType` and `CoordType`. Two notes:
1. It's common practice to have a word "type" at the end of all alias, so when people see it people know it's an alias that redefines something else.
2. It's also common to name alias after its semantic meaning, (Dims and Coord in this example), so the following function signature is very easy to understand:
```c++
void function( DimsType, CoordType );

File: lib/vdc/StructuredGrid.cpp [hunk 4]
```diff
@@ -34,6 +34,10 @@ StructuredGrid::StructuredGrid(const vector<size_t> &dims, const vector<size_t>
     }
 }
 
+const std::array<size_t, 3> StructuredGrid::GetNodeDimensions() const { return GetDimensions(); }
```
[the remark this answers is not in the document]
shaomeng: fixed in 52081bac1

include/vapor/ConstantGrid.h
shaomeng: There was one function signature change, which resulted a longer line. Clang-format tries to align all function names, so it shifted all function signatures, resulting the changes we see here.

include/vapor/UnstructuredGrid.h
```diff
@@ -127,23 +127,24 @@ class VDF_API UnstructuredGrid : public Grid {
...
+    virtual bool GetNodeCells(const DimsType &indices, std::vector<DimsType> &cells) const override;
 
     size_t GetMaxVertexPerFace() const override { return (_maxVertexPerFace); };
 
     size_t GetMaxVertexPerCell() const override { return ((GetTopologyDim() == 3) ? 2 * GetMaxVertexPerFace() : GetMaxVertexPerFace()); };
 
     //! Return the grid node dimmensions
     //!
-    const std::vector<size_t> &GetNodeDimensions() const override { return (_vertexDims); }
+    const std::array<size_t, 3> GetNodeDimensions() const override;
```
clyne, line 146: Shouldn't this be DimsType?

shaomeng: You're right. Fixed this one together with a few others that were missed.

lib/vdc/ConstantGrid.cpp
clyne: That's fine by me. FYI it is also common practice to designate types with "_t" suffix. E.g. size_t, uint_t, etc.

include/vapor/ConstantGrid.h
clyne: Got it. Thanks. Not sure I like that feature, but I can live with it :-)

clyne (approved):

Issue #2739, opened 2021-05-26
Test and build platforms need to be configured to enable OpenMP


clyne:
Currently, our test and build platforms are not set up to compile VAPOR with OpenMP enabled.

sgpearse:
@clyne A few questions:

1) Do we want to enable our USE_OMP flag for our release builds?  It's currently defaulted to OFF.
2) Do we want to enable it for both debug and release CI tests, on Ubuntu and CentOS?  Or, do we want to have a new set of tests where we build with it?  Right now we test for compilation and warnings in release and debug mode.  Would the following set of compilation/warning tests fit our bill?

Test compilation/warnings on:
- Ubuntu/CentOS
- Debug/Release-mode
- USE_OMP ON, and OFF

clyne:
Yes, I believe we want to enable USE_OMP for all platforms that we are able (Mac OS X and Linux). Note for Mac this will require installing the LLVM clang compiler ourselves. The Apple version of the clang compiler does NOT support OpenMP.

We should do our Circle CI testing with USE_OMP  enabled, both for debug and release mode. I don't see any reason to also test with OpenMP disabled at this time. 


CC'ing @shaomeng and @stasj for their input.

shaomeng:
From a test completeness perspective, as long as the build configuration supports *both* enable and disable modes, *both* modes should be tested.

StasJ:
I agree with John that there isn't much reason to test with OpenMP disabled since we are planning to ship all builds with it enabled.

clyne:
We won't be able to enable OpenMP support on Windows. So @shaomeng is probably right; we should build&test with and without OpenMP support.

clyne:
Copy-n-paste email thread on this topic below:

Interesting, do we have a macos circle CI set up?
 
Yes, circle is building the installers on all platforms.

Thanks for the doc.  It helps a lot.

A couple questions:

1) Should we add omp to our third party libraries for OSX?
2) If we're distributing our installers with OpenMP, should we make the settings in Sam's document the default in CMakeLists.txt?
3) Before following sam's instructions, I was getting the error below which is due to the use of c++14 features.  Are we still on c++11 with the omp changes?

In file included from /Users/pearse/VAPOR/apps/vaporgui/AnimationController.h:3:
In file included from /usr/local/VAPOR-Deps/2019-Aug/Qt/5.13.2/clang_64/lib/QtCore.framework/Headers/QTimer:1:
[...]

PR #2770, opened 2021-07-01
Issue2739 - Enable OMP on circleCI tests

sgpearse:
This adds OMP compile and warning tests in debug and release mode to CircleCI.

We will now test the following
- Debug compilation with and without OMP
- Debug warnings with and without OMP
- Release compilation with and without OMP
- Release warnings with and without OMP

Note:  This PR also removes a printf statement that slipped into the repo, and resets the smokeTest baseline files. to omit the print output.

Note: #2739 asks for us to enable OpenMP for our CI tests, as well as our builds.  This PR only applies CI tests.  Our installer builds are done on a different branch, named _installers_.  I'll file a separate PR for that branch that enables OpenMP on OSX, Ubuntu, and CentOS.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -249,7 +249,6 @@ jobs:
           name: make release with omp
           command: |
             cd /root/project/build
-            #cmake -DCMAKE_CXX_COMPILER=/usr/local/opt/llvm/bin/clang++ -DCMAKE_C_COMPILER=/usr/local/opt/llvm/bin/clang -DCMAKE_CXX_FLAGS=-I/usr/local/opt/llvm/include -DCMAKE_EXE_LINKER_FLAGS=-L/usr/local/opt/llvm/lib -DBUILD_TEST_APPS=ON -DUSE_OMP=ON ..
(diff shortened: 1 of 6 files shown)


clyne (approved):

StasJ (approved):
This may be unrelated to the PR but I thought I'd mention that the smoke test results are in our repo and in this PR they appear to have quadrupled in run time since the last time they were pushed.

clyne:
> This may be unrelated to the PR but I thought I'd mention that the smoke test results are in our repo and in this PR they appear to have quadrupled in run time since the last time they were pushed.

Can you quantify the run time? We're definitely increasing the amount of testing. If the run time is becoming unbearable we can look at possibly making some changes.

StasJ:
> > This may be unrelated to the PR but I thought I'd mention that the smoke test results are in our repo and in this PR they appear to have quadrupled in run time since the last time they were pushed.
> 
> Can you quantify the run time? We're definitely increasing the amount of testing. If the run time is becoming unbearable we can look at possibly making some changes.

For CF the runtime doubled from 0.15 to 0.3. The WRF and VDC runtimes quadrupled from 17 to 74.

PR #2776, opened 2021-07-06
Fixed #2763 - unit test timer broken on Linux

### LINKED ISSUE #2763: unit test timer broken on Linux
On linux, the test app always reports time being 0.

To reproduce:
On cisl-vapor, compile the code in release mode and run the following command:
./bin/testGrid -dims 80:80:80
Notice that the time reported is always zero.

clyne:
Fixes #2763. 

@stasj is "__linux__" the right macro to use?


StasJ (approved):
`__linux__` is the correct macro, yes.

PR #2779, opened 2021-07-07
Install OpenMP

StasJ:
Adds OpenMP to the Vapor installer by creating a wrapper for the CMake install command which upon generating an installer automatically creates a new copy of the binaries, finds and changes the link to the OpenMP libs, and then installs that in place of the original.

Relevant for #2739


CMakeLists.txt
```diff
@@ -172,6 +173,10 @@ elseif (APPLE)
 		set (INSTALL_INCLUDE_DIR include/vapor)
 	endif ()
 	set (CMAKE_INSTALL_RPATH "@executable_path/../lib")
+    # set (INSTALL_NAME_DIR "${CMAKE_INSTALL_PREFIX}/lib")
```
clyne, line 176: Can this commented out code be removed?

clyne (commented):
@stasj, just a couple of quick sanity checks:

1. Does this change bundle the OpenMP runtime on both Linux and Mac?
2. Is there a way to disable the bundling for the OpenMP runtime? I.e.
2a. can we build installers on a system that doesn't have OpenMP support?
2b. Even if build system has OpenMP support will the installer exclude the OpenMP runtime if USE_OMP is off?

shaomeng (approved):
Besides the same remove comments feedback, I have no additional feedback, because I have no idea what this PR is doing. I guess we'll just test and see if it works.

CMakeLists.txt
StasJ: I was still doing some testing with this code. I ended up uncommenting it and only enabling it for distributable builds.

StasJ:
> @StasJ, just a couple of quick sanity checks:
> 
> 1. Does this change bundle the OpenMP runtime on both Linux and Mac?

Yes

> 2. Is there a way to disable the bundling for the OpenMP runtime? I.e.
>    2a. can we build installers on a system that doesn't have OpenMP support?

Yes

>    2b. Even if build system has OpenMP support will the installer exclude the OpenMP runtime if USE_OMP is off?

Yes

clyne (approved):

PR #2782, opened 2021-07-08
Fix bad initialization

### LINKED ISSUE #2760: Current pre-release installer crashes on CentOS
Our pre-release installer for 3.4.1 currently crashes on startup on CentOS

bash-4.2$ bin/vapor
vapor[0x45666f]
/lib64/libc.so.6(+0x36400)[0x2ac715414400]
/glade/u/apps/dav/opt/gnu/9.1.0/lib64/libstdc++.so.6(_ZNSsC2ERKSs+0xb)[0x2ac714bd028b]
vapor[0x4fb99d]
vapor[0x4fe289]
vapor[0x4fe888]
vapor[0x522540]
vapor[0x45f64d]
vapor[0x4b6dd4]
vapor[0x4b7606]
vapor[0x4a3c23]
vapor[0x4344df]
/lib64/libc.so.6(__libc_start_main+0xf5)[0x2ac715400555]
vapor[0x436854]

shaomeng:
Fix #2760 crash problem.

### DIFF
--- a/apps/vaporgui/PDatasetTransformWidget.cpp
+++ b/apps/vaporgui/PDatasetTransformWidget.cpp
@@ -12,7 +12,7 @@ using namespace VAPoR;
 PDatasetTransformWidget::PDatasetTransformWidget(VAPoR::ControlExec *ce)
 : PWidget("", _section = new VSectionGroup("Transform",
                                            {
-                                               _twDataset = new PStringDropdown(SelectedDatasetTag, {}, "Dataset"),
+                                               _twDataset = new PStringDropdown("transformWidgetDatasetTag", {}, "Dataset"),
                                                _tw = new PTransformWidget,
                                            })),
   _ce(ce)
--- a/apps/vaporgui/PDatasetTransformWidget.cpp
+++ b/apps/vaporgui/PDatasetTransformWidget.cpp
@@ -8,11 +8,12 @@
 
 using namespace VAPoR;
 
+const std::string PDatasetTransformWidget::SelectedDatasetTag = "transformWidgetDatasetTag";
 
 PDatasetTransformWidget::PDatasetTransformWidget(VAPoR::ControlExec *ce)
 : PWidget("", _section = new VSectionGroup("Transform",
                                            {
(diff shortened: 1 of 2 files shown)


apps/vaporgui/PDatasetTransformWidget.cpp
```diff
@@ -12,7 +12,7 @@ using namespace VAPoR;
 PDatasetTransformWidget::PDatasetTransformWidget(VAPoR::ControlExec *ce)
 : PWidget("", _section = new VSectionGroup("Transform",
                                            {
-                                               _twDataset = new PStringDropdown(SelectedDatasetTag, {}, "Dataset"),
+                                               _twDataset = new PStringDropdown("transformWidgetDatasetTag", {}, "Dataset"),
```
clyne, line 15: @shaomeng this fixes the problem and that is awesome! Nice job. But how about defining SelectedDatasetTag in a way that it will be properly initialized rather than hard-coding the string here? E.g. SelectedDatasetTag could be defined as a static member of PDatasetTransformWidget, or simply as an anonymous namespace variable within the PDatasetTransformWidget.cpp file? I think the former is what we do with all of the params classes.

shaomeng: Well, I don't see other static data member usage across the family of PWidgets. Now I've found the problem and I'd rater defer the whatever proper fix to @StasJ .

shaomeng: A note on static data members: Google C++ Guide simply bans such usage if the data member involves dynamic memory allocation. In our case, a static string would simply not be allowed.

clyne: How about simply using constexpr to force compile time initialization then: https://en.cppreference.com/w/cpp/language/constexpr

shaomeng: It seems only static data members can be declared as a constepr...

I was thinking that there are probably hundreds of PWidgets of all sorts, and there must be a mechanism that Stas is already using in those widgets.

clyne (approved):

StasJ (approved):

PR #2807, opened 2021-07-14
Fixed Issue 2785 - Wireframe renderer broken with CAM data

### LINKED ISSUE #2785: Wireframe renderer broken with CAM data
A regression has been introduced that causes the WireFrame renderer to assert.

To reproduce:
1. Load CAM/AdamPhillips/F2000C5_new120_025/FV_768x1152.bilinear.F2000C5_new120_025.cam.h0.2004-01.nc
2. Create a wireframe renderer, selected 2D variables, and enable

vapor terminates with:

VAssert(coords.size() == dims.size()) failed in utils.cpp, line 45
Terminated

clyne:
Fixed #2785 

Refactored Grid iterators to use std::array instead of std:vector, making the iterators consistent with the rest of the Grid API

### DIFF
--- a/lib/render/ContourRenderer.cpp
+++ b/lib/render/ContourRenderer.cpp
@@ -116,7 +116,11 @@ int ContourRenderer::_buildCache()
     double mv = grid->GetMissingValue();
     float  Z0 = GetDefaultZ(_dataMgr, _cacheParams.ts);
 
-    Grid::ConstCellIterator it = grid->ConstCellBegin(_cacheParams.boxMin, _cacheParams.boxMax);
+    CoordType boxMin = {0.0, 0.0, 0.0};
+    CoordType boxMax = {0.0, 0.0, 0.0};
+    Grid::CopyToArr3(_cacheParams.boxMin, boxMin);
+    Grid::CopyToArr3(_cacheParams.boxMax, boxMax);
+    Grid::ConstCellIterator it = grid->ConstCellBegin(boxMin, boxMax);
 
     size_t           maxNodes = grid->GetMaxVertexPerCell();
     vector<DimsType> nodes(maxNodes);
@@ -126,8 +130,8 @@ int ContourRenderer::_buildCache()
 
     Grid::ConstCellIterator end = grid->ConstCellEnd();
     for (; it != end; ++it) {
-        const vector<size_t> &cell = *it;
-        grid->GetCellNodes(cell.data(), nodes);
+        const DimsType &cell = *it;
+        grid->GetCellNodes(cell, nodes);
 
         bool hasMissing = false;
         for (int i = 0; i < nodes.size(); i++) {
--- a/lib/render/ContourRenderer.cpp
+++ b/lib/render/ContourRenderer.cpp
@@ -116,11 +116,7 @@ int ContourRenderer::_buildCache()
     double mv = grid->GetMissingValue();
     float  Z0 = GetDefaultZ(_dataMgr, _cacheParams.ts);
 
-    CoordType boxMin = {0.0, 0.0, 0.0};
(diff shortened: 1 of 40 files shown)


lib/vdc/CurvilinearGrid.cpp
```diff
@@ -237,14 +237,14 @@ CurvilinearGrid::ConstCoordItrCG::ConstCoordItrCG(const CurvilinearGrid *cg, boo
...
 
     if (ndims == 3) {
         if (_terrainFollowing) {
-            _coords.push_back(*_zCoordItr);
+            _coords[2] = *_zCoordItr;
         } else {
-            _coords.push_back(_cg->_zcoords[0]);
+            _coords[2] = _cg->_zcoords[0];
         }
     }
```
shaomeng, line 249: Should give _coords[2] a default value in condition of ndims != 3, right?

shaomeng (commented):
Just a small improvement needed :)

lib/vdc/CurvilinearGrid.cpp
clyne: Fixed

lib/vdc/CurvilinearGrid.cpp
```diff
@@ -314,28 +315,28 @@ void CurvilinearGrid::ConstCoordItrCG::next()
...
     auto ndims = _cg->GetNumDimensions();
 
-    vector<size_t> maxIndex;
-    maxIndex.reserve(3);
+    if (!ndims) return;
+
+    DimsType maxIndex = {0, 0, 0};
 
-    for (int i = 0; i < ndims; i++) maxIndex.push_back(dims[i] - 1);
+    for (int i = 0; i < ndims; i++) maxIndex[i] = dims[i] - 1;
```
shaomeng, line 326: There are many instances where `dims[i] - 1` is assigned to another variable of unsigned int type. There's a danger when `dims[i]` holds 0, then nobody knows what `dims[i] - 1` will be. 

I think based on the knowledge of how likely that situation can occur, asserts can be inserted in some instances.

shaomeng (approved):

lib/vdc/CurvilinearGrid.cpp
shaomeng: Though this is something not related to this PR, so feel free to not deal with it.

lib/render/ContourRenderer.cpp
```diff
@@ -116,7 +116,11 @@ int ContourRenderer::_buildCache()
     double mv = grid->GetMissingValue();
     float  Z0 = GetDefaultZ(_dataMgr, _cacheParams.ts);
 
-    Grid::ConstCellIterator it = grid->ConstCellBegin(_cacheParams.boxMin, _cacheParams.boxMax);
+    CoordType boxMin = {0.0, 0.0, 0.0};
+    CoordType boxMax = {0.0, 0.0, 0.0};
+    Grid::CopyToArr3(_cacheParams.boxMin, boxMin);
+    Grid::CopyToArr3(_cacheParams.boxMax, boxMax);
+    Grid::ConstCellIterator it = grid->ConstCellBegin(boxMin, boxMax);
```
StasJ, line 123: Instead of having to repeat this process every time we want to get a grid, couldn't `ConstCellBegin` be overloaded to support vectors? Especially since `Box` still returns vectors.

lib/vdc/CurvilinearGrid.cpp
clyne: You can never have to many asserts IMHO :-). Done.

lib/render/ContourRenderer.cpp
clyne: Indeed. Done.

StasJ (approved):

Issue #2799, opened 2021-07-13
Top time step control doesn't honor max time step settings


shaomeng:
To reproduce:
1. Open VAPOR, load 18 time steps from the WRF->Katrina data set.
2. Create a wireframe renderer
3. Go to Scene->Animation panel, set the "Animation Last Timestep" control to be 10.
4. Notice that within Scene->Animation panel, all knobs honor the new setting, i.e., no knob can go beyond time step 10.
5. Notice that the top time step panel does not honor this setting: i.e., if you type 15, it accepts 15 (as the screenshot shows).

![image](https://user-images.githubusercontent.com/814885/125376530-6feebf00-e348-11eb-88c7-4237a3134fe9.png)

StasJ:
I believe this is the correct behavior. The animation control panel is for controlling the animation playback. The timestep input in the top bar is for setting the current timestep. Those are separate functionalities. If you play the animation from the top bar, it will respect the parameters in the animation control panel.

Issue #2791, opened 2021-07-12
BOV reader semi-support time steps


shaomeng:
To reproduce:
1. Open VAPOR, import BOV file: `/glade/p/cisl/vast/vapor/Bugs/BOV_Bugs/bug3.bov`
2. Create and enable a volume renderer. 
3. Advance time step by 1. Notice that the rendering remains the same, though the 2 time steps clearly point to different data files.

sgpearse:
Visit's BOV specification doesn't mention specifying multiple time steps like this.  IMO, this is an enhancement.

BOV Specification: section 3, page 9
https://wci.llnl.gov/sites/wci/files/2020-08/GettingDataIntoVisIt2.0.0.pdf

clyne:
@shaomeng are you sure the file you created (bug3.bov) is how time-varying data sets are supported? My interpretation of VisIt's documentation is that you can only have one set of keyword-value pairs per file. I.e. You need a single file per time step or per variable. I could be wrong, however.

shaomeng:
@clyne VAPOR shows 2 time steps available with the bov file in question, and a user is able to switch between these 2 time steps. It's OK to not show 2 time steps if you don't support 2 time steps. 

Re VisIt comparison: I feel we're in a state where when VAPOR behaves differently from the VisIt and its documentation, it's considered a VAPOR uniqueness. When VAPOR exhibits a bug (or bad behavior in general) and VisIt behaves the same, it's considered "see, VisIt does the same." This isn't a good situation to be in. A mitigation is to

clyne:
Agreed that it should not show two time steps. I think a reasonable behavior for repeating the same keyword multiple times is to simply accept the last value found.

sgpearse:
@shaomeng, you are observing #2812 here, which is not BOV related.  If you import one timestep of a wrf file, you can advance the timesteps as well.

Issue #2790, opened 2021-07-12
BOV reader hangs


shaomeng:
To Reproduce:
1. Open VAPOR, import bov file: `/glade/p/cisl/vast/vapor/Bugs/BOV_Bugs/bug2.bov`
2. Create a volume renderer. VAPOR hangs.

PR #2809, opened 2021-07-14
Issue2798 - BOV Reader

### LINKED ISSUE #2788: BOV reader crashes
To Reproduce:
1. Open VAPOR, and import BOV file at: `/glade/p/cisl/vast/vapor/Bugs/BOV_Bugs/bug1.bov`
2. Create a volume renderer. VAPOR crashes. 

Note: this crash is observed on `cisl-vapor` but not on `casper`.

### LINKED ISSUE #2789: Cannot read BOV
1. Open 256d.bov
2. Create a renderer

Vapor with throw 5 error popups (which should be consolidated) and not render anything.
System: macOS

### LINKED ISSUE #2792: BOV lacks very basic error checking
### Describe the bug
1. Open VAPOR, import BOV file `/glade/p/cisl/vast/vapor/Bugs/BOV_Bugs/bug4.bov`
2. Create and enable a volume renderer, something gets rendered.

However, this shouldn't happen, because the BOV header claims a certain data size (128x512x256) and data format (double) but the data file is only half of the required size. This means, VAPOR read in twice the amount of data that the data file provides and rendered something. Not good...

### LINKED ISSUE #2797: BOV reader complains on legal BOV file
To Reproduce:
1. On `casper`, open vapor, import BOV file: `/glade/p/cisl/vast/vapor/Bugs/BOV_Bugs/bug5.bov`
2. Notice that VAPOR will complain that cannot open file (without a specific reason).

However, this file appears perfectly legal to me, unless I'm missing something.

### LINKED ISSUE #2798: BOV reader false alarm
Test Platform:  `casper`

To Reproduce:
1. Open VAPOR without passing in any command line argument
2. Use menu File->Import->BOV to open file`/glade/p/cisl/vast/vapor/Bugs/BOV_Bugs/bug6.bov`
3. Notice that VAPOR will complain this file and provide a reason: `Invalid value for token: DATA_ENDIAN
The keyword DATA_ENDIAN should only contain one value.`
5. However, if you use vim to look at this file in question, it does contain only one value for the DATA_ENDIAN token.

sgpearse:
Fixes #2798 and fixes #2797 - bug6.bov and bug5.bov
Fixed by trimming white space after values are read, at lines [386](https://github.com/NCAR/VAPOR/blob/a4f4924ad15112a439c27429b1911faa96b1f904/lib/vdc/BOVCollection.cpp#L386), [420](https://github.com/NCAR/VAPOR/blob/a4f4924ad15112a439c27429b1911faa96b1f904/lib/vdc/BOVCollection.cpp#L420), and [463](https://github.com/NCAR/VAPOR/blob/a4f4924ad15112a439c27429b1911faa96b1f904/lib/vdc/BOVCollection.cpp#L463) in the _findToken template functions.

Fixes #2792 and #2790- bug4.bov and bug2.bov
Fixed [by enforcing the DATA_ENDIAN value](https://github.com/NCAR/VAPOR/blob/a4f4924ad15112a439c27429b1911faa96b1f904/lib/vdc/BOVCollection.cpp#L250) to be either BIG or LITTLE, and adding [a sanity check](https://github.com/NCAR/VAPOR/blob/a4f4924ad15112a439c27429b1911faa96b1f904/lib/vdc/BOVCollection.cpp#L267) that compares the grid size * data format size, to the size of the binary file being loaded.

@shaomeng The sanity check is in place, but I believe your bug report has the correctly sized data file for the header description.  128x512x256x8(double) = 134,217,728 bytes, which is the size of /glade/p/cisl/vast/vapor/Bugs/BOV_Bugs/Bx.bot.

#2791 Is not a bug - bug3.bov
This regards #2799, where the maximum timestep of DataStatus is not honored.  FYI, the last values reported in a BOV file are registered as "true values".  If you define TIME twice, the second one will overwrite the first one.  **This will be documented.** 

Fixes #2788 - bug1.bov
Fixed by [correcting the count](https://github.com/NCAR/VAPOR/blob/a4f4924ad15112a439c27429b1911faa96b1f904/lib/vdc/BOVCollection.cpp#L583) of bytes passed into _swapBytes()

Fixes #2789 and #2784 - Relative file paths
Fixed by prepending the BOV header file path to the data file path [when realpath() could not be evaluated](https://github.com/NCAR/VAPOR/blob/a4f4924ad15112a439c27429b1911faa96b1f904/lib/vdc/BOVCollection.cpp#L200) for the data file.

### DIFF
--- a/include/vapor/DCBOV.h
+++ b/include/vapor/DCBOV.h
@@ -24,23 +24,23 @@ class BOVCollection;
 // # DATA_FILE points to a binary data file.  It can be a full file path, or a path relative to the BOV header.
 // DATA_FILE: bovA1.bin
 //
-// # The data size corresponds to NX,NY,NZ in the above example code.
+// # The data size corresponds to NX,NY,NZ in the above example code.  It must contain three values
 // DATA_SIZE: 10 10 10
 //
 // # Allowable values for DATA_FORMAT are: INT,FLOAT,DOUBLE
 // DATA_FORMAT: FLOAT
 //
-// # VARIABLE is a string that specifies the variable being read in DATA_FILE
+// # VARIABLE is a string that specifies the variable being read in DATA_FILE.  Must be alphanumeric (abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890)
 // VARIABLE: myVariable
 //
-// # BRICK_ORIGIN lets you specify a new coordinate system origin for # the mesh that will be created to suit your data.
+// # BRICK_ORIGIN lets you specify a new coordinate system origin for # the mesh that will be created to suit your data.  It must contain three values.
 // BRICK_ORIGIN: 0. 0. 0.
 //
-// # BRICK_SIZE lets you specify the size of the brick.
+// # BRICK_SIZE lets you specify the size of the brick on X, Y, and Z.  It must contain three values.
 // BRICK_SIZE: 10. 20. 5.
 //
 // # BYTE_OFFSET is optional and lets you specify some number of
-// # bytes to skip at the front of the file. This can be useful for # skipping the 4-byte header that Fortran tends to write to files. # If your file does no
+// # bytes to skip at the front of the file. This can be useful for # skipping the 4-byte header that Fortran tends to write to files. # If your file does not have a header then DO NOT USE BYTE_OFFSET.
 // BYTE_OFFSET: 4
 
 //!
@@ -71,10 +71,12 @@ class BOVCollection;
 //! - DATA_COMPONENTS
 //!
 //! If duplicate key/value pairs exist in a BOV header, the value closest to the bottom of the file will be used.
+//! If duplicate values exist for whatever reason, all entries must be valid (except for DATA_FILE, which gets validated after parsing)
 //! Scientific notation is supported for floating point values like BRICK_ORIGIN and BRICK_SIZE.
 //! Scientific notation is not supported for integer values like DATA_SIZE.
 //! Wild card characters are not currently supported in the DATA_FILE token.
 //! Each .bov file can only refer to a single data file.
+//! VARIABLE must be alphanumeric (abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890)
 //!
 //! \author Scott Pearse
 //! \date    May, 2021
--- a/include/vapor/DCBOV.h
+++ b/include/vapor/DCBOV.h
@@ -31,11 +31,11 @@ class BOVCollection;
 //! The following BOV tags are optional:
 //! - BRICK_ORIGIN
 //! - BRICK_SIZE
-//! - DATA_ENDIAN
 //! - TIME
 //! - VARIABLE
 //!
 //! The following BOV tags are currently unsupported:
+//! - DATA_ENDIAN
 //! - CENTERING
 //! - BYTE_OFFSET
 //! - DIVIDE_BRICK
--- a/include/vapor/DCBOV.h
+++ b/include/vapor/DCBOV.h
@@ -29,15 +29,15 @@ class BOVCollection;
 //! - DATA_FORMAT
 //!
 //! The following BOV tags are optional:
-//! - BRICK_ORIGIN
-//! - BRICK_SIZE
-//! - TIME
-//! - VARIABLE
+//! - BRICK_ORIGIN (default: 0., 0., 0.)
+//! - BRICK_SIZE   (default: 1., 1., 1.)
+//! - TIME         (default: 0.)
+//! - VARIABLE     (default: "brickVar")
+//! - BYTE_OFFSET  (default: 0)
 //!
 //! The following BOV tags are currently unsupported:
 //! - DATA_ENDIAN
 //! - CENTERING
-//! - BYTE_OFFSET
 //! - DIVIDE_BRICK
 //! - DATA_BRICKLETS
 //! - DATA_COMPONENTS
--- a/include/vapor/DCBOV.h
+++ b/include/vapor/DCBOV.h
@@ -30,7 +30,7 @@ class BOVCollection;
 // # Allowable values for DATA_FORMAT are: INT,FLOAT,DOUBLE
 // DATA_FORMAT: FLOAT
 //
-// # VARIABLE is a string that specifies the variable being read in DATA_FILE.  Must be alphanumeric (abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890)
+// # VARIABLE is a string that specifies the variable being read in DATA_FILE.  Must be alphanumeric (abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_-)
 // VARIABLE: myVariable
 //
 // # BRICK_ORIGIN lets you specify a new coordinate system origin for # the mesh that will be created to suit your data.  It must contain three values.
@@ -40,8 +40,8 @@ class BOVCollection;
 // BRICK_SIZE: 10. 20. 5.
 //
 // # BYTE_OFFSET is optional and lets you specify some number of
-// # bytes to skip at the front of the file. This can be useful for # skipping the 4-byte header that Fortran tends to write to files. # If your file does not have a header then DO NOT USE BYTE_OFFSET.
-// BYTE_OFFSET: 4
+// # bytes to skip at the front of the file. This can be useful for # skipping the 4-byte header that Fortran tends to write to files. # If your file does not have a header then DO NOT USE
+// BYTE_OFFSET. BYTE_OFFSET: 4
 
 //!
 //! \class DCBOV
@@ -76,7 +76,7 @@ class BOVCollection;
 //! Scientific notation is not supported for integer values like DATA_SIZE.
 //! Wild card characters are not currently supported in the DATA_FILE token.
 //! Each .bov file can only refer to a single data file.
-//! VARIABLE must be alphanumeric (abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890)
+//! VARIABLE must be alphanumeric (abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_-)
 //!
 //! \author Scott Pearse
 //! \date    May, 2021
--- a/include/vapor/DCBOV.h
+++ b/include/vapor/DCBOV.h
@@ -16,6 +16,33 @@ namespace VAPoR {
 
 class BOVCollection;
 
+// Example BOV header file, bovA1.bov
+//
+// # TIME is a floating point value specifying the timestep being read in DATA_FILE
+// TIME: 1.1
+//
+// # DATA_FILE points to a binary data file.  It can be a full file path, or a path relative to the BOV header.
+// DATA_FILE: bovA1.bin
+//
+// # The data size corresponds to NX,NY,NZ in the above example code.
+// DATA_SIZE: 10 10 10
+//
+// # Allowable values for DATA_FORMAT are: INT,FLOAT,DOUBLE
+// DATA_FORMAT: FLOAT
+//
+// # VARIABLE is a string that specifies the variable being read in DATA_FILE
+// VARIABLE: myVariable
+//
+// # BRICK_ORIGIN lets you specify a new coordinate system origin for # the mesh that will be created to suit your data.
+// BRICK_ORIGIN: 0. 0. 0.
+//
+// # BRICK_SIZE lets you specify the size of the brick.
+// BRICK_SIZE: 10. 20. 5.
+//
+// # BYTE_OFFSET is optional and lets you specify some number of
+// # bytes to skip at the front of the file. This can be useful for # skipping the 4-byte header that Fortran tends to write to files. # If your file does no
+// BYTE_OFFSET: 4
+
 //!
 //! \class DCBOV
 //! \ingroup Public_VDCBOV
@@ -35,13 +62,15 @@ class BOVCollection;
 //! - VARIABLE     (default: "brickVar")
 //! - BYTE_OFFSET  (default: 0)
 //!
-//! The following BOV tags are currently unsupported:
+//! The following BOV tags are currently unsupported.  They can be included in a BOV header,
+//! but they will be unused.
 //! - DATA_ENDIAN
 //! - CENTERING
 //! - DIVIDE_BRICK
 //! - DATA_BRICKLETS
 //! - DATA_COMPONENTS
 //!
+//! If duplicate key/value pairs exist in a BOV header, the value closest to the bottom of the file will be used.
 //! Scientific notation is supported for floating point values like BRICK_ORIGIN and BRICK_SIZE.
 //! Scientific notation is not supported for integer values like DATA_SIZE.
 //! Wild card characters are not currently supported in the DATA_FILE token.
--- a/include/vapor/DCBOV.h
+++ b/include/vapor/DCBOV.h
@@ -30,7 +30,7 @@ class BOVCollection;
 // # Allowable values for DATA_FORMAT are: INT,FLOAT,DOUBLE
 // DATA_FORMAT: FLOAT
 //
-// # VARIABLE is a string that specifies the variable being read in DATA_FILE.  Must be alphanumeric (abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_-)
+// # VARIABLE is a string that specifies the variable being read in DATA_FILE.  Must be alphanumeric (abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890_-)
 // VARIABLE: myVariable
 //
 // # BRICK_ORIGIN lets you specify a new coordinate system origin for # the mesh that will be created to suit your data.  It must contain three values.
@@ -76,7 +76,7 @@ class BOVCollection;
 //! Scientific notation is not supported for integer values like DATA_SIZE.
 //! Wild card characters are not currently supported in the DATA_FILE token.
 //! Each .bov file can only refer to a single data file.
-//! VARIABLE must be alphanumeric (abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_-)
+//! VARIABLE must be alphanumeric (abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890_-)
 //!
 //! \author Scott Pearse
 //! \date    May, 2021
--- a/lib/vdc/BOVCollection.cpp
+++ b/lib/vdc/BOVCollection.cpp
@@ -107,18 +107,18 @@ int BOVCollection::Initialize(const std::vector<std::string> &paths)
             if (_dataFormat == _defaultFormat) { return _missingValueError(FORMAT_TOKEN); }
             if (_gridSize == _defaultGridSize) { return _missingValueError(GRID_SIZE_TOKEN); }
 
-            std::ifstream infile(_dataFile);
-            if (infile.bad()) {
-                SetErrMsg(("Failed to open BOV file " + _dataFile).c_str());
-                return -1;
-            }
-
             _populateDataFileMap();
         } else {
             SetErrMsg(("Failed to open BOV file " + paths[0]).c_str());
             return -1;
         }
         header.close();
+
+        // If _dataFile is not an absolute path, prepend with the BOV header's path
+        if (!Wasp::FileUtils::IsPathAbsolute(_dataFile)) {
+            auto paths = {_currentFilePath, _dataFile};
+            _dataFile = Wasp::FileUtils::JoinPaths(paths);
+        }
     }
 
     return 0;
@@ -248,30 +248,9 @@ int BOVCollection::_validateParsedValues()
         _byteOffsetAssigned = true;
     }
 
-    // If _dataFile is not an absolute path, prepend with the BOV header's path
-    if (!Wasp::FileUtils::IsPathAbsolute(_dataFile)) { 
-        auto paths = {_currentFilePath, _dataFile};
-        _dataFile = Wasp::FileUtils::JoinPaths(paths); 
-    }
-
     // Validate whether we can open the data file
     FILE *fp = fopen(_dataFile.c_str(), "rb");
     if (fp == nullptr) return _invalidFileError();
-
-    // Validate the data file's size
-    int    formatSize = _sizeOfFormat(_dataFormat);
-    size_t count = _gridSize[0] * _gridSize[1] * _gridSize[2];
-    auto   readBuffer = std::unique_ptr<unsigned char[]>(new unsigned char[count * formatSize]);
-    size_t numElements = fread(readBuffer.get(), formatSize, count, fp);
-    if (numElements != count) {
-        fclose(fp);
-        return _invalidFileSizeError(numElements);
-    }
-    if (ferror(fp)) {
-        fclose(fp);
-        return _invalidFileError();
-    }
-
     fclose(fp);
 
     return 0;
@@ -368,8 +347,12 @@ template<> int BOVCollection::_findToken<DC::XType>(const std::string &token, st
 
     if (line.length() == 0) return (int)parseCodes::NOT_FOUND;
 
-    while (line[line.length() - 1] == ' ')    // If last char is a space, pop it
-        line.pop_back();
+    while (line.length() > 0) {
+        if (line[line.length() - 1] == ' ')    // If last char is a space, pop it
+            line.pop_back();
+        else
+            break;
+    }
 
     size_t pos = line.find(token);
     if (pos != std::string::npos) {    // We found the token
@@ -403,8 +386,12 @@ template<typename T> int BOVCollection::_findToken(const std::string &token, std
 
     if (line.length() == 0) return (int)parseCodes::NOT_FOUND;
 
-    while (line[line.length() - 1] == ' ')    // If last char is a space, pop it
-        line.pop_back();
+    while (line.length() > 0) {
+        if (line[line.length() - 1] == ' ')    // If last char is a space, pop it
+            line.pop_back();
+        else
+            break;
+    }
 
     size_t pos = line.find(token);
     if (pos != std::string::npos) {    // We found the token
@@ -440,8 +427,12 @@ template<typename T> int BOVCollection::_findToken(const std::string &token, std
 
     if (line.length() == 0) return (int)parseCodes::NOT_FOUND;
 
-    while (line[line.length() - 1] == ' ')    // If last char is a space, pop it
-        line.pop_back();
+    while (line.length() > 0) {
+        if (line[line.length() - 1] == ' ')    // If last char is a space, pop it
+            line.pop_back();
+        else
+            break;
+    }
 
     size_t pos = line.find(token);
     if (pos != std::string::npos) {    // We found the token
--- a/lib/vdc/BOVCollection.cpp
+++ b/lib/vdc/BOVCollection.cpp
@@ -6,6 +6,7 @@
 #include <iostream>
 #include <fstream>
 #include <type_traits>
+#include <sys/stat.h>
 #include "vapor/VAssert.h"
 #include "vapor/utils.h"
 #include <stdio.h>
@@ -62,17 +63,6 @@ const std::string BOVCollection::_intFormatString = "INT";
 const std::string BOVCollection::_floatFormatString = "FLOAT";
 const std::string BOVCollection::_doubleFormatString = "DOUBLE";
 
-namespace {
-    void SplitFilename (const string& str)
-    {
-      size_t found;
-      cout << "Splitting: " << str << endl;
-      found=str.find_last_of("/\\");
-      cout << " folder: " << str.substr(0,found) << endl;
-      cout << " file: " << str.substr(found+1) << endl;
-    }
-}
-
 BOVCollection::BOVCollection()
 : _time(_defaultTime), _dataFile(_defaultFile), _dataFormat(_defaultFormat), _variable(_defaultVar), _dataEndian(_defaultEndian), _centering(_defaultCentering), _byteOffset(_defaultByteOffset),
   _divideBrick(_defaultDivBrick), _dataComponents(_defaultComponents), _tmpDataFormat(_defaultFormat), _tmpDataEndian(_defaultEndian), _tmpByteOffset(_defaultByteOffset), _gridSizeAssigned(false),
@@ -99,9 +89,14 @@ int BOVCollection::Initialize(const std::vector<std::string> &paths)
     for (int i = 0; i < paths.size(); i++) {
         _dataFile = _defaultFile;
 
+        // Save the path to the BOV header so we can add it
+        // to data files given with a relative path
         _currentFilePath = paths[i];
         size_t found = _currentFilePath.find_last_of("/\\");
         _currentFilePath = _currentFilePath.substr(0,found);
+
+        _timeAssigned = false;
+        _variableAssigned = false;
 
         header.open(paths[i]);
         if (header.is_open()) {
@@ -160,15 +155,21 @@ int BOVCollection::_parseHeader(std::ifstream &header)
         rc = _findToken(TIME_TOKEN, line, time);
         if (rc == (int)parseCodes::ERROR)
             return _failureToReadError(TIME_TOKEN);
-        else if (rc == (int)parseCodes::FOUND)
+        else if (rc == (int)parseCodes::FOUND) {
+            if (_timeAssigned == true) return _multipleTimestepError();
             _time = time;
+            _timeAssigned = true;
+        }
 
         std::string variable;
         rc = _findToken(VARIABLE_TOKEN, line, variable);
         if (rc == (int)parseCodes::ERROR)
             return _invalidValueError(VARIABLE_TOKEN);
-        else if (rc == (int)parseCodes::FOUND)
+        else if (rc == (int)parseCodes::FOUND) {
+            if (_variableAssigned == true) return _multipleVariablesError();
             _variable = variable;
+            _variableAssigned = true;
+        }
 
         rc = _findToken(GRID_SIZE_TOKEN, line, _tmpGridSize);
         if (rc == (int)parseCodes::ERROR) return _failureToReadError(GRID_SIZE_TOKEN);
@@ -202,6 +203,9 @@ int BOVCollection::_parseHeader(std::ifstream &header)
 
 int BOVCollection::_validateParsedValues()
 {
+    // Validate that the data file exists.
+    // If not given an absolute path, construct one
+    // from the header file's containing directory.
     char actualPath[PATH_MAX+1];
     char* success = realpath(_dataFile.c_str(), actualPath);
     if (success == nullptr)
@@ -267,6 +271,19 @@ int BOVCollection::_validateParsedValues()
         _byteOffset = _tmpByteOffset;
         _byteOffsetAssigned = true;
     }
+
+    // Validate file size
+    size_t fileSize;
+    size_t dataSize = _gridSize[0]*_gridSize[1]*_gridSize[2]*_sizeOfFormat(_dataFormat);
+    struct stat stat_buf;
+    int rc = stat(_dataFile.c_str(), &stat_buf);
+    if (rc == 0)
+        fileSize = stat_buf.st_size;
+    else
+        return _cannotStatFileError();
+    if (dataSize != fileSize)
+        return _invalidFileSizeError(dataSize, fileSize);
+
     return 0;
 }
 
@@ -278,6 +295,17 @@ void BOVCollection::_populateDataFileMap()
     std::sort(_times.begin(), _times.end());
 
     _dataFileMap[_variable][_time] = _dataFile;
+}
+
+int BOVCollection::_cannotStatFileError() const {
+    SetErrMsg(("Unable to get file status for " + _dataFile).c_str());
+    return -1;
+}
+
+int BOVCollection::_invalidFileSizeError(size_t dataSize, size_t fileSize) const {
+    SetErrMsg(("Data file " + _dataFile + " of size " + to_string(fileSize) + " bytes does not match the size of the the data "
+        "specified in BOV header (" + to_string(dataSize) + " bytes).").c_str());
+    return -1;
 }
 
 int BOVCollection::_invalidFileError(const std::string &token, const std::string &file) const
--- a/lib/vdc/BOVCollection.cpp
+++ b/lib/vdc/BOVCollection.cpp
@@ -96,7 +96,7 @@ int BOVCollection::Initialize(const std::vector<std::string> &paths)
         // to data files given with a relative path
         _currentFilePath = paths[i];
         size_t found = _currentFilePath.find_last_of("/\\");
-        _currentFilePath = _currentFilePath.substr(0,found);
+        _currentFilePath = _currentFilePath.substr(0, found);
 
         header.open(paths[i]);
         if (header.is_open()) {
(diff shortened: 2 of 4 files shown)


lib/vdc/BOVCollection.cpp
```diff
@@ -181,6 +197,20 @@ int BOVCollection::_parseHeader(std::ifstream &header)
...
 {
+    // Validate that the data file exists.
+    // If not given an absolute path, construct one
+    // from the header file's containing directory.
+    char  actualPath[PATH_MAX + 1];
+    char *success = realpath(_dataFile.c_str(), actualPath);
+    if (success == nullptr) {
+        // At this point we can't find the absolute path, so it might be a relative path.
+        // Try prepending the directory of the .bov file to the data file
+        _dataFile = _currentFilePath + "//" + _dataFile;
```
clyne, line 208: @sgpearse, we have a collection of filesystem path manipulation routines in FileUtils.{h,cpp} that provide OS agnostic support for path concatenation, etc. We should use these - and possibly extend as necessary - rather than reinventing the wheel here (and above).

That said, c++17 provides (finally!!) STL utilities for path manipulation. See [here](https://en.cppreference.com/w/cpp/filesystem/path). 

@shaomeng and I had a discussion about moving to C++17 a few weeks ago. Using C++17 features may not be something we want to look at right now. On the other hand, doing the experiment to see if std::filesystem is supported on all of our platforms should not be too hard. At the very least it would be worth looking at what is in std::filesystem so you know what is coming.

cc @StasJ

lib/vdc/BOVCollection.cpp
```diff
@@ -232,6 +263,18 @@ int BOVCollection::_validateParsedValues()
         _byteOffset = _tmpByteOffset;
         _byteOffsetAssigned = true;
     }
+
+    // Validate file size
+    size_t      fileSize;
```
clyne, line 268: This validation seems overly restrictive: it requires the file to be the exact same size as the volume. This would, for example, prevent someone from storing multiple volumes in the same file and using the 'offset' to access the different volumes (variables).

Secondly, why is this check even necessary at all If there is a short file that should be detected when the file is read. Catching this error here seems superfluous. Am I missing something?

lib/vdc/BOVCollection.cpp
```diff
@@ -245,38 +288,67 @@ void BOVCollection::_populateDataFileMap()
...
 }
 
-int BOVCollection::_missingValueError(std::string token) const
+int BOVCollection::_cannotStatFileError() const
+{
+    SetErrMsg(("Unable to get file status for " + _dataFile).c_str());
+    return -1;
+}
+
+int BOVCollection::_invalidFileSizeError(size_t dataSize, size_t fileSize) const
```
clyne, line 297: Again, this seems like a lot of additional code to simply report a file read error. E.g.

```
if ((rc = fread()) < 0) {
    SetErrMsg("Error readfile file %s : %M", path.c_str());
    return(-1);
}
```

lib/vdc/BOVCollection.cpp
sgpearse: Ok, I'll look into it and report back.  Hopefully we can bump to c++17.

lib/vdc/BOVCollection.cpp
sgpearse: @shaomeng and I were discussing this in the #2791 thread.  I think this is a question for him.

lib/vdc/BOVCollection.cpp
sgpearse: I see what you mean.

Most (not all) of my error messages here are called through functions.  I think this makes [the validation code](https://github.com/NCAR/VAPOR/blob/a4f4924ad15112a439c27429b1911faa96b1f904/lib/vdc/BOVCollection.cpp#L198) easier to read, even though there are technically more total lines of code in the file.

Currently, the validation for each token is ~8 lines of code and some tests have more than one error condition.  I thought it would be more readable to call something like _inconsitentValueError(), rather than SetErrMsg(message) and then return -1 for [validating the data format across BOV files](https://github.com/NCAR/VAPOR/blob/a4f4924ad15112a439c27429b1911faa96b1f904/lib/vdc/BOVCollection.cpp#L228).  I think it's just slightly more readable this way.

If this doesn't seem worth it, let me know and I can revert the error reporting functions.

lib/vdc/BOVCollection.cpp
shaomeng: I believe errors of the following nature should be caught: when the code wants to read in `N` bytes and there are `M` bytes left in the file where `M < N`.

How to catch these errors: `std::fread()` returns how many values it successfully reads in. Compare the return value with the expected number of values to read.

lib/vdc/BOVCollection.cpp
shaomeng: Just a note to @clyne : `fread()` returns a value of type `size_t` so `if ((rc = fread()) < 0) `  will give you unexpected answers. (Use a static analyzer to catch this type of errors.)

lib/vdc/BOVCollection.cpp
shaomeng: I'm already using c++17 in my compression code and there are some very nice things about it.

clyne: That's encouraging. Hopefully they work on all of our supported platforms. We shouldn't make the jump until we know this for sure. To be clear, I'm not asking @sgpearse to use C++17 if the timing isn't right. But at very least we should using are existing platform-portable path handing tools, with the long term goal of being able to replace these with something supported by the STL.

lib/vdc/BOVCollection.cpp
clyne: I think @shaomeng and i are  both saying that these errors can be caught by the return value from fread, and that it is not necessary to stat the file before reading it.

lib/vdc/BOVCollection.cpp
clyne: My bad. Correct. The return value should be compared with the total number of bytes requested and if they disagree there was an error or a short file.

lib/vdc/BOVCollection.cpp
StasJ: I would be concerned about Windows support with C++17 features

lib/vdc/BOVCollection.cpp
shaomeng: Well, be careful again: compare to the number of objects to read, ***not*** the number of bytes. I.e. if you want to read 10 floats, compare with 10, instead of 40.

lib/vdc/BOVCollection.cpp
```diff
@@ -473,18 +549,17 @@ template<class T> int BOVCollection::ReadRegion(std::string varname, size_t ts,
         return -1;
     }
 
-    size_t numValues = _gridSize[0] * _gridSize[1] * _gridSize[2];
-
     int  n = 1;
     bool systemLittleEndian = *(char *)&n == 1 ? true : false;
-    bool dataLittleEndian = _dataEndian == "LITTLE" ? true : false;
+    bool dataLittleEndian = _dataEndian == _littleEndianString ? true : false;
     bool needSwap = systemLittleEndian != dataLittleEndian ? true : false;
```
shaomeng, line 555: Why don't you write it in a simpler form (and more intuitive):
```c++
bool dataLittleEndian = _dataEndian == _littleEndianString;
bool needSwap = systemLittleEndian != dataLittleEndian;
```

lib/vdc/BOVCollection.cpp
```diff
@@ -181,6 +197,20 @@ int BOVCollection::_parseHeader(std::ifstream &header)
...
+    // Validate that the data file exists.
+    // If not given an absolute path, construct one
+    // from the header file's containing directory.
+    char  actualPath[PATH_MAX + 1];
+    char *success = realpath(_dataFile.c_str(), actualPath);
+    if (success == nullptr) {
+        // At this point we can't find the absolute path, so it might be a relative path.
+        // Try prepending the directory of the .bov file to the data file
+        _dataFile = _currentFilePath + "//" + _dataFile;
+        success = realpath(_dataFile.c_str(), actualPath);
```
shaomeng, line 209: Did you remember to free the buffer returned by `realpath()`?

shaomeng (changes requested):

lib/vdc/BOVCollection.cpp
clyne: Indeed. It's the number of elements not bytes

lib/vdc/BOVCollection.cpp
```diff
@@ -307,12 +379,14 @@ template<> int BOVCollection::_findToken<DC::XType>(const std::string &token, st
     for (size_t i = 0; i < line.length(); i++) {
         if (line[i] == '#') {
             line.erase(line.begin() + i, line.end());
-            if (line[line.length() - 1] == ' ')    // If last char is a space, pop it
-                line.pop_back();
             break;
         }
     }
 
+    while (line[line.length() - 1] == ' ')    // If last char is a space, pop it
```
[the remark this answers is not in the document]
sgpearse: [The getLine() function](https://github.com/NCAR/VAPOR/blob/9058cf276651c60c65580c083b907042c934ce97/lib/vdc/BOVCollection.cpp#L143) does not pass empty lines to the parser function.

lib/vdc/BOVCollection.cpp
```diff
@@ -381,12 +456,13 @@ template<typename T> int BOVCollection::_findToken(const std::string &token, std
...
```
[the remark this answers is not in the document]
sgpearse: [The getLine() function](https://github.com/NCAR/VAPOR/blob/9058cf276651c60c65580c083b907042c934ce97/lib/vdc/BOVCollection.cpp#L143) does not pass empty lines to the parser function.

lib/vdc/BOVCollection.cpp
```diff
@@ -132,8 +147,9 @@ int BOVCollection::_parseHeader(std::ifstream &header)
         rc = _findToken(DATA_FILE_TOKEN, line, dataFile);
         if (rc == (int)parseCodes::ERROR)
             return _failureToReadError(DATA_FILE_TOKEN);
-        else if (rc == (int)parseCodes::FOUND)
+        else if (rc == (int)parseCodes::FOUND) {
             _dataFile = dataFile;
+        }
```
shaomeng, line 152: In the case when you do find a token from the current line, do you expect to find another token from the same line? If not, would it make sense to skip the rest of the token detection code in the current iteration?

shaomeng (changes requested):

lib/vdc/BOVCollection.cpp
```diff
@@ -218,6 +248,7 @@ int BOVCollection::_validateParsedValues()
     }
 
     // Validate endian type
+    if (_tmpDataEndian != _bigEndianString && _tmpDataEndian != _littleEndianString) return _invalidEndianError(ENDIAN_TOKEN);
```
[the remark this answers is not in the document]
sgpearse: My opinion is to keep it, since we support endianness with our raw2vdc converter.  If we don't want to support this, we should deprecate it in the BOV reader and in raw2vdc.

lib/vdc/BOVCollection.cpp
sgpearse: The return value not a buffer - the buffer (actualPath) is a local array, so it doesn't need to be freed.

https://stackoverflow.com/questions/1563168/example-of-realpath-function-in-c

lib/vdc/BOVCollection.cpp
```diff
@@ -181,6 +197,20 @@ int BOVCollection::_parseHeader(std::ifstream &header)
 
 int BOVCollection::_validateParsedValues()
 {
+    // Validate that the data file exists.
+    // If not given an absolute path, construct one
+    // from the header file's containing directory.
+    char  actualPath[PATH_MAX + 1];
+    char *success = realpath(_dataFile.c_str(), actualPath);
```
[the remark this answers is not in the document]
sgpearse: The return value not a buffer - the buffer (actualPath) is a local array, so it doesn't need to be freed.

https://stackoverflow.com/questions/1563168/example-of-realpath-function-in-c

lib/vdc/BOVCollection.cpp
shaomeng: That's right! No need to be freed!

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
shaomeng: That's right! No need to be freed!

lib/vdc/BOVCollection.cpp
```diff
@@ -181,6 +197,20 @@ int BOVCollection::_parseHeader(std::ifstream &header)
 
 int BOVCollection::_validateParsedValues()
 {
+    // Validate that the data file exists.
+    // If not given an absolute path, construct one
+    // from the header file's containing directory.
+    char  actualPath[PATH_MAX + 1];
```
shaomeng, line 203: Is the function `realpath()` available on Windows?

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
shaomeng: but there's no way to test if the code is correct or not. Actually, it has already caused a few bugs even before I could test if the correctness of big/small conversion code. So to reduce further bugs, deleting it is the best way.

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
StasJ: The action immediately before this deletes parts of the line passed into the parser function.

lib/vdc/BOVCollection.cpp
sgpearse: Nope.  Switched to use Wasp::FileUtils on current branch.

lib/vdc/BOVCollection.cpp
sgpearse: Nope, fixed.

lib/vdc/BOVCollection.cpp
sgpearse: Endian stuff has been removed.

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
sgpearse: Endian stuff has been removed.

lib/vdc/BOVCollection.cpp
```diff
@@ -6,12 +6,17 @@
...
+#include "vapor/utils.h"
 #include <stdio.h>
 
-#ifdef _WINDOWS
+#ifdef WIN32
+    #include <Windows.h>
     #define _USE_MATH_DEFINES
     #pragma warning(disable : 4251 4100)
+#else
+    #include <limits.h>
```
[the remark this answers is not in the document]
sgpearse: Fixed.

lib/vdc/BOVCollection.cpp
```diff
@@ -6,12 +6,17 @@
 #include <iostream>
 #include <fstream>
 #include <type_traits>
+#include <sys/stat.h>
```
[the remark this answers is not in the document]
sgpearse: Removed.

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
sgpearse: Fixed.

lib/vdc/BOVCollection.cpp
```diff
@@ -339,12 +413,13 @@ template<typename T> int BOVCollection::_findToken(const std::string &token, std
     for (size_t i = 0; i < line.length(); i++) {
         if (line[i] == '#') {
             line.erase(line.begin() + i, line.end());
-            if (line[line.length() - 1] == ' ')    // If last char is a space, pop it
-                line.pop_back();
             break;
         }
     }
 
+    while (line[line.length() - 1] == ' ')    // If last char is a space, pop it
```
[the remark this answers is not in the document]
sgpearse: Fixed.

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
sgpearse: Fixed.

lib/vdc/BOVCollection.cpp
sgpearse: Fixed at the end of the _validateParsedValues function.

lib/vdc/BOVCollection.cpp
sgpearse: Fixed at the end of the _validateParsedValues function.

lib/vdc/BOVCollection.cpp
sgpearse: Currently using the WASP library's FileUtils for this.

StasJ (approved):

lib/vdc/BOVCollection.cpp
```diff
@@ -7,11 +8,16 @@
...
 #include <stdio.h>
 
-#ifdef _WINDOWS
+#ifdef WIN32
+    #include <Windows.h>
     #define _USE_MATH_DEFINES
     #pragma warning(disable : 4251 4100)
+#else
+    #include <climits>
 #endif
```
shaomeng, line 21: Header `climits` seems to be a C++ standard one. Is it not available on Windows so you have to do this complicated trick?

lib/vdc/BOVCollection.cpp
```diff
@@ -132,45 +145,64 @@ int BOVCollection::_parseHeader(std::ifstream &header)
...
         rc = _findToken(OFFSET_TOKEN, line, _tmpByteOffset);
-        if (rc == (int)parseCodes::PARSE_ERROR) return _invalidValueError(OFFSET_TOKEN);
+        if (rc == (int)parseCodes::PARSE_ERROR)
+            return _invalidValueError(OFFSET_TOKEN);
+        else if (rc == (int)parseCodes::FOUND)
+            continue;
 
         // All other variables are currently unused.
         //
+        _findToken(ENDIAN_TOKEN, line, _dataEndian);
```
shaomeng, line 205: I thought endian stuff is removed?

lib/vdc/BOVCollection.cpp
```diff
@@ -60,8 +66,8 @@ const std::string BOVCollection::_doubleFormatString = "DOUBLE";
 
 BOVCollection::BOVCollection()
 : _time(_defaultTime), _dataFile(_defaultFile), _dataFormat(_defaultFormat), _variable(_defaultVar), _dataEndian(_defaultEndian), _centering(_defaultCentering), _byteOffset(_defaultByteOffset),
```
shaomeng, line 68: I thought endian stuff is removed?

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
shaomeng: Seems not totally removed?

shaomeng:
@sgpearse since there's still not a VAPOR BOV documentation available, can you give me a list of keywords from the VisIt documentation that is supported, and any limitation that you imposed on top of VisIt specifications, so I can run some tests?

lib/vdc/BOVCollection.cpp
sgpearse: I'm still finding unused tokens, in case we want to implement them later.  Check out the lines below for other examples.

lib/vdc/BOVCollection.cpp
sgpearse: Let me check.

lib/vdc/BOVCollection.cpp
sgpearse: I'm still finding unused tokens, in case we want to implement them later.

lib/vdc/BOVCollection.cpp
```diff
@@ -217,21 +249,37 @@ int BOVCollection::_validateParsedValues()
...
+
+    // Validate the data file's size
+    int    formatSize = _sizeOfFormat(_dataFormat);
+    size_t count = _gridSize[0] * _gridSize[1] * _gridSize[2];
+    auto   readBuffer = std::unique_ptr<unsigned char[]>(new unsigned char[count * formatSize]);
+    size_t numElements = fread(readBuffer.get(), formatSize, count, fp);
+    if (numElements != count) {
+        fclose(fp);
+        return _invalidFileSizeError(numElements);
+    }
```
shaomeng, line 275: Scott, 2 comments:
1) when you validate the file size, remember to take into consideration the offset bytes.
2) your current implementation reads the entire file just to count its size. This is very cumbersome. You should use `std::fseek()` to try to access the desired location and then use the return value to determine if good or bad. 
https://en.cppreference.com/w/cpp/io/c/fseek

shaomeng: should check return value of `fseek()` in other places too.

sgpearse:
@shaomeng Of course.  I know the missing documentation makes this hard.  DCBOV.h currently lists supported tokens.  Thanks for reminding me to edit the header for DATA_ENDIAN.

Supported tokens are as follows:

```
//! The following BOV tags are mandatory for Vapor to ingest data:
//! - DATA_FILE
//! - DATA_SIZE
//! - DATA_FORMAT
//!
//! The following BOV tags are optional:
//! - BRICK_ORIGIN
//! - BRICK_SIZE
//! - TIME
//! - VARIABLE
```

lib/vdc/BOVCollection.cpp
shaomeng: Isn't it additional 1) bug potentials and 2) complexity (imagine someone else debugging) without any benefit?

sgpearse: Regarding 1)
No, because the parsing code is the same for all tokens.  When I started this, I wrote the code to parse all BOV tokens, and reduced the tokens that we supported as time allowed.  If we want to extend the tokens that are supported, we need to parse these anyways.

Regarding 2)
It's really just one line of code.  I don't think it adds complexity.  If a user adds a bad value for an unused token, yes this will throw an error.  But it's one line per token, and if we're following VisIt's example, I'd like to parse all of the tokens for completeness.  Even if they're not all used.

lib/vdc/BOVCollection.cpp
sgpearse: @shaomeng  It seems this is ok after building and testing on an installer.  File updated.

Note - the windows installer takes > 1 hour to build on cisl-vaporwin10 right now.  Changes like this are hard to test.  God help me if this introduces a regression.

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
sgpearse: I believe you're looking at an outdated file.

lib/vdc/BOVCollection.cpp
clyne: @sgpearse please use FileUtils::Dirname to get the directory path here:

https://github.com/NCAR/VAPOR/blob/527fa55a484b2af30478774dbbe6ecea8e07c00d/lib/vdc/BOVCollection.cpp#L97

and use FileUtils::JoinPaths to concatenate paths here:

https://github.com/NCAR/VAPOR/blob/527fa55a484b2af30478774dbbe6ecea8e07c00d/lib/vdc/BOVCollection.cpp#L261

cc @StasJ

lib/vdc/BOVCollection.cpp
clyne: Please see comments above regarding removing file size validation entirely from this function. If there is an error reading the file it will be caught when the user actually attempts to access the file. This is consistent with what we do elsewhere.

clyne (commented):
> @shaomeng Of course. I know the missing documentation makes this hard. DCBOV.h currently lists supported tokens. Thanks for reminding me to edit the header for DATA_ENDIAN.
> 
> Supported tokens are as follows:
> 
> ```
> //! The following BOV tags are mandatory for Vapor to ingest data:
> //! - DATA_FILE
> //! - DATA_SIZE
> //! - DATA_FORMAT
> //!
> //! The following BOV tags are optional:
> //! - BRICK_ORIGIN
> //! - BRICK_SIZE
> //! - TIME
> //! - VARIABLE
> ```

What about BYTE_OFFSET?

lib/vdc/BOVCollection.cpp
```diff
@@ -217,21 +242,37 @@ int BOVCollection::_validateParsedValues()
...
     // Validate byte offest
     if (_tmpByteOffset != _byteOffset && _byteOffsetAssigned == true)
         return _inconsistentValueError(OFFSET_TOKEN);
     else {
         _byteOffset = _tmpByteOffset;
         _byteOffsetAssigned = true;
     }
+
+    // If _dataFile is not an absolute path, prepend with the BOV header's path
+    if (!Wasp::FileUtils::IsPathAbsolute(_dataFile)) { _dataFile = _currentFilePath + "//" + _dataFile; }
```
clyne, line 254: Does it really make sense do generate the path to the data file here? The name of this function is 'validateParsedValues'. One would expect the function to simply validate the inputs. Instead, as a side effect this function is modifying _dataFile. Why not simply handle the pathing issue in the Initialize loop where the rest of the file  path handling code resides?

lib/vdc/BOVCollection.cpp
```diff
@@ -298,21 +354,22 @@ std::array<double, 3> BOVCollection::GetBrickOrigin() const { return _brickOrigi
...
             line.erase(line.begin() + i, line.end());
-            if (line[line.length() - 1] == ' ')    // If last char is a space, pop it
-                line.pop_back();
             break;
         }
     }
 
+    if (line.length() == 0) return (int)parseCodes::NOT_FOUND;
+
+    while (line[line.length() - 1] == ' ')    // If last char is a space, pop it
```
clyne, line 370: Won't this generate an invalid memory access if line contains only a single space?

clyne (commented):
Still a few more things that need to be addressed.

lib/vdc/BOVCollection.cpp
shaomeng: Ok, how about this: 
1) document in the code in the clearest possible language that these are dummy values being passed around without being used by anybody.
2) document in the code in the clearest possible language that what will happen when a good value is passed in, when a bad value is passed in, when a duplicate value is passed in, etc.

The goal is, for example, 6 months later Sam is tasked to complete the current BOV reader to support all tokens, Sam can know the nuances that you described in the past month in multiple PR threads. (Possibly: rename all the unused tokens/handles etc. with a prefix/postfix `dummy`.)

shaomeng:
> //! The following BOV tags are mandatory for Vapor to ingest data:
> //! - DATA_FILE
> //! - DATA_SIZE
> //! - DATA_FORMAT
> //!
> //! The following BOV tags are optional:
> //! - BRICK_ORIGIN
> //! - BRICK_SIZE
> //! - TIME
> //! - VARIABLE
> ```
I suppose when the optional tags ain't provided, a default value will be assigned? If so, can you document what the default value is for each optional tag?

sgpearse:
> I suppose when the optional tags ain't provided, a default value will be assigned? If so, can you document what the default value is for each optional tag?

Sure, I've put it in the DCBOV's class description in the header file.

lib/vdc/BOVCollection.cpp
sgpearse: Fixed.

lib/vdc/BOVCollection.cpp
sgpearse: I can do that.  

I was thinking that validating something can mean to check if its correct, or to make something valid.  But since this is the only variable I'm correcting, I'll put it in the Initialize() function instead.

sgpearse:
> What about BYTE_OFFSET?

Added to the optional tag list.

lib/vdc/BOVCollection.cpp
```diff
@@ -217,21 +249,37 @@ int BOVCollection::_validateParsedValues()
...
+
+    // If _dataFile is not an absolute path, prepend with the BOV header's path
+    if (!Wasp::FileUtils::IsPathAbsolute(_dataFile)) { _dataFile = _currentFilePath + "//" + _dataFile; }
+
+    // Validate whether we can open the data file
+    FILE *fp = fopen(_dataFile.c_str(), "rb");
+    if (fp == nullptr) return _invalidFileError();
+
+    // Validate the data file's size
+    int    formatSize = _sizeOfFormat(_dataFormat);
```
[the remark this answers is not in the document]
sgpearse: This seems like it was a miscommunication.  Checking for short reads has always [been done](https://github.com/NCAR/VAPOR/blob/f70d912c40f64e628639367ec83ec34dc16504ed/lib/vdc/BOVCollection.cpp#L520) by checking the return code of fread.  Extra validation code was just added at Sam's request in [this bug report](https://github.com/NCAR/VAPOR/issues/2792).

The recently added validation code has been removed.

lib/vdc/BOVCollection.cpp
sgpearse: Fixed.

lib/vdc/BOVCollection.cpp
```diff
@@ -7,12 +8,10 @@
 #include <fstream>
 #include <type_traits>
 #include "vapor/VAssert.h"
+#include "vapor/utils.h"
+#include "vapor/FileUtils.h"
 #include <stdio.h>
```
shaomeng, line 13: Why don't you also switch this header to its C++ counterpart: cstdio?

lib/vdc/BOVCollection.cpp
```diff
@@ -473,28 +514,26 @@ template<class T> int BOVCollection::ReadRegion(std::string varname, size_t ts,
...
     for (size_t k = min[2]; k <= max[2]; k++) {
         size_t zOffset = _gridSize[0] * _gridSize[1] * k;
         for (size_t j = min[1]; j <= max[1]; j++) {
             size_t xOffset = min[0];
             size_t yOffset = _gridSize[0] * j;
             size_t offset = formatSize * (xOffset + yOffset + zOffset) + _byteOffset;
 
-            fseek(fp, offset, SEEK_SET);
-            size_t rc = fread(readBuffer, formatSize, count, fp);
+            size_t rc = fseek(fp, offset, SEEK_SET);
```
shaomeng, line 542: Scott, the function fseek returns a type of signed int. Please, don't do a signed -> unsigned conversion, please.

shaomeng: after you fix this, please check other function return types, and make sure you don't make unintended conversions between signed and unsigned integers.

shaomeng (changes requested):
A few more Q in the file DCBOV.cpp:
1) shouldn't the macro be WIN32 or _WIN32?
https://github.com/NCAR/VAPOR/blob/e599b72e8a4617585cef2add29b4d1967057c623/lib/vdc/DCBOV.cpp#L9

2) When the initialize function called twice, will the previous BOVCollection object get lost? 
https://github.com/NCAR/VAPOR/blob/e599b72e8a4617585cef2add29b4d1967057c623/lib/vdc/DCBOV.cpp#L37-L40

lib/vdc/BOVCollection.cpp
sgpearse: > 1. document in the code in the clearest possible language that these are dummy values being passed around without being used by anybody.
> 2. document in the code in the clearest possible language that what will happen when a good value is passed in, when a bad value is passed in, when a duplicate value is passed in, etc.

I've added additional comments in the BOVCollection.h/cpp files wherever these variables are defined or used.

I've also added more information to DCBOV.h, where you can find the rules for things like what happens when a bad value is passed in for different types of tokens, and an example BOV header file.

> How big the effort is to add them back in, and how big is the chance that they will be added back in?

I think that's hard to predict.  John has talked about how wacky these binary data files can get, and I think that VisIt wouldn't have these tokens without good reason.

The unused values haven't resulted in any bugs, and they comply with the specification I was given.  Removing them, and then re-adding them later would be more effort than letting things be.

include/vapor/DCBOV.h
```diff
@@ -16,6 +16,33 @@ namespace VAPoR {
...
+// # DATA_FILE points to a binary data file.  It can be a full file path, or a path relative to the BOV header.
+// DATA_FILE: bovA1.bin
+//
+// # The data size corresponds to NX,NY,NZ in the above example code.  It must contain three values
+// DATA_SIZE: 10 10 10
+//
+// # Allowable values for DATA_FORMAT are: INT,FLOAT,DOUBLE
+// DATA_FORMAT: FLOAT
+//
+// # VARIABLE is a string that specifies the variable being read in DATA_FILE.  Must be alphanumeric (abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890)
```
shaomeng, line 33: I would suggest adding a hyphen and an underscore into the allowed list, since those 2 are pretty common and useful in variable names.

shaomeng (changes requested):

include/vapor/DCBOV.h
sgpearse: No problem.  I'll push soon.

lib/vdc/BOVCollection.cpp
sgpearse: Thanks, fixed.  Pushing soon.

lib/vdc/BOVCollection.cpp
sgpearse: You're right.  Fixed.  Pushing soon.

sgpearse:
@shaomeng regarding your two questions in DCBOV:

> shouldn't the macro be WIN32 or _WIN32?
I'm not sure, but this is how the other DC classes are configuring themselves for windows.

> When the initialize function called twice, will the previous BOVCollection object get lost?

@clyne may have some information on this, but my understanding is that each DC class (DCBOV, DCCF, DCWRF, etc) has a single "Collection" object that only gets initialized once.

sgpearse:
> I suppose when the optional tags ain't provided, a default value will be assigned? If so, can you document what the default value is for each optional tag?

I've added the defaults for the optional tokens in the DCBOV.h documentation.

lib/vdc/BOVCollection.cpp
StasJ: `/` is not a valid file separator on Windows. `FileUtils::JoinPaths` will do it correctly.

sgpearse: @stasj I'm 90% sure `//` is, but this is moot thanks to the FileUtils function.

clyne:
> @shaomeng regarding your two questions in DCBOV:
> 
> > shouldn't the macro be WIN32 or _WIN32?
> 
> I'm not sure, but this is how the other DC classes are configuring themselves for windows.
> 
> > When the initialize function called twice, will the previous BOVCollection object get lost?
> 
> @clyne may have some information on this, but my understanding is that each DC class (DCBOV, DCCF, DCWRF, etc) has a single "Collection" object that only gets initialized once.

Initialize() will reset the state each time it is called.  That is the intended behavior.

clyne (approved):

clyne:
@shaomeng have all of your concerns been addressed here so we can accept this?

lib/vdc/BOVCollection.cpp
shaomeng: Scott, 5 lines down, fread() will return a size_t type. Can you declare a separate variable with the proper type instead of re-using the same variable?

include/vapor/DCBOV.h
shaomeng: could you also remove the duplicate `0` (zero) in this allowed list? It helps the code to look better.

shaomeng:
> > When the initialize function called twice, will the previous BOVCollection object get lost?
> 
> @clyne may have some information on this, but my understanding is that each DC class (DCBOV, DCCF, DCWRF, etc) has a single "Collection" object that only gets initialized once.

The answer is yes. When initialize is called the 2nd time, line 39 will create a new object and the old object is lost, resulting in a memory leak. Please do one of the following: 
1) properly delete the old object, 
2) simply wrap the pointer in a unique_ptr so you don't need to worry about it anymore.

shaomeng:
> @shaomeng have all of your concerns been addressed here so we can accept this?

No, not addressed yet.

shaomeng (changes requested):
Q to Scott: I cannot understand the logic of using two variables to store the same token, and what problems that the logic tries to catch. Can you maybe use `_tmpDataFormat` and `_dataFormat` as an example to illustrate a problem that they will catch?

lib/vdc/BOVCollection.cpp
```diff
@@ -83,6 +93,11 @@ int BOVCollection::Initialize(const std::vector<std::string> &paths)
     std::ifstream header;
     for (int i = 0; i < paths.size(); i++) {
         _dataFile = _defaultFile;
+
+        // Save the path to the BOV header so we can add it
+        // to data files given with a relative path
+        _currentFilePath = Wasp::FileUtils::Dirname(paths[i]);
+
         header.open(paths[i]);
```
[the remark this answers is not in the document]
sgpearse: Yes, ifstream closes the file upon its destructor.

lib/vdc/BOVCollection.cpp
```diff
@@ -91,9 +106,15 @@ int BOVCollection::Initialize(const std::vector<std::string> &paths)
                 return -1;
             }
 
+            // If _dataFile is not an absolute path, prepend with the BOV header's path
+            if (!Wasp::FileUtils::IsPathAbsolute(_dataFile)) {
+                auto paths = {_currentFilePath, _dataFile};
```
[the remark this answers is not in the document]
sgpearse: Thanks, fixed.

lib/vdc/BOVCollection.cpp
```diff
@@ -132,45 +147,64 @@ int BOVCollection::_parseHeader(std::ifstream &header)
...
-        else if (rc == (int)parseCodes::FOUND)
+        else if (rc == (int)parseCodes::FOUND) {
             _dataFile = dataFile;
+            continue;
+        }
 
         double time;
         rc = _findToken(TIME_TOKEN, line, time);
         if (rc == (int)parseCodes::PARSE_ERROR)
             return _failureToReadError(TIME_TOKEN);
```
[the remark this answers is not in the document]
sgpearse: If the user wants to define an optional token, we need to validate it.  When the parser returns a PARSE_ERROR, something is wrong with the value, so we throw an error.

sgpearse:
> Q to Scott: I cannot understand the logic of using two variables to store the same token, and what problems that the logic tries to catch. Can you maybe use _tmpDataFormat and _dataFormat as an example to illustrate a problem that they will catch?

The _tmp* values need to be consistent across all BOV files.  IE - We can't have a single dataset that has DATA_FORMAT headers of both FLOAT and DOUBLE.  They need to be consistent among datasets.

When we read each BOV header, we assign a temporary variable (IE - _tmpDataFormat) to the current file being parsed.  After parsing, we do validation.  We check if this variable has already been assigned.  If not, we assign the _tmpDataFormat to _dataFormat.  If it _has been assigned_, and it disagrees with a previous file, we throw an error.

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
shaomeng: Got it, thanks!

shaomeng:
> > Q to Scott: I cannot understand the logic of using two variables to store the same token, and what problems that the logic tries to catch. Can you maybe use _tmpDataFormat and _dataFormat as an example to illustrate a problem that they will catch?
> 
> The _tmp* values need to be consistent across all BOV files. IE - We can't have a single dataset that has DATA_FORMAT headers of both FLOAT and DOUBLE. They need to be consistent among datasets.
> 
> When we read each BOV header, we assign a temporary variable (IE - _tmpDataFormat) to the current file being parsed. After parsing, we do validation. We check if this variable has already been assigned. If not, we assign the _tmpDataFormat to _dataFormat. If it _has been assigned_, and it disagrees with a previous file, we throw an error.

Got it, thanks!

lib/vdc/BOVCollection.cpp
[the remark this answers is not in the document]
shaomeng: It makes sense to validate it if it is provided. Just to make sure, if a BOV file has no time token at all, will a _failureToReadError be returned?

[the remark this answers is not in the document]
sgpearse: Nope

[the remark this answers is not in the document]
shaomeng: sounds good!

include/vapor/DCBOV.h
sgpearse: Sure, fixed.  Good eye.

shaomeng (approved):

lib/vdc/BOVCollection.cpp
sgpearse: Thanks, fixed.

PR #2811, opened 2021-07-19
Fix broken windows build

sgpearse:
Fixes our windows build!

### DIFF
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -154,7 +154,7 @@ using namespace VAPoR;
 
 const QEvent::Type MainForm::ParamsChangeEvent = (QEvent::Type)QEvent::registerEventType();
 const QEvent::Type MainForm::ParamsIntermediateChangeEvent = (QEvent::Type)QEvent::registerEventType();
-const std::string  MainForm::_documentationURL = "http://www.docs.vapor.ucar.edu";
+const std::string  MainForm::_documentationURL = "https://vapor.readthedocs.io/en/readthedocs/";
 
 namespace {
 
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -1171,25 +1171,25 @@ void MainForm::createBookmark()
     p->BeginGroup("Create Bookmark");
 
     vector<BookmarkParams> bookmarks;
-    for (auto* b : p->GetBookmarks()) bookmarks.push_back(BookmarkParams(*b));
+    for (auto *b : p->GetBookmarks()) bookmarks.push_back(BookmarkParams(*b));
     p->ClearBookmarks();
 
     Base16StringStream ss;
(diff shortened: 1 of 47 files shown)


apps/vaporgui/MainForm.cpp
```diff
@@ -1186,10 +1186,10 @@ void MainForm::createBookmark()
     int  customViewportWidth = vpp->GetValueLong(vpp->CustomFramebufferWidthTag, 0);
     int  customViewportHeight = vpp->GetValueLong(vpp->CustomFramebufferHeightTag, 0);
 
-    const int     iconSize = BookmarkParams::DefaultIconSize();
-    const int     iconDataSize = iconSize * iconSize * 3;
-    unsigned char iconData[iconDataSize];
-    char          iconDataString[64];
+    const int      iconSize = BookmarkParams::DefaultIconSize();
+    const int      iconDataSize = iconSize * iconSize * 3;
+    unsigned char *iconData = new unsigned char[iconDataSize];
```
shaomeng, line 1191: If this pointer is already pointing to a memory block, then this line will cause the old block to leak.

shaomeng (changes requested):

apps/vaporgui/MainForm.cpp
shaomeng: suggest avoid managing memory manually by using a unique ptr:
`auto iconData = std::make_unique<unsigned char[]>( iconDataSize );`

StasJ: How could this pointer already be pointing to a memory block?

apps/vaporgui/MainForm.cpp
```diff
@@ -1209,6 +1210,8 @@ void MainForm::createBookmark()
     Base16StringStream is;
     is.write((char *)iconData, iconDataSize);
 
+    if (iconData != nullptr) delete[] iconData;
```
StasJ, line 1213: You can't delete a unique ptr

StasJ (changes requested):
I know that I already made some of these fixes before so I wonder what happened to them... We've had windows building since the OSPRay PR and the Bookmarks PR and but that wouldn't have been possible without the changes here so they must've gotten rolled back somehow.

apps/vaporgui/MainForm.cpp
shaomeng: > How could this pointer already be pointing to a memory block?

That's right, it cannot. The original code won't produce a memory leak.

sgpearse:
@stasj, regarding your comments:

>  know that I already made some of these fixes before so I wonder what happened to them...

1) I recall fixing some of these problems too.  One of the lines breaking the windows build is in PGroup.cpp:6.  On main, this was last modified by StasJ in October 2020.  This change was before our 3.3 release , but now it's breaking the windows build.

2) Could you please fix the errors regarding your code, with whatever data structures suitable?  I need to focus on #2704, and this code is under your ownership.

shaomeng (approved):

StasJ (approved):

PR #2813, opened 2021-07-20
Fix #2800 ROMS DVR artifacts

### LINKED ISSUE #2800: DVR problem on ROMS data
The DVR is giving different results based on which side of the domain we are viewing from, in the ROMS/Kauffman dataset.

To reproduce:
1) On casper, open the session file /glade/p/cisl/vast/vapor/Bugs/2800.vs3
2) Note that the DVR is different when viewing through the YZ plane, versus the XY plane.

Screen Shot 2021-07-13 at 9 35 44 AM

Screen Shot 2021-07-13 at 9 36 41 AM

StasJ:
Fix #2800 and improve DVR support for non-hexahedral cells.

### DIFF
--- a/share/shaders/VolumeCellBase.frag
+++ b/share/shaders/VolumeCellBase.frag
@@ -664,7 +664,7 @@ void main(void)
             if (i++ > 8) {
                 break;
             }
-            if (t1-t0 <= EPSILON) {
+            if (abs(t1-t0) <= EPSILON) {
                 float unitDistanceScaled = unitDistance / length(dir * scales);
                 float step = unitDistanceScaled/7.f/samplingRateMultiplier * GetSamplingNoise();
                 t1 += step;
--- a/share/shaders/VolumeCellBase.frag
+++ b/share/shaders/VolumeCellBase.frag
@@ -664,10 +664,10 @@ void main(void)
             if (i++ > 8) {
                 break;
             }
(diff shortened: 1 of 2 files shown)


shaomeng (approved):
The rendering images look good to me!

share/shaders/VolumeCellBase.frag
```diff
@@ -551,34 +653,46 @@ void main(void)
...
                 vec4 color = Traverse(eye, dir, rayLightingNormal, tMin, tMax, t1, initialCell, entranceFace, t1);
                 BlendToBack(accum, color);
             }
             
             // Failsafe to prevent infinite recursion due to float precision error
-            if (i++ > 8 || t1-t0 <= EPSILON)
+            if (i++ > 8) {
                 break;
+            }
+            if (t1-t0 <= EPSILON) {
```
clyne, line 667: Is t1 guaranteed to be greater than t0 (i.e. no need to take absolute value)? Will the comparison with EPSILON work regardless of the user coordinates (it's not clear what coordinate system t0, t1 are in)

StasJ: `t` represents distance along the view ray as in `R=o+t*d`. This is supposed to test `t1 <= t0` rather than `t1 == t0`.

clyne (approved):

Issue #2820, opened 2021-07-22
Flow hangs with MPAS-O


StasJ:
1. Import MPAS-O
2. Create/enable a 2D pathline flow renderer with X-Field=snowFlux, Y-Field=rainFlux, Color=snowFlux
3. Play time steps to end of dataset
1. Change Y-field to xCell

Vapor will hang. I've left it running for 15 minutes with no change while the previous flow computation only takes about 10 seconds. Command-Q no longer works and vapor has to be force-quit externally.

OS: macOS Big Sur

PR #2827, opened 2021-07-26
Fix Issue 2820

shaomeng:
this PR fixes issue 2820 by implementing another termination condition when doing time-varying advection.

### DIFF
--- a/include/vapor/FlowRenderer.h
+++ b/include/vapor/FlowRenderer.h
@@ -94,10 +94,10 @@ private:
     //
     // Member functions
     //
-    int  _genSeedsRakeUniform(std::vector<flow::Particle> &seeds) const;
-    int  _genSeedsRakeRandom(std::vector<flow::Particle> &seeds) const;
-    int  _genSeedsRakeRandomBiased(std::vector<flow::Particle> &seeds) const;
-    int  _genSeedsFromList(std::vector<flow::Particle> &seeds) const;
+    int _genSeedsRakeUniform(std::vector<flow::Particle> &seeds) const;
+    int _genSeedsRakeRandom(std::vector<flow::Particle> &seeds) const;
+    int _genSeedsRakeRandomBiased(std::vector<flow::Particle> &seeds) const;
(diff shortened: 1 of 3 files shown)


clyne (approved):

clyne:
@shaomeng please resolve clang format issues

StasJ (commented):
This appears to fix the hanging issue by stopping if the number of steps exceeds 10000.
1. What happens if the simulation requires > 10000 steps?
2. In the case that it hangs, how long would the user have to wait for it to get to 10000 steps before it stops?
3. Doesn't the bug still exist and it would just give incorrect results instead of hanging?

shaomeng:
> This appears to fix the hanging issue by stopping if the number of steps exceeds 10000.
> 
>     1. What happens if the simulation requires > 10000 steps?
> 
>     2. In the case that it hangs, how long would the user have to wait for it to get to 10000 steps before it stops?
> 
>     3. Doesn't the bug still exist and it would just give incorrect results instead of hanging?

Good questions, so here's some clarification.

This particular bug (#2820) is caused by one particle that moves so slowly that would require a huge number of steps to advance to the targeting time step. So, what appears to be hanging is actually one particle moving slowly. If you give it enough time, it *will* reach the target. 

Then it comes to *how* to resolve it. I'd think super slow moving particles are hard to avoid 100%, so setting up a cut off limit that would not affect the vast majority of particles could be a reasonable solution. Note that in this reported bug, the velocity field variables have nothing to to with actual velocities, so super slow moving particles are probably easier to occur.

Finally, the termination scheme is set up so that advection needs to proceed at least 10,000 steps or 10X the previous max num of steps. E.g., if a previous "normal" particle went for 5,000 steps, then this problematic particle will go up to 50,000 steps. I think this is a reasonable approach. The caveat is that particles that would require 55,000 or 60,000 steps to finish are terminated prematurely. However, I hope the threshold value (10,000 to start with) is set so that not many particles are affected.

StasJ (approved):

PR #2834, opened 2021-07-26
BOV: Multiple timesteps for a single variable

### LINKED ISSUE #2821: BOV: not supporting implicit multiple time steps
To reproduce: 
1) Open VAPOR.
2) Import 2 BOV files representing 2 time steps: 
`/glade/p/cisl/vast/vapor/Bugs/BOV_Bugs/bug7.bov` and `/glade/p/cisl/vast/vapor/Bugs/BOV_Bugs/bug8.bov`.
3) Create a renderer. Notice that VAPOR shows there's only one time step.

sgpearse:
Fixes #2821 by adding documentation to DCBOV.h.  

The documentation regards how duplicate timestep/variable definitions are handled among imported BOV files.

> //! If multiple files specify a single TIME value for a single variable, the first-read file will be used for that variable at that timestep.

### DIFF
--- a/include/vapor/DCBOV.h
+++ b/include/vapor/DCBOV.h
@@ -51,16 +51,16 @@ class BOVCollection;
 //! following VisIt document: https://github.com/NCAR/VAPOR/files/6341067/GettingDataIntoVisIt2.0.0.pdf
 //!
 //! The following BOV tags are mandatory for Vapor to ingest data:
-//! - DATA_FILE
-//! - DATA_SIZE
-//! - DATA_FORMAT
+//! - DATA_FILE    (type: string)
+//! - DATA_SIZE    (type: three integer values that are >1)
+//! - DATA_FORMAT  (type: string of either INT, FLOAT, or DOUBLE)
 //!
 //! The following BOV tags are optional:
-//! - BRICK_ORIGIN (default: 0., 0., 0.)
-//! - BRICK_SIZE   (default: 1., 1., 1.)
-//! - TIME         (default: 0.)
-//! - VARIABLE     (default: "brickVar")
-//! - BYTE_OFFSET  (default: 0)
+//! - BRICK_ORIGIN (type: three floating point values,   default: 0., 0., 0.)
+//! - BRICK_SIZE   (type: three floating point values,   default: 1., 1., 1.)
+//! - TIME         (type: one floating point value,      default: 0.)
+//! - VARIABLE     (type: one alphanumeric string value, default: "brickVar")
+//! - BYTE_OFFSET  (type: one integer value,             default: 0)
 //!
 //! The following BOV tags are currently unsupported.  They can be included in a BOV header,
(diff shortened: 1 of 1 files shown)


shaomeng:
Scott, can you double check if it's the 1st occurrence, or the last occurrence, being used? 

Also, can you specify this value, and all other values, if they should be int or float in nature? (I.e., are 1.1 and 1.2 considered the same?)

sgpearse:
> Scott, can you double check if it's the 1st occurrence, or the last occurrence, being used?

Yes, it's the first occurrence that is kept.  Once a file is parsed and validated, [I add it to a lookup table](https://github.com/NCAR/VAPOR/blob/721bc8e491eae7c512f92b82e3dc842a1470c028/lib/vdc/BOVCollection.cpp#L271) based on a timestamp and variable name.  If we have a new timestamp/variable pair, we add it to the map.

> Also, can you specify this value, and all other values, if they should be int or float in nature? (I.e., are 1.1 and 1.2 considered the same?)

Sure thing.  I've added [more comments to DCBOV.h](https://github.com/NCAR/VAPOR/blob/721bc8e491eae7c512f92b82e3dc842a1470c028/include/vapor/DCBOV.h#L53), but this will also be added and further explained in the online documentation.

shaomeng:
@sgpearse , I don't think this PR fixes the problem. The problem is, user selects 2 files, but one is discarded, ***silently***. This "fix" will exhibit exactly the same behavior as before. Does this make sense? Happy to discuss if it doesn't.

A proper fix will not silently discard input by a user. Instead, it will either 1) properly process user input as other data importers do, or 2) warn users that not all input files will be processed, and clearly tell users what is chosen to be processed. 
Does this make sense? If not, happy to discuss. 

Among the 2 possible fixes listed above, I suggested 2) because it involves less effort. 1) will produce better software but it will also require more work. Happy to know if a possible 3rd fix exists, but I don't think this PR is currently a fix yet.

clyne:
I'm going back to what I said earlier; I don't think it is unreasonable to silently discard a file if the  time isn't specified. Synthesizing a time step simply assumes that what the user really wanted was to specify time varying data. Moreover, it guesses at a time value. Both of these assumptions may lead to unexpected behavior as well. I'd support leaving the original behavior alone,  documenting the default value of TIME, and providing an example in our documentation that shows how to create a time varying data volume.

shaomeng:
> I'm going back to what I said earlier; I don't think it is unreasonable to silently discard a file if the time isn't specified. 
Silently discarding a user input is bad behavior at least, isn't it? 

Scott has code to display a warning message if 3 other fields were missing in a BOV header; I'd think it's a trivial task to make TIME the 4th field to trigger a warning message. Why confuse the user when you can easily display an informative message?

Maybe @sgpearse can help assess if it is trivial or not.

sgpearse:
@shaomeng I've implemented this per your request in #2837.  Closing this one.

Issue #2833, opened 2021-07-26
2D BOV data


sgpearse:
The BOV reader currently only supports 3D grids.

Enhancement: Add support for 2D grids.

Issue #2826, opened 2021-07-24
BOV crashes when z = 1


shaomeng:
To reproduce:
1. Open VAPOR and import BOV: `/glade/p/cisl/vast/vapor/data/Source/BOV/2D.bov`
2. Create any renderer that would read the data. VAPOR crashes.

PR #2835, opened 2021-07-26
Constrain BOV reader to support 3D grids, and add a check for brick size

sgpearse:
This constrains the BOV reader to 3D data.  I've added an enhancement to support 2D data [here](https://github.com/NCAR/VAPOR/issues/2833).

This also adds a guard to check that the BRICK_SIZE token is >0.

### DIFF
--- a/include/vapor/DCBOV.h
+++ b/include/vapor/DCBOV.h
@@ -16,7 +16,7 @@ namespace VAPoR {
 
 class BOVCollection;
 
-// Example BOV header file, bovA1.bov
+// Example BOV header file, bovA1.bov.
 //
 // # TIME is a floating point value specifying the timestep being read in DATA_FILE
 // TIME: 1.1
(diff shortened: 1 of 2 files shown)


shaomeng:
@sgpearse I'm not sure if this PR solves the problem in issue #2826 . It's true that VAPOR no longer crashes, but VAPOR refuses to take in a volume of dimension (256x512x1) now. My understanding is that with improvements made by John, all grids are considered 3D right now and 256x512x1 is a perfectly legal volume dimension. 

@clyne  can clarify probably.

sgpearse:
@shaomeng - I'm taking a fresh look at this and I think you're right.  I should have a better fix in soon.  Closing this for now.

clyne:
> @sgpearse I'm not sure if this PR solves the problem in issue #2826 . It's true that VAPOR no longer crashes, but VAPOR refuses to take in a volume of dimension (256x512x1) now. My understanding is that with improvements made by John, all grids are considered 3D right now and 256x512x1 is a perfectly legal volume dimension.
> 
> @clyne can clarify probably.

The grid class support really isn't relevant. The DC class - which simply imports data in various formats - provides API for querying which variables are 1D, 2D, 3D, etc. To support grids that contain a dimension of length 1, the DCBOV class will need to know that variables with these dimensions are 2D or 1D variables. The correct behavior for the DCBOV class should probably be so simply remove any dimensions of length 1. E.g. 10x10x1 -> 10x10, and 10x1x10 -> 10x10, and so on.

PR #2836, opened 2021-07-27
Issue2826: BOV support for 2D grids, and add validation for negative grid/brick sizes

### LINKED ISSUE #2826: BOV crashes when z = 1

sgpearse:
Fixes #2826 with 2D grids at DCBOV.cpp:290.

This also adds validation for negative values passed to DATA_SIZE and BRICK_SIZE.  

Note - this required changing the types for those values from size_t to int.  This is because stringstream is used for converting the user's string values to numeric values.  Unfortunately, stringstream does not fail or report an error when converting negative values to size_t.  Therefore, I had to convert to int, and check for negative values during validation.  I don't think this is a problem since grid sizes shouldn't hit INT_MAX any time soon.

### DIFF
--- a/lib/vdc/BOVCollection.cpp
+++ b/lib/vdc/BOVCollection.cpp
@@ -36,7 +36,7 @@ const std::string BOVCollection::DATA_COMPONENTS_TOKEN = "DATA_COMPONENTS";
 
 const std::array<double, 3> BOVCollection::_defaultOrigin = {0., 0., 0.};
 const std::array<double, 3> BOVCollection::_defaultBrickSize = {1., 1., 1.};
-const std::array<size_t, 3> BOVCollection::_defaultGridSize = {0, 0, 0};
+const std::array<int, 3>    BOVCollection::_defaultGridSize = {0, 0, 0};
 const DC::XType             BOVCollection::_defaultFormat = DC::XType::INVALID;
 const std::string           BOVCollection::_defaultFile = "";
 const std::string           BOVCollection::_defaultVar = "brickVar";
@@ -247,6 +247,9 @@ int BOVCollection::_validateParsedValues()
     if (_tmpBrickSize != _brickSize && _brickSizeAssigned == true)
         return _inconsistentValueError(BRICK_SIZE_TOKEN);
     else {
+        for (size_t i = 0; i < _tmpBrickSize.size(); i++) {
+            if (_tmpBrickSize[i] < 0) return _invalidValueError(BRICK_SIZE_TOKEN);
+        }
         _brickSize = _tmpBrickSize;
         _brickSizeAssigned = true;
     }
@@ -341,7 +344,7 @@ std::string BOVCollection::GetTimeDimension() const { return _timeDimension; }
 
 std::vector<float> BOVCollection::GetUserTimes() const { return _times; }
 
-std::array<size_t, 3> BOVCollection::GetDataSize() const { return _gridSize; }
+std::array<int, 3> BOVCollection::GetDataSize() const { return _gridSize; }
 
 DC::XType BOVCollection::GetDataFormat() const { return _dataFormat; }
 
--- a/lib/vdc/BOVCollection.cpp
+++ b/lib/vdc/BOVCollection.cpp
@@ -36,7 +36,7 @@ const std::string BOVCollection::DATA_COMPONENTS_TOKEN = "DATA_COMPONENTS";
 
 const std::array<double, 3> BOVCollection::_defaultOrigin = {0., 0., 0.};
 const std::array<double, 3> BOVCollection::_defaultBrickSize = {1., 1., 1.};
-const std::array<int, 3>    BOVCollection::_defaultGridSize = {0, 0, 0};
+const std::array<size_t, 3> BOVCollection::_defaultGridSize = {0, 0, 0};
 const DC::XType             BOVCollection::_defaultFormat = DC::XType::INVALID;
 const std::string           BOVCollection::_defaultFile = "";
 const std::string           BOVCollection::_defaultVar = "brickVar";
@@ -76,7 +76,9 @@ BOVCollection::BOVCollection()
     _dataFiles.clear();
     _times.clear();
     _gridSize = _defaultGridSize;
-    _tmpGridSize = _defaultGridSize;
+    _tmpGridSize[0] = (int)_defaultGridSize[0];
+    _tmpGridSize[1] = (int)_defaultGridSize[1];
+    _tmpGridSize[2] = (int)_defaultGridSize[2];
     _brickOrigin = _defaultOrigin;
     _tmpBrickOrigin = _defaultOrigin;
     _brickSize = _defaultBrickSize;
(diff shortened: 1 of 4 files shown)


shaomeng:
> Note - this required changing the types for those values from size_t to int. This is because stringstream is used for converting the user's string values to numeric values. Unfortunately, stringstream does not fail or report an error when converting negative values to size_t. Therefore, I had to convert to int, and check for negative values during validation. I don't think this is a problem since grid sizes shouldn't hit INT_MAX any time soon.

Scott, I agree that value on a single dimension is unlikely to exceed INT_MAX, but the total number of values is pretty easily to exceed INT_MAX. In other words, the following code will result in int overflows:
```
int dim_x = userInput(); 
int dim_y = userInput(); 
int dim_z = userInput(); 
size_t last_pencil_end = dim_x * dim_y * dim_z; // int overflow can easily happen
```
Does it make sense that there's a real danger? Happy to explain in more detail if not.

So, you should store dimension values in size_t type as soon as stringstream gives you a number, and you validated that the number is non-negative. Subsequently, you perform any  calculations in size_t type to avoid the scenario demonstrated above. Happy to discuss if you have questions.

clyne (commented):
This PR does not address the problem of failing to support 2D grids. For better or worse VAPOR treats variables with different topological dimensions as different animals. E.g. 3D variables can be volume rendered, and sliced, etc. 2D variables can be used as height fields, etc. A grid with a dimension length of one has a degenerate dimension that should be removed. I.e. a 10x10x1 grid should be 10x10. Moreover, a grid with dimensions 10x1x10 is also a 10x10 grid. You could argue that the issue of degenerate dimensions should be handled elsewhere in VAPOR, but at the moment it is not. Thus for the BOV reader to support 2D variables it should remove any dimensions of length 1. Otherwise a variable defined as having a grid of 128x128x1, for example, will be treated as a 3D variable and you won't be able to contour it, use it as a height field, etc. 

I would be fine deferring action on this and evaluating where the best place is to handle degenerate grid dimensions.

Re supporting (preventing) negative dimension lengths: this approach doesn't make sense to me. A grid can't have a negative length. It's meaningless. You've changed all of the internal representation, and even the public API, to use ints to represent a quantity that can't be negative. Why not simply modify parseHeader to throw an error for negative values?

sgpearse:
@shaomeng thanks for the chat earlier.  I've made the changes we discussed and stepped through the parser to verify that values are being read as integers in _tmpGridSize, then get cast to size_t in the official _gridSize variable (after being validated).

> Re supporting (preventing) negative dimension lengths: this approach doesn't make sense to me. A grid can't have a negative length. It's meaningless. You've changed all of the internal representation, and even the public API, to use ints to represent a quantity that can't be negative. Why not simply modify parseHeader to throw an error for negative values?

@clyne The variables _tmpGridSize and _gridSize were previously both size_t.  The problem arose when one of the dimension lengths was negative (-10 10 10).  I am using `std::stringstream` to convert strings into numeric values.  Unfortunately `std::stringstream` doesn't throw an error when converting the string "-10" to size_t.  It just assigns a huge positive number that can't be validated, because huge positive numbers are still considered valid.

The approach in the recent push reads the DATA_SIZE values as integers and stores them in _tmpGridSize.  Then during validation, we check if all values > 0.  If so, we cast those integer values into the official copy, `std::array<size_t,3> _gridSize`.

lib/vdc/BOVCollection.cpp
```diff
@@ -247,6 +251,9 @@ int BOVCollection::_validateParsedValues()
     if (_tmpBrickSize != _brickSize && _brickSizeAssigned == true)
         return _inconsistentValueError(BRICK_SIZE_TOKEN);
     else {
+        for (size_t i = 0; i < _tmpBrickSize.size(); i++) {
+            if (_tmpBrickSize[i] < 0) return _invalidValueError(BRICK_SIZE_TOKEN);
```
shaomeng, line 255: Q1: is brick size equal to 0 OK?

Q2: You're comparing a double with an integer. What's the expected behavior: is it to cast everything to double and then perform the comparison, or to cast everything to integer and then perform the comparison?

sgpearse:
@sgpearse will throw an error on 2D (10x10x1, etc) grids.

lib/vdc/BOVCollection.cpp
sgpearse: 1) Yes
2) No problem, I ditched the implicit cast in the current HEAD of this branch.

sgpearse:
The current HEAD of this branch now does not use the implicit cast when doing comparisons.  Additionally, we now throw errors when the DATA_SIZE token does not have all values > 1.

lib/vdc/BOVCollection.cpp
```diff
@@ -216,12 +218,14 @@ int BOVCollection::_parseHeader(std::ifstream &header)
 int BOVCollection::_validateParsedValues()
 {
     // Validate grid dimensions
-    if (_tmpGridSize[0] < 1 || _tmpGridSize[1] < 1 || _tmpGridSize[2] < 1)
+    if (_tmpGridSize[0] < 2 || _tmpGridSize[1] < 2 || _tmpGridSize[2] < 2)
         return _invalidDimensionError(GRID_SIZE_TOKEN);
-    else if (_tmpGridSize != _gridSize && _gridSizeAssigned == true)
+    else if ((_tmpGridSize[0] != _gridSize[0] || _tmpGridSize[1] != _gridSize[1] || _tmpGridSize[2] != _gridSize[2]) && _gridSizeAssigned == true)
```
shaomeng, line 223: Can you use parenthesis to group logical operations together so the overall logic is clearer? 

In other words, does a compiler treat `(a || b || c && d)` the same as you think it would?

lib/vdc/DCBOV.cpp
```diff
@@ -286,7 +286,8 @@ template<class T> void DCBOV::_generateCoordinates(int dim, const vector<size_t>
     std::array<double, 3> origin = _bovCollection->GetBrickOrigin();
     std::array<double, 3> brickSize = _bovCollection->GetBrickSize();
 
-    double increment = brickSize[dim] / (dataSize[dim] - 1);
+    double denom = (dataSize[dim] - 1);
+    double increment = denom == 0 ? 0 : brickSize[dim] / denom;
```
shaomeng, line 290: Scott, do you remember we were talking about making your intentions clear to the compiler (and to your code reviewers) by explicitly expressing how you would like type conversions to happen? 🤔🤔🤔

shaomeng (changes requested):

lib/vdc/BOVCollection.cpp
sgpearse: I blame this on clang-format.  They way I originally wrote it was as so:

```
    else if (( _tmpGridSize[0] != _gridSize[0] 
                || _tmpGridSize[1] != _gridSize[1] 
                || _tmpGridSize[2] != _gridSize[2]) 
              && _gridSizeAssigned == true)

```

Yes, the logic does what's intended, which is to make sure all _tmpGridSize values match those of _gridSize; and then that gridSize has also been assigned.  If it hasn't been assigned, then _tmpGridSize is allowed to differ from the default value of _gridSize.

lib/vdc/DCBOV.cpp
sgpearse: Missed that one.  Fixed.

Are you using a static analyzer to find these problems?  When I run clang-tidy on this file it reports no warnings.

lib/vdc/BOVCollection.cpp
clyne: The look adequately and unambiguously  parenthesized to me. Not sure what @shaomeng is asking for.

shaomeng: My bad, I missed the inner parenthesis. Now I see the logic very clear. Sorry guys. (Human reviewing aren't reliable sometimes...)

lib/vdc/DCBOV.cpp
shaomeng: Just spotted it on this github page. clang-tidy sounds like something could be helpful to us :)

shaomeng (approved):

clyne (approved):

PR #2837, opened 2021-07-27
Issue2821: Tighter constraints on TIME token, and relevant documentation

### LINKED ISSUE #2821: BOV: not supporting implicit multiple time steps

sgpearse:
Fixes #2821 by making the TIME token mandatory in each file, and throwing an error when duplicate files refer to a single variable at the same timestep.

Also adds documentation on how this requirement for TIME, and adds info on default values for the optional tokens in BOV files.

### DIFF
--- a/lib/vdc/BOVCollection.cpp
+++ b/lib/vdc/BOVCollection.cpp
@@ -128,7 +128,7 @@ int BOVCollection::Initialize(const std::vector<std::string> &paths)
 
             rc = _populateDataFileMap();
             if (rc < 0) {
-                SetErrMsg("Duplicate time entries found in BOV files.  Each file must uniquely describe one variable, at one timestep.");
+                SetErrMsg("Problem indexing data files.");
                 return -1;
             }
         } else {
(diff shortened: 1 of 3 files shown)


lib/vdc/BOVCollection.cpp
```diff
@@ -123,8 +124,13 @@ int BOVCollection::Initialize(const std::vector<std::string> &paths)
             if (_dataFile == _defaultFile) { return _missingValueError(DATA_FILE_TOKEN); }
             if (_dataFormat == _defaultFormat) { return _missingValueError(FORMAT_TOKEN); }
             if (_gridSize == _defaultGridSize) { return _missingValueError(GRID_SIZE_TOKEN); }
+            if (_time == _defaultTime) { return _missingValueError(TIME_TOKEN); }
 
-            _populateDataFileMap();
+            rc = _populateDataFileMap();
```
clyne, line 129: It would be better to have the error message inside of _populateDataFileMap() for a couple of reasons:

1. Our error reporting convention states that functions that return a negative int on failure should report an error with SetErrMsg() (prefer a boolean return status for functions that do NOT report a message).
2. Better future proofing: if some other error condition gets added to _populateDataFileMap() later the current error message may not make sense.

clyne (approved):

shaomeng (approved):

PR #2845, opened 2021-08-04
Provide a data reader for UGRID files

### LINKED ISSUE #2714: Provide a data reader for UGRID files
Provide a data reader capable of supporting the proposed UGRID extensions for unstructured grids:

https://ugrid-conventions.github.io/ugrid-conventions/

clyne:
Fixes #2714 

This PR provides a data importer for [UGRID](https://ugrid-conventions.github.io/ugrid-conventions/) files. At present the reader handles all of the supported UGRID topologies except "Fully unstructured 3D" (general 3D unstructured meshes are not supported internally by VAPOR).


include/vapor/DC.h
```diff
@@ -641,6 +641,12 @@ class VDF_API DC : public Wasp::MyBase {
         //
         size_t GetTopologyDim() const;
 
+        //! Generate a mesh name from a list of strings. The 
+        //! generated name is simply the join of the elements of \p s
+        //! using 'x' as a delimiter
+        //
+        static string MakeMeshName(std::vector<string> s);
```
shaomeng, line 648: Why does this function have to be static?

lib/vdc/DCUGRID.cpp
```diff
@@ -0,0 +1,630 @@
...
+    if (values.size()) { v = values[0]; }
+}
+
+// Read a netcdf attribute of type string
+//
+void getAttString(NetCDFCFCollection *ncdfc, string varName, string attName, string &v)
+{
+    v = "";
+    ncdfc->GetAtt(varName, attName, v);
+}
```
shaomeng, line 67: It's considered easier to read code if your function name is "getSomething" and your function return type is "something." These two functions seem perfectly fit in that pattern.

lib/vdc/NetCDFCFCollection.cpp
```diff
@@ -14,6 +14,17 @@ using namespace VAPoR;
 using namespace Wasp;
 using namespace std;
 
+namespace {
+vector<string> make_unique(vector<string> v)
+{
+    sort(v.begin(), v.end());
```
shaomeng, line 20: STD has a function named `make_unique` so I don't think it's a good idea to make your custom function the same name.

clyne:
@sgpearse this PR is failing the data manager regression tests. From what I can tell it is because this version of the code reports an additional coordinate variable, "zf", in the file 24Maycontrol.04000.000000.nc. The new behavior is correct. Not sure how to go about updating the baselines after this PR is accepted.

include/vapor/DC.h
clyne: So it can be used without instantiating a class as is done in DCCF.cpp

lib/vdc/DCUGRID.cpp
clyne: done

lib/vdc/NetCDFCFCollection.cpp
clyne: done

include/vapor/DC.h
shaomeng: This this the use case that is required occasionally: a helper function that takes input and produces an output, and the operation is not necessarily associated with a particular class. 

There are at least 2 ways to achieve the goal: 1) make it a function under a particular namespace. E.g.: many functions under the `std::` namespace in C++. 2) declare the function static and insert it into a particular class.

Way 2) is a little messy, because the static functions will out-live the class object itself, which is really against the core ideas of object oriented programming. Way 1) is adapted by the standard C++ to meet this particular requirement, so I suspect it's got other advantages. 

I suggest adopting Way 1) here, but it's ultimately the decision of the DC author.

apps/vaporgui/MainForm.cpp
```diff
@@ -888,6 +892,11 @@ void MainForm::_createFileMenu()
     _dataImportMPAS_Action->setToolTip("Specify one or more MPAS output files to import into the "
                                        "current session");
 
+    _dataImportUGRID_Action = new QAction(this);
```
StasJ, line 895: There is a simplified method for adding data importer menu items. The DCP importer is an example.

lib/vdc/DataMgr.cpp
```diff
@@ -2620,10 +2649,15 @@ void DataMgr::_ugrid_setup(const DC::DataVar &var, std::vector<size_t> &vertexDi
...
         faceOffset = auxvar.GetOffset();
-    } else {
-        VAssert(!"NodeFaceVar Required");
-        faceOffset = 0;
     }
+#ifdef DEAD
+}
+else
+{
+    VAssert(!"NodeFaceVar Required");
```
StasJ, line 2657: Should this be a fatal error?

apps/vaporgui/MainForm.cpp
clyne: Done

lib/vdc/DataMgr.cpp
clyne: This has been removed.

include/vapor/DC.h
clyne: This method has no purpose that isn't associated with the DC class (i.e. it will NOT out-live the class. That is why it was put into the same namespace. Putting it in some other made up namespace would only cause confusion. I'm inclined to leave it where it is.

StasJ (approved):

include/vapor/DC.h
shaomeng: From a design perspective, if all purposes of a function is associated with a particular class, then wouldn't it make sense to have this function be a member function of that class?

include/vapor/DCUGRID.h
```diff
@@ -0,0 +1,84 @@
...
+protected:
+    int initialize(const vector<string> &paths, const std::vector<string> &options) override;
+
+    int initAuxilliaryVars(NetCDFCFCollection *ncdfc, std::map<string, DC::AuxVar> &auxVarsMap) override;
+
+    int initDataVars(NetCDFCFCollection *ncdfc, std::map<string, DC::DataVar> &dataVarsMap) override;
+
+    int initMesh(NetCDFCFCollection *ncdfc, std::map<string, DC::Mesh> &meshMap) override;
+
+    string getMapProjection() const override { return ("+proj=eqc +ellps=WGS84 +lon_0=0.0 +lat_0=0.0"); }
```
shaomeng, line 46: How about we put the function implementation in the `.cpp` file?

Otherwise, making a change to the function body in the header file will cause a re-compilation of almost the entire project.

lib/vdc/DCUGRID.cpp
```diff
@@ -0,0 +1,627 @@
...
+    if (_faceNodeConnectivityMap.find(f->GetVarname()) == _faceNodeConnectivityMap.end()) { return (DC::Read(fd, data)); }
+
+    const vector<int> &connVar = _faceNodeConnectivityMap[f->GetVarname()];
+    std::copy_n(connVar.data(), connVar.size(), data);
+    return (0);
+}
+
+int DCUGRID::ReadRegion(int fd, const vector<size_t> &min, const vector<size_t> &max, int *data)
+{
+    DC::FileTable::FileObject *f = _fileTable.GetEntry(fd);
```
shaomeng, line 590: Does this pointed object need to be recycled (freed or closed)? If so, why not put it in a unique_ptr ?

lib/vdc/DataMgr.cpp
```diff
@@ -3131,58 +3148,83 @@ int DataMgr::_getVar(size_t ts, string varname, int level, int lod, float *data)
...
+    if (!lons.size()) return;
+
+    size_t nx = dims[0];
+    size_t ny = dims.size() > 1 ? dims[1] : 1;
+    for (size_t j = 0; j < ny; j++) {
+        GeoUtil::UnwrapLongitude(lons.begin() + (j * nx), lons.begin() + (j * nx) + nx);
+        GeoUtil::ShiftLon(lons.begin() + (j * nx), lons.begin() + (j * nx) + nx);
+    }
+    min = *(std::min_element(lons.begin(), lons.end()));
+    max = *(std::max_element(lons.begin(), lons.end()));
```
shaomeng, line 3165: It seems you can use `std::minmax_element()` to avoid one loop of the array.
https://en.cppreference.com/w/cpp/algorithm/minmax_element

The same applies to code 20 lines down.

include/vapor/DCUGRID.h
clyne: Putting inexpensive getters in .h files is done throughout the code base, and widely done in practice. What else would be gained by moving the definition other than saving a few microseconds (at most) during compilation?

lib/vdc/DCUGRID.cpp
clyne: Because the memory referenced is used outside of the scope of this method.

include/vapor/DCUGRID.h
shaomeng: I'm afraid the save is more than microseconds. 
-- With implementation being in the header, every file that depends on DC will be re-compiled. That's almost the entire VAPOR, and that compilation is in the order of minutes. 
-- With implementation being in the body, only DC is re-compiled and everybody else just need to re-link, which takes microseconds. 

Plus, it really makes the header look tidy and easier to read when you hide details away.

lib/vdc/DCUGRID.cpp
shaomeng: Ok, as long as it is recycled by someone at some point.

lib/vdc/DataMgr.cpp
clyne: Done.

include/vapor/DCUGRID.h
clyne: > I'm afraid the save is more than microseconds.
> -- With implementation being in the header, every file that includes this header will be re-compiled. That's almost the entire VAPOR, and that compilation is in the order of minutes.
> -- With implementation being in the body, only DC is re-compiled and everybody else just need to re-link, which takes microseconds.
> 

I don't believe that the compilation time will change significantly if the definition is moved. You'll have to do an experiment and convince me with the results :-)

> Plus, it really makes the header look tidy and easier to read when you hide details away.

That's your personal opinion. 

@shaomeng I am not just being stubborn.  I'm making a point. Changes requested by reviewers should have a meaningful impact to the code in terms of performance, readability, robustness, future-proofing, following best practices, or simply following our own conventions. I don't want developers to have to change their code simply because a reviewer has an opinion on how things should look that may or may not agree with the opinion of others. That just waists everyone's time and does not lead to better code.

shaomeng: Regarding compilation time, I'm *not* saying that the 1st time you compile the project there will be a significant difference. I'm saying that when somebody debugs the code, for example change the function implementation or simply insert a print statement, function body being in the header will trigger all the downstream files to be re-compiled again. That re-compile of all the downstream files cost minutes, and that cost is totally avoidable. 

I can tell my personal experience debugging a crashing problem in the last release cycle. I needed to insert a few print statements in some grid classes. Guess how much time it takes to re-compile a few print statements? Minutes. Why? Almost the entire project is re-compiled, because almost every file includes a grid class.

shaomeng: Regarding where to put function implementations, I don't think it's my personal opinion. It's well established C++ practice, with *real* benefits (e.g. the case I raised above). Moreover, it's stated in VAPOR coding convention.

shaomeng: Regarding opinion or not and in general how to review code and more general how to produce better code, we'll probably need a deeper discussion. There is definitely great discrepancy between one review and another review. For example, I was asked to not use trailing returns in a PR because it was thought by some developers that trailing returns are difficult to read.

clyne: > Regarding where to put function implementations, I don't think it's my personal opinion. It's well established C++ practice, with _real_ benefits (e.g. the case I raised above). Moreover, it's stated in VAPOR coding convention.

I should have been more clear about what is reasonable to put into a header file. I was not suggesting that complicated code which might require debugging (print statements) is reasonable to put in a .h file. But simple statements such as those that define many getters functions should not require debugging. The case in point returns a string literal. Something is really going off the rails if you have to put a print statement inside a single line function that returns a constant value :-). Moreover, I'm not aware of any well established C++ practice that recommends against simple expressions in .h files. Do you have a reference for this? Qt, Vtk, and Google Open Source all have lots of .h files with one-line functions (and more) defined in them. The Google Source Chromium guide simply says "Avoid this [definitions] in header files except for cheap non-virtual getters and setters. ", which seems reasonable to me.

clyne: > Regarding where to put function implementations, I don't think it's my personal opinion. It's well established C++ practice, with _real_ benefits (e.g. the case I raised above). Moreover, it's stated in VAPOR coding convention.

I don't see anything in our coding [conventions](https://github.com/NCAR/VAPOR/wiki/Vapor-Coding-Convention) about this issue.

clyne: > ause it was thought by some developers that trailing returns are difficult to read. Is that an opinion or not?

I remember this discussion and I think at the time we were not aware that trailing returns actually had some practical benefit (not just syntactical sugar): they allow return types to be auto deduced by the compiler. I'd certainly be willing to revisit the issue.

shaomeng: Well, the separation of .h and .cpp files is something taught at every college class, for example, [this one](http://websites.umich.edu/~eecs381/handouts/CppHeaderFileGuidelines.pdf). I can only imagine that some guidelines don't specify it in explicit terms is because it's well accepted that there's no need to specify it.

Coming back to that Google document, it specifies non-virtual functions. But the function in question here has `override` in its signature, indicating a virtual function.

Lastly, I think it's a bad assumption to make that "I don't need to come back and debug this function." As a matter of fact, I personally needed to change function bodies living in headers for debugging purposes in the last release cycle.

shaomeng: > I don't see anything in our coding [conventions](https://github.com/NCAR/VAPOR/wiki/Vapor-Coding-Convention) about this issue.

I don't think this sentence means anything else given its context.
`Any public class should generally be declared in its own header file, and defined in its own source file.` 
There's no mention "if you meet condition 1, 2, and 3, then you can leave the definition in the header.

shaomeng: > I remember this discussion and I think at the time we were not aware that trailing returns actually had some practical benefit (not just syntactical sugar): they allow return types to be auto deduced by the compiler. I'd certainly be willing to revisit the issue.

I didn't try to raise attention to the benefit (or lack of benefit) of trailing returns (though I welcome a discussion on it). I was trying to raise the point of consistency when exercising code review activities. I'd agree that personal opinion isn't a valid basis for change requests. I also think consistently applying a certain standard is important.

include/vapor/DC.h
clyne: It is a member function of a class. It's a static member function. One of the intended use cases for static member functions is to allow their use without instantiating the class.

include/vapor/DCUGRID.h
clyne: @shaomeng, I'm sorry but you haven't convinced me that there is a good reason to prohibit simple member functions from being defined in a .h file. There appears to be no practical reason for such a prohibition. I don't really believe that there is a need for debugging a function that does no more than return a literal of even a member variable., do you? You've mentioned having trouble debugging more complex functions defined in .h files. If so, I'd suggest submitting an issue, or even a PR, to move those troublesome functions. There also does not appear to be any consensus amongst the C++ community that I can see. You've claimed that every college class teaches not to do this, but that claim seems wholly unsubstantiated. The existence of a single example doesn't do much to support this argument. On the contrary I've identified three large, modern C++ development efforts that do not abide by any such convention (add a fourth, Intel OSPRay to the list as well). And, FWIW the Google C++ Style Guide recommends limiting functions defined in .h files to 10 lines of code. 

If you really think the VAPOR code base would be improved by adopting a convention on this that we all abide by, you should propose it to the group as a change to our coding conventions. If you can convince the majority I am happy to comply (and will expect others to as well).

shaomeng: I'm sorry John, I think the VAPOR coding convention has a clear specification, but my interpretation could be wrong. What do you think this sentence in it means? 
`Any public class should generally be declared in its own header file, and defined in its own source file.`
If it doesn't mean every function body goes to .cpp files, I'm happy to learn what the guideline actually is regarding what to put in headers and in .cpp files. 

(Note, what this specification means, and if real code abides to it, are two different issues.)

clyne: Sam, there is a qualifier in the statement that may be leading to the confusion: "generally". In a context like this "generally" means that most of the time the rule applies, but "generally" provides the flexility to ignore it. Lacking specificity about when the rule can be ignored it is left up to the judgement of the reader.

shaomeng: Well, talking about "generally." When judging how and when "generally" applies, my understanding is that there should be some strong and evident benefits to support an occasional break of the rule. That strong and evident benefit should be in the nature of enhancing the functionality, robustness, portability (and alike) of the program. One example of such scenario is templated function.

The justification to apply "generally," however, should *not* be: "I don't want to spend 5 extra seconds to abide to the rules." Because, that's almost the definition of abusing the flexibility provided by the rules. Do you think this is a reasonable understanding of "generality" in that sentence, and how to interpret rules in general?

Regard to this specific case, I haven't seen any benefit to the software that you provide to support this occasion of breaking the rule. If I missed it, maybe you could raise it again?

clyne: @shaomeng I'm not debating this anymore. Since there does not appear to be any consensus from the C++ community if you want to ban all method definitions from .h files you'll need to convince the group to clearly state this in our coding guidelines.

PR #2847, opened 2021-08-10
udpate smoke tests with zf variable

sgpearse:
Adds the zf coordinate variable to our smoke test baseline file for NetCDF-CF files.

Note - These baseline files were build without OpenMP.  I'm currently experiencing linker errors when running from source with OpenMP.  My installers are running fine though.  Not sure what's going on at the moment.


clyne (commented):
@sgpearse since we're making changes to these tests already, how about we add an option to silence the Elapsed time reporting so that we're not confronted with confusing diffs every time we need to change these regression tests?

clyne (approved):

PR #2856, opened 2021-09-24
Vapor API WIP

StasJ:
This is an initial version of the Vapor API. It provides a basis to utilize the vapor through a scripting interface. It will also provide the ability to remove the majority of the control code from the GUI. The current functionality allows for rendering session files. For demo purposes, it compiles as an executable rather than as a library. In `vapi/main.cpp`, you can specify a session file to load as well as an image to output.

Note: you need to enable BUILD_PYTHON in cmake to build. Also, the test driver looks for a file named 'session.vs3' and outputs a file named 'out.png'


apps/CMakeLists.txt
```diff
@@ -28,6 +28,10 @@ if (BUILD_GUI)
 	endif()
 endif()
 
+if (BUILD_PYTHON)
+    add_subdirectory (vapi)
```
clyne, line 32: I assume everything except main.cpp in apps/vapi will moved to the lib directory?

apps/vapi/GLContextProviderMacOS.mm
```diff
@@ -0,0 +1,46 @@
+#include "GLContextProviderMacOS.h"
```
clyne, line 1: It looks like the source file in Objective-C. It would be good to add a comment to that effect as well as a brief explanation as to why not C++

apps/vapi/Log.cpp
```diff
@@ -0,0 +1,22 @@
+#include "Log.h"
```
clyne, line 1: Do we really need another logging facility? Can we not make use of the existing one that we have, and modify it if needed? It would be preferable to be consistent throughout the codebase.

apps/vapi/RenderManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+#define INCLUDE_DEPRECATED_LEGACY_VECTOR_MATH
+#include <vapor/LegacyVectorMath.h>
+
+
+using namespace VAPoR;
+
+RenderManager::RenderManager(ControlExec *ce) : _controlExec(ce) {}
+
+RenderManager::~RenderManager() { delete _glManager; }
```
clyne, line 21: _glManager is not is not allocated in the constructor and  may not be initialized, no?

apps/vapi/RenderManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+
+
+int RenderManager::Render(String imagePath)
+{
+    //    GL_ERR_BREAK();
+    if (!_glManager) _glManager = new GLManager;
+
+    static bool temp = false;
+    if (!temp) {
```
clyne, line 197: Can we use a member variable here?

apps/vapi/RenderManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+        // TODO: REMOVE TEMP
+    }
+
+    Framebuffer defaultFB;
+    defaultFB.Generate();
+    defaultFB.SetSize(600, 480);
+    defaultFB.MakeRenderTarget();
+
+    getViewpointParams()->SetWindowSize(600, 480);
+    //    getViewpointParams()->SetValueLong(ViewpointParams::UseCustomFramebufferTag, "", true);
```
clyne, line 209: Can this be removed?

apps/vapi/RenderManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+    _glManager->matrixManager->MatrixModeProjection();
+    _glManager->matrixManager->PushMatrix();
+    setUpProjMatrix();
+
+    _glManager->matrixManager->MatrixModeModelView();
+    _glManager->matrixManager->PushMatrix();
+    setUpModelViewMatrix();
+
+    int rc = _controlExec->EnableImageCapture(imagePath, getWinName());
```
clyne, line 221: We're not handling error messages generated from MyBase::SetErrMsg() here.

apps/vapi/Session.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    if (!p->GetProjectionString().empty()) {
+        options.push_back("-proj4");
+        options.push_back(p->GetProjectionString());
+    }
+
+    if (name.empty()) name = ControlExec::MakeStringConformant(FileUtils::Basename(files[0]));
+
+    OpenDataset(name, format, files, options);
+    return -1;
```
clyne, line 79: Should be 0, not -1. Correct?

apps/vapi/Session.cpp
```diff
@@ -0,0 +1,154 @@
...
+        LogWarning("Nothing to render");
+        return -1;
+    }
+
+    return _renderManager->Render(imagePath);
+}
+
+void Session::SetTimestep(int ts) { NavigationUtils::SetTimestep(_controlExec, ts); }
+
+void Session::loadAllParamsDatasets()
```
clyne, line 120: This method doesn't seem to handle error conditions (e.g. data sets that fail to open)

clyne (commented):
Nice work, Stas. A few comments and questions:

1. Presumably most of the new source file will move from apps/vapi to somewhere in 'lib'?
2. Is there a reason that we need yet another logging facility? For the sake of consistency it would certainly be preferable to find a way to make use of our existing error reporting mechanism (even if some refactoring is needed).
3. Given that this is work in progress would it make sense to create a feature branch rather than merging into main? I don't think there is any harm in merging this PR into main because it doesn't touch the existing code in any significant way. But as we move forward that may change. What are your thoughts, @sam, @stasj?

apps/vapi/Session.cpp
StasJ: Fixed, thanks.

apps/vapi/Session.cpp
StasJ: Yes, fixed.

apps/vapi/RenderManager.cpp
StasJ: They are configured to automatically output to STDERR.

apps/vapi/RenderManager.cpp
StasJ: Yes, but I left this in intentionally for later. It will either be used or removed by the final version.

apps/vapi/RenderManager.cpp
StasJ: This is some temporary code so the demo works for the WIP and there is a TODO reminder to remove it 2 lines down. It will be removed later.

apps/vapi/RenderManager.cpp
StasJ: Fixed, thanks.

apps/vapi/Log.cpp
StasJ: I added this because I believe it is a better fit for a python API than `WASP::MyBase` which works better for a GUI application. I can remove it eventually but I'd prefer to leave it during development as it has features that make debugging easier.

apps/vapi/GLContextProviderMacOS.mm
StasJ: The native macOS libraries are written Objective-C. I've added a comment in that regard.

apps/CMakeLists.txt
StasJ: Yes, I will move it later.

StasJ:
> Nice work, Stas. A few comments and questions:
> 
> 1. Presumably most of the new source file will move from apps/vapi to somewhere in 'lib'?

Yes, it is currently acting as a command line utility which is why I placed it in apps but I will move it eventually.

> 2. Is there a reason that we need yet another logging facility? For the sake of consistency it would certainly be preferable to find a way to make use of our existing error reporting mechanism (even if some refactoring is needed).

I added this because I believe it is a better fit for a python API than WASP::MyBase which works better for a GUI application. I can remove it eventually but I'd prefer to leave it during development as it has features that make debugging easier.

> 3. Given that this is work in progress would it make sense to create a feature branch rather than merging into main? I don't think there is any harm in merging this PR into main because it doesn't touch the existing code in any significant way. But as we move forward that may change. What are your thoughts, @sam, @StasJ?

I'm fine with doing either. We could just close the PR once the comments are addressed and have this be the feature branch.

apps/vapi/RenderManager.cpp
```diff
@@ -18,7 +18,10 @@ using namespace VAPoR;
 
 RenderManager::RenderManager(ControlExec *ce) : _controlExec(ce) {}
 
-RenderManager::~RenderManager() { delete _glManager; }
+RenderManager::~RenderManager()
+{
+    if (_glManager) delete _glManager;
```
clyne, line 23: Doesn't _glManager need to be explicitly initialized by the constructor?

StasJ: The constructor isn't currently guaranteed to have an OpenGL context.

clyne: Yes, i know. Doesn't it need to be zeroed (set to nulptr) in the constructor?

clyne:
> > Nice work, Stas. A few comments and questions:
> > 
> > 1. Presumably most of the new source file will move from apps/vapi to somewhere in 'lib'?
> 
> Yes, it is currently acting as a command line utility which is why I placed it in apps but I will move it eventually.
> 
> > 1. Is there a reason that we need yet another logging facility? For the sake of consistency it would certainly be preferable to find a way to make use of our existing error reporting mechanism (even if some refactoring is needed).
> 
> I added this because I believe it is a better fit for a python API than WASP::MyBase which works better for a GUI application. I can remove it eventually but I'd prefer to leave it during development as it has features that make debugging easier.
> 
> > 1. Given that this is work in progress would it make sense to create a feature branch rather than merging into main? I don't think there is any harm in merging this PR into main because it doesn't touch the existing code in any significant way. But as we move forward that may change. What are your thoughts, @sam, @StasJ?
> 
> I'm fine with doing either. We could just close the PR once the comments are addressed and have this be the feature branch.

@StasJ, what I'm hoping to avoid is having to re-review the same code later, but at the same time I understand the need to have some of the code be a WIP that should be cleaned up later. I also don't want us to rely on anyone's memory to go back and remove temporary code, commented out code etc.  Perhaps the way to proceed here is to open an issue that identifies the cleanup items, and get this PR merged. Whether it should be merged to main or to a feature branch is another topic. Thoughts on this?

StasJ:
> > > Nice work, Stas. A few comments and questions:
> > > 
> > > 1. Presumably most of the new source file will move from apps/vapi to somewhere in 'lib'?
> > 
> > 
> > Yes, it is currently acting as a command line utility which is why I placed it in apps but I will move it eventually.
> > > 1. Is there a reason that we need yet another logging facility? For the sake of consistency it would certainly be preferable to find a way to make use of our existing error reporting mechanism (even if some refactoring is needed).
> > 
> > 
> > I added this because I believe it is a better fit for a python API than WASP::MyBase which works better for a GUI application. I can remove it eventually but I'd prefer to leave it during development as it has features that make debugging easier.
> > > 1. Given that this is work in progress would it make sense to create a feature branch rather than merging into main? I don't think there is any harm in merging this PR into main because it doesn't touch the existing code in any significant way. But as we move forward that may change. What are your thoughts, @sam, @StasJ?
> > 
> > 
> > I'm fine with doing either. We could just close the PR once the comments are addressed and have this be the feature branch.
> 
> @StasJ, what I'm hoping to avoid is having to re-review the same code later, but at the same time I understand the need to have some of the code be a WIP that should be cleaned up later. I also don't want us to rely on anyone's memory to go back and remove temporary code, commented out code etc. Perhaps the way to proceed here is to open an issue that identifies the cleanup items, and get this PR merged. Whether it should be merged to main or to a feature branch is another topic. Thoughts on this?

I am okay with opening an issue. We could just leave those comments unresolved and open an issue linking to this PR to resolve the unresolved comments before the final PR is accepted.

apps/vapi/RenderManager.cpp
StasJ: It is initialized to `nullptr` in the class declaration.

clyne:
Opening a separate issue to resolve unresolved comments in this PR seems both efficient, and the least likely to missing something. Let's try that.

StasJ:
> Opening a separate issue to resolve unresolved comments in this PR seems both efficient, and the least likely to missing something. Let's try that.

Done: #2869

clyne:
Great. I will approve as soon as the ubuntu test failures are resolved.

PR #2878, opened 2021-10-26
fix #2871 by introducing additional directories

### LINKED ISSUE #2871: Separate binary executables to different folders
As of 10/5/21, there are 30 binary executables generated in the `bin` directory with a mix of useful binaries, debug binaries, and automatic testing binaries, manual testing binaries, and irrelevant binaries. They can be put in separate directories so the functionality of each directory is clearer.

shaomeng:
This PR fixes issue #2871 by moving 3 binaries to a directory named `debug_binaries`, and 7 binaries to a directory named `test_binaries`.


sgpearse (approved):
With this PR, our build directory will have 4 places we put binaries (bin, test_apps, debug_binaries, and test_binaries).  A little confusing.  I think we should migrate the binaries in build/test_apps into either test_binaries or debug_binaries.  But that can be done outside of this PR.  

Opening #2882.

shaomeng:
Hey @sgpearse just to clarify, there are 3 places to put binaries. `test_app` does not have binaries.

StasJ (approved):

PR #2883, opened 2021-10-26
fix Issue 2865

### LINKED ISSUE #2865: Memory Error---testGrid
Utility program `testGrid` has memory error as reported by valgrind:

```
==150751== Mismatched free() / delete / delete []
==150751==    at 0x483CFBF: operator delete(void*) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==150751==    by 0x1259F9: DeleteHeap() (gridTools.cpp:40)
==150751==    by 0x122938: main (testGrid.cpp:276)
```

shaomeng:
fix #2865 by eliminating a memory bug.


shaomeng:
Hey @sgpearse , I ran clang-format on this single changed file, but probably because I have a different version of clang-format than CircleCI uses, there are disagreement on the placement of `&` and `*`. 

My suggest would be that you change the clang-format specification file to explicitly specify the placement of `&` and `*`, so we'll be consistent with these 2 symbols, regardless different clang-format releases. Does this sound good?

test_apps/smokeTests/gridTools.cpp
```diff
@@ -54,9 +54,10 @@ template<typename T> vector<T *> AllocateBlocksType(const vector<size_t> &bs, co
         nblocks *= nb;
     }
 
-    T *buf = new T[nblocks * block_size];
+    void *tmp = std::malloc(sizeof(T) * nblocks * block_size);
```
clyne, line 57: I'm curious. Can you explain how swapping malloc for new fixes the problem?

shaomeng: The old problem is a mismatch between `new T[]` and `delete`. The fix uses malloc and free, which are a perfect pair.

clyne (approved):

shaomeng:
@sgpearse clang-format issue is resolved; this PR is ready for review.

shaomeng:
> Hey @sgpearse , I ran clang-format on this single changed file, but probably because I have a different version of clang-format than CircleCI uses, there are disagreement on the placement of `&` and `*`.
> 
> My suggest would be that you change the clang-format specification file to explicitly specify the placement of `&` and `*`, so we'll be consistent with these 2 symbols, regardless different clang-format releases. Does this sound good?

The clang-format control of where to place `&` and `*` is named [`PointerAlignment`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html), and it turns out that our clang-format config file has it specified already:

https://github.com/NCAR/VAPOR/blob/d1063b1aabc897ef12b8f36bb161e34d416e2ba4/.clang-format#L60

Weirdly, with the same config, clang-format 13 behaves differently than version 12 and 11. So I have no solution to provide a consistent `&` and `*` placement now.


Note, another related configuration is `SpaceAroundPointerQualifiers` but it doesn't seem to affect what we observe here.

clyne:
> > Hey @sgpearse , I ran clang-format on this single changed file, but probably because I have a different version of clang-format than CircleCI uses, there are disagreement on the placement of `&` and `*`.
> > My suggest would be that you change the clang-format specification file to explicitly specify the placement of `&` and `*`, so we'll be consistent with these 2 symbols, regardless different clang-format releases. Does this sound good?
> 
> The clang-format control of where to place `&` and `*` is named [`PointerAlignment`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html), and it turns out that our clang-format config file has it specified already:
> 
> https://github.com/NCAR/VAPOR/blob/d1063b1aabc897ef12b8f36bb161e34d416e2ba4/.clang-format#L60
> 
> Weirdly, with the same config, clang-format 13 behaves differently than version 12 and 11. So I have no solution to provide a consistent `&` and `*` placement now.
> 
> Note, another related configuration is `SpaceAroundPointerQualifiers` but it doesn't seem to affect what we observe here.

That's discouraging news. If we can't get consistent behavior across different versions of clang-format, than I'm inclined to agree with Scott that it may cause more problems than it solves... We might want to discuss this topic further outside of this PR

shaomeng:
>
> That's discouraging news. If we can't get consistent behavior across different versions of clang-format, than I'm inclined to agree with Scott that it may cause more problems than it solves... We might want to discuss this topic further outside of this PR

Well, it seems to me clang-format version 13 has fixed a bug that present in its earlier versions, because the behavior of version 13 is what clang-format documentation describes, and version 12/11 behaves differently from the documentation.

sgpearse (approved):

PR #2898, opened 2021-11-10
Add a timestep slider to the animation panel

### LINKED ISSUE #2897: Restore time animation slider
The animation tab was refactored and the time slider was replaced with a text box + arrows, which do not allow easy browsing of time steps. Per @leighorf:

I'm using 3.5. Under Scene->Animation there is no time slider. There used to be one that made it easy to jump around the datasets in time. Is this still available? If not, consider it a major feature request. As it stands now I have to type in the time step into a box.

StasJ:
Fixes #2897


clyne (approved):
Awesome. Thanks!

sgpearse (approved):

PR #2901, opened 2021-11-12
2637 - Rotating Slice

### LINKED ISSUE #2637: Slice Renderer: Rotate slice
In Vapor 2.X we could rotate slices along an axis.

In 3.X they can only be aligned to the XY, XZ, and YZ planes.

Users are currently requesting us to implement our previous functionality.

Duplicates #1128, requested by @leighorf

sgpearse:
Adds ability to rotate slice about X, Y, and Z axes.  Also allows specifying an origin of rotation.

Fixes #2637 

~~Note: CentOS build is currently broken.~~


apps/vaporgui/VizWin.cpp
```diff
@@ -858,3 +860,58 @@ void VizWin::updateManip(bool initialize)
 
     GL_ERR_BREAK();
 }
+
+void VizWin::_updateSliceOriginGlyph()
+{
+    VAPoR::SliceParams *sp = dynamic_cast<VAPoR::SliceParams *>(_getRenderParams());
+    if (sp == nullptr) return;
+
+    double xOrigin = sp->GetValueDouble(SliceParams::XOriginTag, 0.);
```
clyne, line 869: Ditto here. Let's generalize this so that _updateSliceOriginGlyph can be used for other things besides the slice renderer

include/vapor/ConvexHull.h
```diff
@@ -0,0 +1,152 @@
+// A C++ program to find convex hull of a set of points. Refer
+// https://www.geeksforgeeks.org/orientation-3-ordered-points/
+// for explanation of orientation()
+#include <iostream>
+#include <stack>
+#include <stdlib.h>
+using namespace std;
+
+/*struct Point
```
clyne, line 9: Remove commented out code

include/vapor/ConvexHull.h
```diff
@@ -0,0 +1,152 @@
...
+        // points next-to-top, top, and points[i] makes
+        // a non-left turn
+        while (S.size() > 1 && orientation(nextToTop(S), S.top(), points[i]) != 2) S.pop();
+        S.push(points[i]);
+    }
+
+    return S;
+
+    // Now stack has the output points, print contents of stack
+    /*while (!S.empty())
```
clyne, line 132: clean dead code (or ifdef appropriately if it might be needed later)

include/vapor/SliceParams.h
```diff
@@ -40,18 +39,19 @@ class PARAMS_API SliceParams : public RenderParams {
...
-
+    int                 GetSampleRate() const;
+    void                SetSampleRate(int rate);
+    int                 GetDefaultSampleRate() const;
+    void                SetCachedValues(std::vector<double> values);
     std::vector<double> GetCachedValues() const;
 
     static const string _sampleRateTag;
-    static const string SampleLocationTag;
+    static const string XRotationTag;
```
clyne, line 49: Simply moving these to RenderParams would make them usable by other renderers (e.g. contour renderer)

clyne, line 49: Also, for consistency, why three component flags and not simply a vector RotationTag and an OrientationTag? This would be consistent with other vector quantities (e.g. RGB, domain extents, etc.)

include/vapor/SliceRenderer.h
```diff
@@ -36,52 +37,73 @@ class RENDER_API SliceRenderer : public Renderer {
         int                 compressionLevel;
         int                 textureSampleRate;
         int                 orientation;
+        double              xRotation;
```
clyne, line 40: A miner nit, but other vector arguments are passed in vector form why not be consistent? Instead of XRotation, yRotation, zRotation, why not:

`std::vector<double> rotation`

lib/render/SliceRenderer.cpp
```diff
@@ -178,144 +188,231 @@ int SliceRenderer::_resetBoxCache()
...
-            missingValue = grid->GetMissingValue();
-            if (varValue == missingValue)
-                dataValues[index + 1] = 1.f;
-            else
-                dataValues[index + 1] = 0.f;
-
-            dataValues[index] = varValue;
+    // Define a rectangle that encloses our polygon in 3D space.  We will sample along this
+    // rectangle to generate our 2D texture.
+    std::vector<glm::vec3> rectangle3D = {glm::vec3(), glm::vec3(), glm::vec3(), glm::vec3()};
```
clyne, line 353: Is there a reason why inverseProjection is a lamba and not a function defined at the file scope? You redefine the same lambda later in the code, which can lead to a maintenance problem. What if one gets changed and the other doesn't?

lib/render/SliceRenderer.cpp
```diff
@@ -178,144 +188,231 @@ int SliceRenderer::_resetBoxCache()
...
+    glm::vec2 offset = {delta.x / 2., delta.y / 2.};
 
-        for (int i = 0; i < _textureSideSize; i++) {
-            varValue = grid->GetValue(coords);
+    int index = 0;
+    for (int j = 0; j < _ySamples; j++) {
+        for (int i = 0; i < _xSamples; i++) {
+            glm::vec3           samplePoint = inverseProjection(offset.x + _rectangle2D[0].x + i * delta.x, _rectangle2D[0].y + offset.y + j * delta.y);
+            std::vector<double> p = {samplePoint.x, samplePoint.y, samplePoint.z};
+            varValue = grid->GetValue(p);
```
clyne, line 404: Grid::GetValue(const std::vector<double> &coords) is being deprecated. Better to use: Grid::GetValue(const CoordType &coords)

I.e. change `std::vector<double> p` to `CoordType p`

clyne (commented):
@sgpearse looks really good. I just had a few inline nits. 

The GUI is ... a little challenging to use. But it is what we discussed. Let's give Stas a chance to get the contour renderer refactored and maybe we can have a discussion on the GUI.

apps/vaporgui/VizWin.cpp
sgpearse: My thought was that the Contour and Slice renderers would share a base class that would both have an XOriginTag, and could both render the glyph using the same logic in VizWin.  I couldn't think of any use cases outside the Slice and Contour renderers.

The "orientable" flag would fix this, but 1) it introduces inconsistency in how the VizWin operates (see comment above) and 2) each RenderParams class would need GetOrientable() functions and initialization logic.  Are there advantages to using the flag over inheriting from a Slice/Contour base class?

I'm open all week if this would be easier to talk about, or just give me the signal and I can implement the flag.

apps/vaporgui/VizWin.cpp
```diff
@@ -858,3 +860,58 @@ void VizWin::updateManip(bool initialize)
 
     GL_ERR_BREAK();
 }
+
+void VizWin::_updateSliceOriginGlyph()
+{
+    VAPoR::SliceParams *sp = dynamic_cast<VAPoR::SliceParams *>(_getRenderParams());
```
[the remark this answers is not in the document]
sgpearse: Hmm.  I have a few questions regarding this file.  Just want to be clear on how to proceed.

1) Querying class types is the same approach we use elsewhere in the code.  For example, we do it when deciding whether to draw a Box/Manip for the Flow seeding tab.  One example is at [line 808](https://github.com/NCAR/VAPOR/blob/15e4643d8f4680af4bb722c81d0d6a839954798a/apps/vaporgui/VizWin.cpp#L808).

`if (gp->ActiveTab() == FlowEventRouter::SeedingTabName) _manipFlowSeedFlag = true;`

Do we want to introduce inconsistency here?  We've talked about refactoring the VizWin and this seems to be adding to that case.

2) An "orientable renderparams flag" does not sound like a parater "tag" that can be changed by the user.  Slice and contour renderers will always be orientable, and this will never change.  Just want to be clear that we're proposing private boolean flag with a getter, and not a tag.

apps/vaporgui/PSliceOriginSelector.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "PSection.h"
+
+class PDoubleSliderEdit;
+
```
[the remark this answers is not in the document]
sgpearse: No problem.  Fixed.

include/vapor/ConvexHull.h
sgpearse: Fixed with BUILD_STANDALONE macro.

include/vapor/ConvexHull.h
sgpearse: BUILD_STANDALONE

include/vapor/SliceParams.h
sgpearse: Single values (vs vectors) allow use of the low level interface.  No validation needs to be done for rotation on the X, Y, or Z axis.  So all we need is a single parameter for each rotation, and an LLI widget like what we have at [line 29 of SliceEventRouter.cpp](https://github.com/NCAR/VAPOR/blob/9fd7e90a3794383d826107c9a4160c3b1ddea680/apps/vaporgui/SliceEventRouter.cpp#L30).

This follows [the policy we wrote on PWidgets](https://docs.google.com/document/d/1LlTr1rHSRcCucN6U0KzW_HZUq_Q2cyPsQLD3nCJyf_k/edit#) a while back.

include/vapor/SliceRenderer.h
sgpearse: Copied from above:

> Single values (vs vectors) allow use of the low level interface. No validation needs to be done for rotation on the X, Y, or Z axis. So all we need is a single parameter for each rotation, and an LLI widget like what we have at line 29 of SliceEventRouter.cpp.
> 
> This follows the policy we wrote on PWidgets a while back.

lib/render/SliceRenderer.cpp
sgpearse: That's a good point.  Fixed.

lib/render/SliceRenderer.cpp
```diff
@@ -178,144 +188,231 @@ int SliceRenderer::_resetBoxCache()
...
+}
 
-            index += 2;
-            coords[X] += deltas[X];
-        }
-        coords[Z] += deltas[Z];
-    }
+// Huges-Moller algorithm to get an orthogonal vector
+// https://blog.selfshadow.com/2011/10/17/perp-vectors/
+glm::vec3 SliceRenderer::_getOrthogonal(const glm::vec3 u) const
```
StasJ, line 370: While this may be correct, it requires squinting or a proof to confirm. A more self-evident algorithm is to take the cross product with any other vector in the same space

lib/render/SliceRenderer.cpp
```diff
@@ -435,34 +524,71 @@ void SliceRenderer::_getModifiedExtents(vector<double> &min, vector<double> &max
...
+            lgl->Vertex3f(vert1.x, vert1.y, vert1.z);
+            lgl->Vertex3f(vert2.x, vert2.y, vert2.z);
+        }
+        glm::vec3 vert1 = _rectangle3D[3];
+        glm::vec3 vert2 = _rectangle3D[0];
+        lgl->Vertex3f(vert1.x, vert1.y, vert1.z);
+        lgl->Vertex3f(vert2.x, vert2.y, vert2.z);
+    }
+    lgl->End();
+#endif
```
StasJ, line 575: Could this be moved to a function for readability?

lib/render/SliceRenderer.cpp
```diff
@@ -435,34 +524,71 @@ void SliceRenderer::_getModifiedExtents(vector<double> &min, vector<double> &max
...
+    glEnable(GL_DEPTH_TEST);
+    glDepthMask(GL_FALSE);
+
     _initializeState();
 
-    if (_isDataCacheDirty()) {
-        rc = _resetDataCache();
+    // If we're in fast mode, degrade the quality of the slice for better interactivity
+    if (fast) {
+        _textureSideSize = 50;
```
StasJ, line 584: For very large data I worry this may be too slow. What about just showing your debug outline in fast mode?

StasJ (commented):
As it stands, extending the arbitrary slice capability to the contour renderer would require a large amount of copied code. Since that is one of the goals of this effort, could that capability be separated into something that can be reused by the contour renderer?

lib/render/SliceRenderer.cpp
sgpearse: Got it.  Fixed.

lib/render/SliceRenderer.cpp
sgpearse: From what I've read, there are several approaches to this problem.  Huges-Moller seemed simplest.  If we take a cross product of our normal with any other vector, we could end up with the wrong result.  For example, the normal x normal would be 0. 

Note: I am using your suggested approach for defining _axis2 of the basis function.  From my research, the first axis requires something more along these lines.

lib/render/SliceRenderer.cpp
sgpearse: Sure, no problem.

lib/render/SliceRenderer.cpp
sgpearse: That's something I've been thinking about too.  I was told to get a PR in though, and worry about performance later.  I think this is worth revisiting if it's too slow.

include/vapor/ConvexHull.h
```diff
@@ -0,0 +1,152 @@
...
+{
+    glm::vec2 p = S.top();
+    S.pop();
+    glm::vec2 res = S.top();
+    S.push(p);
+    return res;
+}
+
+// A utility function to swap two points
+void swap(glm::vec2 &p1, glm::vec2 &p2)
```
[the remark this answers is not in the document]
sgpearse: Because I've borrowed this code and I'm averse to changing it.

include/vapor/ConvexHull.h
```diff
@@ -0,0 +1,152 @@
...
+using namespace std;
+
+/*struct Point
+{
+    double x, y;
+};*/
+
+// A global point needed for  sorting points with reference
+// to  the first point Used in compare function of qsort()
+glm::vec2 p0;
```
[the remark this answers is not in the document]
sgpearse: Good point.  I was hoping to be as hands-off as possible with this code since it's not mine, but it looks like it should be broken into a source file with namespace.

sgpearse:
@StasJ 

> As it stands, extending the arbitrary slice capability to the contour renderer would require a large amount of copied code. Since that is one of the goals of this effort, could that capability be separated into something that can be reused by the contour renderer?

This was one of my concerns last week but we didn't get to chat about it.  I was wondering if the only difference between the Slice and Contour renderers would lie in the shader.  I originally thought that the SliceRenderer would need to be broken out into reusable components for the Contours, but could binding a Contour shader be the only thing we need to do?

IE - There's a function named _configureShader that binds the Slice shader.  Could a contour renderer just point to a contour shader, with most other things being the same?

include/vapor/SliceRenderer.h
```diff
@@ -36,52 +37,73 @@ class RENDER_API SliceRenderer : public Renderer {
...
+        double              yOrigin;
+        double              zOrigin;
         std::vector<float>  tf_lut;
         std::vector<double> tf_minMax;
         std::vector<double> boxMin, boxMax;
         std::vector<double> domainMin, domainMax;
         std::vector<double> sampleLocation;
     } _cacheParams;
 
+    struct _vertex {
```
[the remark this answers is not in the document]
sgpearse: I admit that the use of _vertex is a little obscure.  It serves to map 3D vertices to 2D vertices within the slice plane.  Once we find the connectivity of the 2D vertices, we can know the 3D connectivity too.

Since _vertex is using both elements of the struct, I don't think a union makes sense.  Am I misunderstanding something?

include/vapor/ConvexHull.h
```diff
@@ -0,0 +1,152 @@
...
+{
+    //glm::vec2 points[] = {{0, 3}, {1, 1}, {2, 2}, {4, 4},
+    //                  {0, 0}, {1, 2}, {3, 1}, {3, 3}};
+    glm::vec2 points[] = {
+                      {0.397329, -.5},      {-0.397329, .5},
+                      {-0.48774, -.5},     {0.48774,   .5},
+                      {0.625,    0.288675}, {-.625,     -0.288675}};
+    int n = sizeof(points)/sizeof(points[0]);
+    convexHull(points, n);
+    return 0;
```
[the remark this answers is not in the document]
sgpearse: This is one of the things I've been pondering.  Since I'm using somebody else's implementation, I've been inclined to change as little as possible.

Let me break this out into header/source and give it a namespace.  I'll get back to you when ready.

apps/vaporgui/VizWin.cpp
```diff
@@ -858,3 +860,58 @@ void VizWin::updateManip(bool initialize)
...
+    int            lod = sp->GetCompressionLevel();
+    string         varName = sp->GetVariableName();
+    vector<string> fieldVars = sp->GetFieldVariableNames();
+
+    ParamsMgr *      paramsMgr = _controlExec->GetParamsMgr();
+    AnimationParams *aParams = (AnimationParams *)paramsMgr->GetParams(AnimationParams::GetClassType());
+    int              timeStep = aParams->GetCurrentTimestep();
+
+    DataStatus *dataStatus = _controlExec->GetDataStatus();
+    string      dataMgrName = _getCurrentDataMgrName();
```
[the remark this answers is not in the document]
sgpearse: I'm a little bit unclear.  VizWin has a convenience function _getCurrentDataMgrName().  Is this bad?

I'm trying to find an alternative call from the Renderer class but I'm not seeing it when I grep in lib/render or include/vapor.  Is there a superior alternative?

apps/vaporgui/VizWin.cpp
```diff
@@ -858,3 +860,58 @@ void VizWin::updateManip(bool initialize)
...
+    scales[2] *= scales2[2];
+
+    int            refLevel = sp->GetRefinementLevel();
+    int            lod = sp->GetCompressionLevel();
+    string         varName = sp->GetVariableName();
+    vector<string> fieldVars = sp->GetFieldVariableNames();
+
+    ParamsMgr *      paramsMgr = _controlExec->GetParamsMgr();
+    AnimationParams *aParams = (AnimationParams *)paramsMgr->GetParams(AnimationParams::GetClassType());
+    int              timeStep = aParams->GetCurrentTimestep();
```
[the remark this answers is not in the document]
sgpearse: Is it better to use SettingsParams?  AnimationParams is the only class used throughout this file to acquire the current timestep.

apps/vaporgui/VizWin.cpp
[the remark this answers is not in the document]
clyne: > Hmm. I have a few questions regarding this file. Just want to be clear on how to proceed.
> 
> 1. Querying class types is the same approach we use elsewhere in the code.  For example, we do it when deciding whether to draw a Box/Manip for the Flow seeding tab.  One example is at [line 808](https://github.com/NCAR/VAPOR/blob/15e4643d8f4680af4bb722c81d0d6a839954798a/apps/vaporgui/VizWin.cpp#L808).
> 
> `if (gp->ActiveTab() == FlowEventRouter::SeedingTabName) _manipFlowSeedFlag = true;`
> 
> Do we want to introduce inconsistency here? We've talked about refactoring the VizWin and this seems to be adding to that case.
> 

That's a hack and shouldn't have passed review IMHO :-)

> 1. An "orientable renderparams flag" does not sound like a parater "tag" that can be changed by the user.  Slice and contour renderers will always be orientable, and this will never change.  Just want to be clear that we're proposing private boolean flag with a getter, and not a tag.

Not all parameters are settable - they are for information only and do not have setters

include/vapor/ConvexHull.h
clyne: Did you forget to push the commit?

apps/vaporgui/VizWin.cpp
clyne: Hard coding is rarely a good practice, which is what testing the class type amounts to. The contour and slice renderer do not share a base class, so this isn't an option without refactoring. Refactoring simply because contour and and slice both support arbitrary cutting planes would paint us into a corner. What happens is another renderer comes along that requires an orientable plane, but also shares some property with the Barb renderer. Do we then need multiple inheritance?  

Something like a RenderParams::GetOrientable() that defaults to false (so other renderers can ignore it) seems like the way to go here.

apps/vaporgui/VizWin.cpp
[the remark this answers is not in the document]
clyne: Hard coding is rarely a good practice, which is what testing the class type amounts to. The contour and slice renderer do not share a base class, so this isn't an option without refactoring. Refactoring simply because contour and and slice both support arbitrary cutting planes would paint us into a corner. What happens is another renderer comes along that requires an orientable plane, but also shares some property with the Barb renderer. Do we then need multiple inheritance?  

Something like a RenderParams::GetOrientable() that defaults to false (so other renderers can ignore it) seems like the way to go here.

clyne:
> As it stands, extending the arbitrary slice capability to the contour renderer would require a large amount of copied code. Since that is one of the goals of this effort, could that capability be separated into something that can be reused by the contour renderer?

@StasJ, agreed. We need to extract the common code. I suggested to @sgpearse that you decide what can/should be reused to 1) keep the PR smaller, and 2) avoid unnecessary work.

include/vapor/SliceParams.h
sgpearse: I agree that the Contour and Slice renderers should have these tags.  But putting these tags into RenderParams makes it monolithic.  Most renderers have no dependency on rotation, so why should they keep in their state?

IE - The RenderParams base class should only have tags that are used among all of its subclasses.  Yet it has tags like _xFieldVariableNameTag. 
 Most renderers do not use field variables.  Unless I'm missing something, I believe this should be avoided.

clyne: It's pretty common in OOP to have derived classes that don't use all of their superclasses capabilities. If these flags only applied to one renderer it would make sense to put them there. But since we've identified at least two in seems reasonable to make them common to the base class and have a getter that indicates if they are applicable (as we've done with field variables, height variables, etc.). The alternatives are 1) replication. Not good. 2) refactoring and introducing a new intermediate derived class every time a situation like this arises. Not practical

apps/vaporgui/VizWin.cpp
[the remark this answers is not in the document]
sgpearse: Got it, makes sense.  Fixed.

Note that the tags XOrigin, YOrigin, and ZOrigin have been moved from SliceParams to RenderParams.  I anticipate we will need to do the same thing with the XRotation, YRotation, and ZRotation tags but I'm going to hold off on this for now until I have a better understanding on how the ContourParams will inherit them.

include/vapor/ConvexHull.h
sgpearse: Weird.  It should be there now.

include/vapor/SliceParams.h
sgpearse: Interesting.  Fixed.  Note that the SampleRateTag has also been added to RenderParams for the same reasons.

This kind of problem seems like a balancing act.  Borrowing from Stas' vehicle metaphor: 

Say we have a class Vehicle, with subclasses Car, Boat, and Airplane.  Boat and Airplane need a Rudder member.  The approach we're taking right now is to put Rudder in the Vehicle base class, even though Car doesn't use it.  This is fine if members like Rudder are reasonably limited, like in our case with Rotation and Origin tags.  However new intermediate classes eventually become necessary if Airplane needed members for ailerons, flaps, and the numerous components that don't exist on Cars and Boats.

I don't think we're at the point of adding things like "Ailerons and Flaps" to RenderParams yet.

include/vapor/RenderParams.h
```diff
@@ -433,6 +436,13 @@ class PARAMS_API RenderParams : public ParamsBase {
     static const string _yFieldVariableNameTag;
     static const string _zFieldVariableNameTag;
     static const string _constantOpacityTag;
+    static const string XOriginTag;
```
clyne, line 439: There is no documentation for these public tags. Prior to making many of the tags (e.g. _constantOpacityTag) public they were only accessible via public setters and getters that had doxygen documentation. The new flags introduced either need to have setters and getters that are documented, or the member variables themselves should be documented with Doxygen.

clyne (commented):
One or two more nits. Getting close!

StasJ:
> > As it stands, extending the arbitrary slice capability to the contour renderer would require a large amount of copied code. Since that is one of the goals of this effort, could that capability be separated into something that can be reused by the contour renderer?
> 
> @StasJ, agreed. We need to extract the common code. I suggested to @sgpearse that you decide what can/should be reused to 1) keep the PR smaller, and 2) avoid unnecessary work.

One potential solution for making this reusable would be to package the slicing capability into a utility function in the VDC library. This function would take a 3D grid and the slice parameters and it would return a 2D grid representing the slice.

lib/render/SliceRenderer.cpp
StasJ: Thats fine with me, my way also has its drawbacks. Just a note though: normal x normal is not any other vector, it is the same vector, which is why it doesn't work.

include/vapor/SliceRenderer.h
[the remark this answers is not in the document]
StasJ: Sorry, my comment indeed does not make sense after your explanation but I was confused by this code: You have a struct called `_vertex` that contains two vertices, one 2D and one 3D, without any further explanation. It is not apparent to the reader what this is trying to accomplish. Since you are mapping from 3D coords to 2D coords you could use `map<vec3, vec2>` or `vector<pair<vec3, vec2>>` but maybe just a more descriptive name would suffice?

apps/vaporgui/VizWin.cpp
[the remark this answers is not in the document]
StasJ: In this function, `sp` is the Slice RenderParams which has a function to get the current timestep so getting the animation params is not necessary.

apps/vaporgui/VizWin.cpp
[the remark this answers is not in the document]
StasJ: I looked into it and it does not appear there is an easy way to get the data mgr name from the renderer params so I'd say this is outside of the scope of this PR.

include/vapor/RenderParams.h
sgpearse: Got it.  Fixed.  This info belongs in our [coding conventions](https://github.com/NCAR/VAPOR/wiki/Vapor-Coding-Convention) so I added a Doxygen section that covers this.

lib/render/SliceRenderer.cpp
sgpearse: @stasj I'm not sure why, but this comment is not getting an "Outdated" label for the code it's referring to.  You can see the change [here](https://github.com/NCAR/VAPOR/blob/a8c58b18bbb21e5036531de1c8cabc80d6d44574/lib/render/SliceRenderer.cpp#L535).

include/vapor/SliceRenderer.h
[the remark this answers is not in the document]
sgpearse: Ok, I changed the name.  Also added a comment in the header since I think there's more to be described than can fit in a variable name.

include/vapor/ConvexHull.h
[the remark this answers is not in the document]
sgpearse: @stasj ConvexHull has now been broken into a source and header file.

sgpearse:
@stasj There was a bug regarding the downsampling I fixed yesterday and I suspect it was what you are seeing with Leigh's data.  I tried reproducing your screenshots with the current code but wasn't able to.  Could you try again?

I do see the UGRID problem.  This PR is getting pretty long so I'd like to file a separate issue.  One thing I'm noticing is that the UGRID dataset has a Z stretch of ~4 million.

lib/params/RenderParams.cpp
```diff
@@ -51,6 +51,13 @@ const string RenderParams::_RefinementLevelTag = "RefinementLevel";
 const string RenderParams::_transferFunctionsTag = "MapperFunctions";
 const string RenderParams::_stretchFactorsTag = "StretchFactors";
 const string RenderParams::_currentTimestepTag = "CurrentTimestep";
+const string RenderParams::XOriginTag = "XOrigin";
```
clyne, line 54: Sorry to nit pick on this, but "origin" and "rotation"  are perhaps too general for something that is intended to describe position and orientation of a specific thing, a cutting or slice plane. Lots of things could have an origin or a rotation. How about something like

XSlicePlaneOrigin, XSlicePlaneRotation, etc?

sgpearse: No problem.  Origin and Rotation parameter tags are updated.

IE:

XOriginTag = XSlicePlaneOriginTag
XRotationTag = XSlicePlaneRotationTag

clyne (approved):
Awesome. Great work, @sgpearse

StasJ (approved):

Issue #2906, opened 2021-11-17
Slice + UGRID problem


sgpearse:
Import the UGRID data in /glade/p/cisl/vast/vapor/data/Source/UGRID/NOAA-geoflow/small

Create a slice with the following rotation parameters:
X = -19.8
Y = 0
Z = 16.2

The slice will not be sampled properly and will stretch outside of the domain.  Note that the Z stretch factor for this dataset is ~4 million.

![](https://user-images.githubusercontent.com/2772687/142095913-c068b690-278d-4a76-8eb6-c1923392afe7.png)

PR #2932, opened 2021-12-09
Slicer - Creates 2D grids of slice data

sgpearse:
This PR allows Renderers to construct a Slicer class that will deliver 2D grids based on RenderParams instances.  The spatial extents of a Grid class are insufficient to define a Slice's position and winding order, so Slicer also provides accessors to this information.

See SliceRenderer::_regenerateSlice() for an example of how renderers can sample data with it.  

SliceRenderer::_initVertexVBO() shows how to access an ordered list of six vertices that will correctly orient the two triangles of the texture.

SliceRenderer::_drawDebugPolygons() shows how to get debugging info from the Slicer.

ToDo: Run valgrind

### DIFF
--- a/apps/cf2vdc/cf2vdc.cpp
+++ b/apps/cf2vdc/cf2vdc.cpp
@@ -66,8 +66,8 @@ size_t gcd(size_t n1, size_t n2)
 
 size_t lcm(size_t n1, size_t n2) { return ((n1 * n2) / gcd(n1, n2)); }
 
-int copyVarHelper(DC &dc, VDC &vdc, int fdr, int fdw, vector<size_t> &buffer_dims, vector<size_t> &src_hslice_dims, vector<size_t> &dst_hslice_dims, size_t src_nslice, size_t dst_nslice, double mv,
-                  float *buffer)
+int copyVarHelper(DC &dc, VDCNetCDF &vdc, int fdr, int fdw, vector<size_t> &buffer_dims, vector<size_t> &src_hslice_dims, vector<size_t> &dst_hslice_dims, size_t src_nslice, size_t dst_nslice,
+                  double mv, float *buffer)
 {
     VAssert(buffer_dims.size() == src_hslice_dims.size());
     VAssert(buffer_dims.size() == dst_hslice_dims.size());
@@ -116,7 +116,7 @@ int copyVarHelper(DC &dc, VDC &vdc, int fdr, int fdw, vector<size_t> &buffer_dim
     return (0);
 }
 
-int CopyVar2d3dMask(DC &dc, VDC &vdc, size_t ts, string varname, int lod)
+int CopyVar2d3dMask(DC &dc, VDCNetCDF &vdc, size_t ts, string varname, int lod)
 {
     // Only data variables can have masks
     //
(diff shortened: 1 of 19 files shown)


clyne (commented):
Scott, for the most part this looks good, but before getting into nitty gritty details I have two comments around
the design choice: 1. should "Slicer" be a class, and 2. is there a reason
it should be so tightly tied to the DataMgr and RenderParms classes? To me the functionality required looks more like a function than a class object. For one,
Slicer only does one thing: GetSlice(). Consider simply having a function
with a definition something like:

`RegularGrid *SliceGridAlongPlane(Grid *grid3d, Foo planeDescription, size_t w, size_t h, float *data)`

where grid3d is the 3d variable to be sampled, planeDescription describes
the plane (e.g. origin, normal vector), w and h give the width and
height of the sampled output plane, and data points to memory of
size w*h)

Not including DataMgr and RenderParams as arguments to this function
may mean that a few extra lines of code are required to use it, but
it makes the function far more general (it could be used, for
example, by the Python engine). It also does a better job of
information hiding; Slicer doesn't really need full access to the
DataMgr and RenderParams to do its job.
~

sgpearse:
Thanks for the review.  I'll get on this after resolving #2906.  

The answers to your questions and comments are below.

> 1. should "Slicer" be a class?

The reason it's a class is because it holds the state for the following fields that need to be queried at different times during rendering:

1) The array of data allocated for the new RegularGrid
2) Winding-order of the vertices of the returned slice
3) Polygon edges for debugging

If the memory needs to be allocated by the renderer, that's fine.  My goal was to have an object create a Grid and maintain it, but it wasn't clear where we wanted memory management to happen.

Follow up: why do we want the renderer to manage the slice's memory?  This seems to mean  that all Renderers and PythonEngine interfaces will need to individually manage the allocated memory, rather than having it handled by the Slicer's destructor. 

> For one,
> Slicer only does one thing: GetSlice(). Consider simply having a function
> with a definition something like...

This follows the first rule of SOLID - Single responsibility principle.  I believe this is a good thing in general.  Am I missing something?

clyne:
> Thanks for the review. I'll get on this after resolving #2906.
> 
> The answers to your questions and comments are below.
> 
> > 1. should "Slicer" be a class?
> 
> The reason it's a class is because it holds the state for the following fields that need to be queried at different times during rendering:
> 
> 1. The array of data allocated for the new RegularGrid
> 2. Winding-order of the vertices of the returned slice
> 3. Polygon edges for debugging

These variables are all calculated by GetSlice and could easily be returned by GetSlice(). Prefer local or parameters over member variables, and member variables over global variables. This improves readability of the code, and helps avoid side effects. E.g. if I call member function `foo` how do I know if it modifies/accesses member variable `bar`

> 
> If the memory needs to be allocated by the renderer, that's fine. My goal was to have an object create a Grid and maintain it, but it wasn't clear where we wanted memory management to happen.
> 
> Follow up: why do we want the renderer to manage the slice's memory? This seems to mean that all Renderers and PythonEngine interfaces will need to individually manage the allocated memory, rather than having it handled by the Slicer's destructor.

For one thing, if the user of Slicer deletes a Slicer instance then the Grid returned by GetSlicer becomes invalid. 

> 
> > For one,
> > Slicer only does one thing: GetSlice(). Consider simply having a function
> > with a definition something like...
> 
> This follows the first rule of SOLID - Single responsibility principle. I believe this is a good thing in general. Am I missing something?

clyne:
> 

We have a DataMgrUtils module that provides functions that operate on the DataMgr. It might make sense to have "GridUtils" class for functions that operate on the Grid class.

Issue #2922, opened 2021-12-02
DockerHub is changing its terms of service


sgpearse:
DockerHub is currently hosting images for our Centos and Ubuntu builds.  These images have already been configured to build Vapor installers, so all CircleCI needs to do is instantiate an image, do a git pull, then make an installer.  No cloning, downloading of third party libraries, or installing dependencies is needed with this approach and it saves on CircleCI's build time.

The [terms of service are changing](https://www.docker.com/blog/updating-product-subscriptions/?utm_campaign=2021-12-02-discount-db15&utm_medium=email&utm_source=marketo&mkt_tok=NzkwLVNTQi0zNzUAAAGBG-cZ42_FbSr922S_XoFaFz9S6vOiJcoHTwyW84K3rBjfb3FILklL_I78bGRUcpPqF5PHc75kLbBpGCHm7vyFJhOahSoqr98xdz4JaAr_4w) on January 31, 2022, and NCAR may not have an account for this previously free approach.

We should be ready to do away with our docker images in January unless NCAR creates a policy on this.  All this means is that our CentOS and Ubuntu builds on CircleCI will include cloning Vapor, downloading third party libraries, and installing dependencies.  This will take longer than instantiating a docker image, but it won't require a subscription to DockerHub and the Linux builds on CircleCI are cheap.

sgpearse:
Note: 
DockerHub has now disabled automated builds for non-paying users.  This means that we are forced to build docker images locally (which can be ~10 gigabytes in size) and manually upload them to DockerHub.  I've done this in the past but it's been untenable.

Issue #2728, opened 2021-05-12
Contours: Allow for rendering 3D variables


sgpearse:
Our contour renderer only works with 2D variables.  Users have expressed the need for having them in 3D too: i.e. draw contours on a slice from a 3D volume

clyne:
Related to #2637

clyne:
The plane manipulator that was implemented to support #2637 is somewhat cumbersome to use. I looked at what both VISiT and ParaView are doing and found their approach to be not terribly satisfying either (though ParaView supports mouse interactions that make manipulating the plane much easier). There are two problems with our current implementation. First, while specifying the orientation as rotations about the primary axis does the job, it assumes that there is some default orientation about which these transformations apply. Specifying the orientation via a normal is less ambiguous and, IMHO, more natural for describing the orientation of a plane. This is what both ViSIT and ParaView do. 

The more problematic issue is the "Slice Origin"; it provides more degrees of freedom than are necessary, which makes it confusing to use. For example, if the plane is perpendicular to the Z axis, changing the Y and X coordinates is a no-op. I find it pretty challenging to position a non-axis-aligned plane where I want it. A simpler, more intuitive approach might be to specify a displacement along the plane normal.  The displacement could be from either the origin or the center of the domain. I think this would give the flexility to arbitrarily position the plane in a way that is far more intuitive. 

@StasJ , @sgpearse, thoughts?

StasJ:
> The plane manipulator that was implemented to support #2637 is somewhat cumbersome to use. I looked at what both VISiT and ParaView are doing and found their approach to be not terribly satisfying either (though ParaView supports mouse interactions that make manipulating the plane much easier). There are two problems with our current implementation. First, while specifying the orientation as rotations about the primary axis does the job, it assumes that there is some default orientation about which these transformations apply. Specifying the orientation via a normal is less ambiguous and, IMHO, more natural for describing the orientation of a plane. This is what both ViSIT and ParaView do.
> 
> The more problematic issue is the "Slice Origin"; it provides more degrees of freedom than are necessary, which makes it confusing to use. For example, if the plane is perpendicular to the Z axis, changing the Y and X coordinates is a no-op. I find it pretty challenging to position a non-axis-aligned plane where I want it. A simpler, more intuitive approach might be to specify a displacement along the plane normal. The displacement could be from either the origin or the center of the domain. I think this would give the flexility to arbitrarily position the plane in a way that is far more intuitive.
> 
[...]

PR #2939, opened 2021-12-17
Functional slice sampling implementation, and address UGRID problem #2906

### LINKED ISSUE #2906: Slice + UGRID problem

sgpearse:
Fixes the following issues from the (now closed) PR #2932:

1) SliceRenderer now uses the SliceGridAlongPlane() function, which is located in the VDC library.
2) Addresses the UGRID problem reported in #2906

This is a moderate refactoring of the last implementation and it uses a new set of algorithms for generating convex hulls, and finding the minimal area rectangles of those convex hulls.  These algorithms are implemented in header files located in VAPOR/include/vapor/GTE.

Note: there is a slight problem that still persists with the UGRID dataset.  This is due to the extreme difference in scale between the X/Y axes (which span tens of millions of meters), compared to the Z axis (which spans 1 meter).  This difference in scale introduces floating point error when projecting certain planes between 2D and 3D coordinates.  The data type for spatial coordinates has been changed from float to double in this PR, and this helps the problem.  But there is still slight error in UGRID's extreme case.  See the image below, showing the texture being drawn slightly outside of the domain.
 
If the user changes their Z coordinates to something reasonable, this error goes away.  I've placed a grid file that demonstrates this in /glade/p/cisl/vast/vapor/data/Source/UGRID/NOAA-geoflow/small, called **gridStretchedZ.nc**.  This file has Z coordinates spanning from 0 to 10,000 instead of 1 to 2.  Load the UGRID data with this file instead of grid.nc and error from the 2D/3D projection arithmetic goes away.  

I'm not aware of glm supporting precision higher than double, and I think it's reasonable to require UGRID users to have a more reasonable representation of the Z coordinates.

Screen Shot 2021-12-17 at 2 39 29 PM

Fixes #2906


include/vapor/SliceGridAlongPlane.h
```diff
@@ -0,0 +1,72 @@
+#pragma once
+
+#include <vapor/glutil.h>
+#ifdef Darwin
+    #include <OpenGL/gl.h>
+#else
+    #include <GL/gl.h>
```
StasJ, line 7: I don't know if the VDC library is technically linking against OpenGL.

include/vapor/SliceGridAlongPlane.h
```diff
@@ -0,0 +1,72 @@
...
+    std::vector<double> rotation;
+    std::vector<double> boxMin;
+    std::vector<double> boxMax;
+};
+
+glm::tvec3<double, glm::highp> getOrthogonal(
+    const glm::tvec3<double, glm::highp>& u
+);
+
+void rotate(
```
StasJ, line 25: These probably should not be in the global namespace.

include/vapor/SliceGridAlongPlane.h
```diff
@@ -0,0 +1,72 @@
...
+    const std::vector<glm::tvec2<double, glm::highp>>& rectangle2D,
+    float* dataValues
+); 
+
+void getWindingOrder(
+    std::vector<double>& windingOrder,
+    const std::vector<glm::tvec3<double, glm::highp>>& rectangle3D
+);
+
+VAPoR::RegularGrid* SliceGridAlongPlane(
```
StasJ, line 64: Missing documentation.

StasJ (changes requested):
This PR adds thousands of lines of code from an external library. Shouldn't this be in our third party libs rather than in the Vapor source tree?

include/vapor/SliceRenderer.h
```diff
@@ -59,35 +60,22 @@ class RENDER_API SliceRenderer : public Renderer {
...
     void      _createDataTexture(float *dataValues);
-    int       _saveTextureData();
-    void      _populateData(float *dataValues, Grid *grid) const;
-    glm::vec3 _getOrthogonal(const glm::vec3 u) const;
-    glm::vec3 _rotateVector(glm::vec3 vector, glm::quat rotation) const;
+    int       _regenerateSlice();
+    int       _getGrid3D(Grid*& grid) const;
+    void      _drawDebugPolygons();
 
     double _newWaySeconds;
```
clyne, line 80: Some of these private member variables and functions don't appear to be used. Please delete or ifdef out if needed for diagnostics.

lib/render/SliceRenderer.cpp
```diff
@@ -176,284 +164,60 @@ void SliceRenderer::_resetColormapCache()
...
-        vertex.twoD = {x, y};
-        unorderedTwoDPoints[count].x = vertex.twoD.x;
-        unorderedTwoDPoints[count].y = vertex.twoD.y;
-        count++;
+    Grid* grid3d = nullptr;
+    int rc = _getGrid3D( grid3d );
+    if ( rc < 0 ) return -1;
+
+    // Get data values from a slice
+    float* dataValues = new float[_textureSideSize*_textureSideSize];
```
clyne, line 174: It doesn't look like this memory is ever being freed. is there a reason not to use a smart pointer?

lib/render/SliceRenderer.cpp
```diff
@@ -176,284 +164,60 @@ void SliceRenderer::_resetColormapCache()
...
+    pd.origin = {_cacheParams.xOrigin, _cacheParams.yOrigin, _cacheParams.zOrigin};
+    pd.rotation = {_cacheParams.xRotation, _cacheParams.yRotation, _cacheParams.zRotation};
+    pd.boxMin = _cacheParams.boxMin;
+    pd.boxMax = _cacheParams.boxMax;
+    RegularGrid* slice = SliceGridAlongPlane( grid3d, pd, _textureSideSize, dataValues, _windingOrder, _rectangle3D);
+    float missingValue = slice->GetMissingValue();
+
+    // Apply opacity to missing values
+    int    textureSize = 2 * _textureSideSize * _textureSideSize;
+    float *textureValues = new float[textureSize];
```
clyne, line 185: Good place to use a smart pointer

lib/vdc/SliceGridAlongPlane.cpp
```diff
@@ -0,0 +1,241 @@
...
+        rectangle3D[1].x, rectangle3D[1].y, rectangle3D[1].z, // prpl/red 
+        rectangle3D[2].x, rectangle3D[2].y, rectangle3D[2].z, // red/teal
+        rectangle3D[3].x, rectangle3D[3].y, rectangle3D[3].z  // white/green
+    };
+    
+    windingOrder = temp;
+}
+
+VAPoR::RegularGrid* SliceGridAlongPlane(
+    VAPoR::Grid *grid3d,
```
clyne, line 189: grid3d should be const shouldn't it?

lib/vdc/SliceGridAlongPlane.cpp
```diff
@@ -0,0 +1,241 @@
...
+    planeDescription description,
+    size_t sideSize,
+    float *dataValues,
+    std::vector<double>& windingOrder,
+    //std::vector<glm::tvec3<double, glm::highp>>& rectangle3D
```
clyne, line 194: remove commented out code

include/vapor/SliceGridAlongPlane.h
clyne: Agreed. @sgpearse you don't need to include these definitions in the public header file if they are only used internally by SliceGridAlongPlane.cpp

clyne (commented):
Looks pretty good, @sgpearse. Just a few nits. 

Agree with @stasj: is there a reason to keep the Eberly library in the source? Why not drop it into the 3rd party libs?

sgpearse:
@clyne and @StasJ - comments have been addressed.

Regarding the Geometric Tools files - I agree these should be in our third-party libs.  I submitted this code because it was faster and easier.  Updating our libraries takes a considerable amount of time, and it's harder now than it has been in the past.  Resolving #2922 will help.

Either way I'm updating them now.  I'm hopeful this will be smoother than what happened with the Ospray upgrade.  I'll change this PR into a draft until our builds pass platform tests.

lib/vdc/SliceGridAlongPlane.cpp
sgpearse: Fixed.

lib/vdc/SliceGridAlongPlane.cpp
sgpearse: Yep.  Fixed.

lib/render/SliceRenderer.cpp
sgpearse: Fixed.

lib/render/SliceRenderer.cpp
sgpearse: An oversight.  The function prototype I was given in the last review was using a raw pointer and using a smart ptr didn't occur to me.  Fixed.

include/vapor/SliceRenderer.h
sgpearse: Fixed.

include/vapor/SliceGridAlongPlane.h
sgpearse: Fixed.

include/vapor/SliceGridAlongPlane.h
sgpearse: Fixed.

include/vapor/SliceGridAlongPlane.h
sgpearse: Fixed.

include/vapor/SliceRenderer.h
clyne: There are still unused private member variables, e.g. _newWaySeconds

include/vapor/SliceGridAlongPlane.h
```diff
@@ -0,0 +1,38 @@
...
+//! \param[out] windingOrder A vector of double values that contain the spatial coordinates of the vertices that correspond to the two triangles that define the texture being drawn by the "data" array.
+//! \param[out] rectangle3D A set of four vertices that define the rectangle containing the returned 2D grid's texture.  Useful for debugging.
+//!
+
+VAPoR::RegularGrid* SliceGridAlongPlane(
+    const VAPoR::Grid *grid3d, 
+    planeDescription description, 
+    size_t sideSize, 
+    //std::shared_ptr<float> data, 
+    std::unique_ptr<float>& data,
```
StasJ, line 34: Wouldn't the grid contain this data?

sgpearse: No, it actually does not.  The Grid class does contain the extents of the data, but this is not have enough information to define the vertices of the texture's triangles.  Many slices can be defined within the extents of a Grid.

include/vapor/SliceRenderer.h
sgpearse: My bad.  I took another pass and removed all unused private members and functions.

clyne (approved):

include/vapor/SliceGridAlongPlane.h
StasJ: There may be something I’m missing here, but would that not contradict the TwoDDataRenderer generating a textured triangle mesh from a Grid class?

sgpearse: I'm not completely familiar with how the TwoD renderers work, but I do know that currently they can be only oriented on the XY plane.  I believe this restriction is what allows them to render without explicitly stating their texture vertex locations.

clyne: Just adding a note that we're going to approve this as is (@StasJ need your approval) and if there is an opportunity to refactor to address the issue raised by Stas it can be done as part of #2728

StasJ (approved):

clyne:
@sgpearse there were a substantial number of commits made to this PR *after* it was approved. Most of them look like DevOps changes. Can you explain?

clyne:
@sgpearse please split into separate PRS: one for code and one for devops

sgpearse:
Fixes #2906

sgpearse:
@clyne @stasj - The DevOps changes have been backed out, and it looks like they've been merged with the acceptance of PR #2958.

Subsequent changes were needed on this branch regarding clang-format, and some new data types that were introduced into main's VAPoR namespace.  Tests are now passing and this PR is approximately where it was two weeks ago.

include/vapor/SliceRenderer.h
```diff
@@ -14,7 +14,9 @@
 #include <vapor/utils.h>
 #include <vapor/Renderer.h>
 
-#define DEBUG 1
+// clang-format off
```
clyne, line 17: Why is it necessary to disable clang-format in this relatively cut-n-dry header file ? We've not yet decided what, if anything, we plan to do with clang-format moving forward. We decided to discuss this issue at our post mortem after the release.

include/vapor/SliceRenderer.h
```diff
@@ -48,7 +50,8 @@ class RENDER_API SliceRenderer : public Renderer {
         double              zOrigin;
         std::vector<float>  tf_lut;
         std::vector<double> tf_minMax;
-        std::vector<double> boxMin, boxMax;
+        //std::vector<double> boxMin, boxMax;
```
clyne, line 53: remove commented out code.

lib/render/SliceRenderer.cpp
```diff
@@ -21,6 +21,8 @@
 
 //#define DEBUG 1
 
+// clang-format off
```
clyne, line 24: ditto disabling clang-format here.

include/vapor/SliceRenderer.h
sgpearse: Removed.  

To answer your question: 
At our meeting two weeks ago we were discussing clang-format.  I said I'd like to disable it for my work in this PR because it was impeding my work.  Nobody objected to it, and it's used on a subjective basis elsewhere in the repo.

include/vapor/SliceRenderer.h
sgpearse: Fixed.

lib/render/SliceRenderer.cpp
sgpearse: Removed.

include/vapor/SliceRenderer.h
clyne: I remember the conversation, but maybe I misunderstood. I thought there was some particular, isolated problem that you were running into, not planning on wholesale disabling the formatter.

clyne (approved):

PR #2940, opened 2021-12-20
fix #2935: flow crash

### LINKED ISSUE #2935: Replacing data used by flow renderer causes crash
Open a release build of Vapor on Casper

1. Open VDC `Bugs/2935/MFR1`
2. Create/enable a flow renderer with appropriate vars
3. Open VDC `Bugs/2935/MFR` and replace existing dataset

Vapor will crash (usually).

Somewhat nondeterministic. Using a 3.5 installer on Casper seems to be most consistent but a source release build will crash eventually too.
Other renderers appear to work.

shaomeng:
Fixes #2935

### DIFF
--- a/include/vapor/VaporField.h
+++ b/include/vapor/VaporField.h
@@ -111,7 +111,7 @@ private:
     // Member variables
     //
     std::vector<double>      _timestamps;    // in ascending order
-    VAPoR::DataMgr          *_datamgr = nullptr;
+    VAPoR::DataMgr *         _datamgr = nullptr;
     const VAPoR::FlowParams *_params = nullptr;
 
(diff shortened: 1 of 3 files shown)


StasJ (commented):
Since the cache is no longer being used, should we remove the cache to simplify the code rather than just clearing it each render call?

shaomeng:
> Since the cache is no longer being used, could we remove the cache to simplify the code rather than just clearing it each render call?

[This comment](https://github.com/NCAR/VAPOR/issues/2935#issuecomment-997214009) explains the reasoning which I tend to agree.

StasJ (approved):

PR #2963, opened 2022-01-10
fix vapor_wrf.py

### LINKED ISSUE #2962: vapor_wrf.py broken
To reproduce:

1) load vapor with any dataset
2) create a new python script with any input variable and a new output variable
3) test the following script:

`import vapor_wrf`

Error: ModuleNotFoundError: No module named 'vapor'

sgpearse:
Fixes #2962: broken file vapor_wrf.py by removing non-existent module.


clyne:
This PR will allow vapor_wrf to be imported, but have you tested the functions in the module to verify that they still work?

sgpearse:
Yes, I was able to compute RH successfully only after deleting 'import vapor'

clyne:
> Yes, I was able to compute RH successfully only after deleting 'import vapor'

What about functions that use the vapor API such as PV, or CT

sgpearse:
> What about functions that use the vapor API such as PV, or CT

My tests weren't exhaustive, and PV and CT do not work under my test criteria.

clyne:
Well I'm guessing that none of the functions in vapor_wrf that reference the old vapor API will work. So I don't think we want to accept this PR if it includes several functions that are broken. Note that vapor_wrf reimplements a bunch of functions that exist in GeoCAT's [wrf_python](https://wrf-python.readthedocs.io/en/latest/) package. Thus the long term solution that should be possible by the Python API work @stasj is doing, is to simply allow users to install wrf_python on their system and import the package into vapor. We could then retire our vapor_wrf module, and eliminate a lot of technical dept. The short term solution might be to figure out which vapor_wrf modules work correctly, and only include those that do in vapor_wrf.

sgpearse:
@clyne - sounds good.  I'll convert this to a draft until I've removed the non-working functions in this PR.

sgpearse:
@clyne - I've removed all non-functional methods from vapor_utils.py and vapor_wrf.py.  I've documented the errors I encountered in [this document](https://docs.google.com/document/d/1AVwXz6k4WLrH_J46oBtv5Rmi8--ZSpAfCbrAUKVfLk4/edit?usp=sharing).  The doc contains a session file that you can use to reproduce errors after importing a Katrina file into the session.  Note: The vapor_utils.py tests did account for Katrina's staggered grid.

clyne:
@sgpearse something weird is going on with vapor_utils.py. With the possible exception of `vector_rotate`, which I have no idea what it does, all of these utilities have been migrated to VAPOR3 and *were* working. I've attached a session file that uses CurlFinDiff (which in turn calls DerivFinDiff), and it seems to work just fine. The weird part is that DerivFinDiff invokes deriv_findiff? functions, which do not appear to be defined. There are, however, derivative functions defined that are named _deriv_findiff? (with a leading underscore). My limited understanding of Python is that the leading underscore is simply a convention of Python for internal functions that are not  intended "public" in the C++ sense. So I don't know why an invocation of deriv_findiff2, for example, would resolve to _deriv_findiff2. 

My suggestion is to make sure the deriv_findiff? functions are consistently named with a leading underscore, and retest. Start with the attached session file since it is working for me. 

Not sure that this is a priority for 3.6...

[Enstrophy.vs3.gz](https://github.com/NCAR/VAPOR/files/7872138/Enstrophy.vs3.gz)

sgpearse:
@clyne It turns out that deriv_findiff was a function in the old Vapor2 code base that never made it into vapor_utils.py.  After adding it, some of the broken functions in vapor_utils.py started working again.  

[This table](https://docs.google.com/spreadsheets/d/18ByW5risbkLT2Of4Ty_LpspVicJFqf7f9gFQLnBJc6g/edit#gid=0) shows which functions are now working after this fix.  

Many of the vapor_wrf.py functions still do not work because they rely on a module that was deprecated in Vapor3, which can be seen [here](https://sourceforge.net/p/vapor/git/ci/master/tree/share/python/vapor.py#l57).  The SHEAR function now works, after unstaggering the input vector variables.

clyne:
@sgpearse I believe deriv_findiff did indeed make it into VAPOR3 vapor_utils.py, but it was renamed to DerivFinDiff. DerivFinDiff is a "public" function (not that Python has a notion of public and private functions like C++ does), and it is also used internally by other public functions. For example, GradFinDif calls DerivFinDiff. The public function DivFinDiff calls deriv_findiff. I believe that it *should* be calling DerivFinDiff. I.e. you don't need to import deriv_findiff from vapor2. You just need to substitute DerivFinDiff for deriv_fin_diff. Note also that there are a couple of places where deriv_findiff2 and deriv_findiff4 are called. These functions should have an underscore prepended to them.

sgpearse:
@clyne - I believe you're correct.  My tests are currently passing when replacing deriv_findiff with DerivFindiff, but I'd like to re-test all functions and update [the spreadsheet](https://docs.google.com/spreadsheets/d/18ByW5risbkLT2Of4Ty_LpspVicJFqf7f9gFQLnBJc6g/edit#gid=0).  I'll re-request review tomorrow when they hopefully all pass.

clyne (approved):
Thanks for doing this, @sgpearse

PR #2970, opened 2022-01-12
Fix #2844 VAPOR fails to reliably identify file formats of files supplied on the command line

### LINKED ISSUE #2844: VAPOR fails to reliably identify file formats of files supplied on the command line
VAPOR can't always automatically determine a data file's type (format) when the file is supplied on the command line. Moreover, the current implementation can lead to misleading error messages (see #2842). 

The current file format detection algorithm works as follows: for each supported file type attempt to initialize the DC class for that format using the derived DC classes' Initialization method. If DC::Initialize() succeeds, assume that the correct DC class (and thus the format) has been identified.

A couple of problems result from this approach:

1. Some DC classes will successfully open (initialize) files that are of a different format. For example, the DCCF class will successfully initialize itself with some WRF files. The workaround for this is to test WRF files before CF files. But this problem will only grow as more formats are added.
2. If a file is corrupt/invalid in some way the DC::Initialize method will fail and thus the file will not be correctly identified. Moreover, the user will only be informed that the file type could not be identified, and not informed of the reason why the initialization failed.

Fixing this problem is tricky. Some of the file formats don't have 'magic numbers' that would enable a reliable file detection method. Thus it might never be possible to have a fool proof method. However, adding a DC::FileDetect() method that looks for specific items expected to be found in a file of a given type could improve the situation.

StasJ:
Fix #2844

### DIFF
--- a/lib/vdc/BOVCollection.cpp
+++ b/lib/vdc/BOVCollection.cpp
@@ -115,7 +115,7 @@ int BOVCollection::Initialize(const std::vector<std::string> &paths)
 
         header.open(paths[i]);
         if (header.is_open()) {
-            if (getFileSize(paths[i]) > 1000) {
+            if (getFileSize(paths[i]) > 1000000) {
                 SetErrMsg(("BOV header file larger than 1MB.  This text file shouldn't need to be larger than a few KB." + paths[0]).c_str());
                 return -1;
             }
--- a/lib/vdc/BOVCollection.cpp
+++ b/lib/vdc/BOVCollection.cpp
@@ -63,16 +63,16 @@ const std::string BOVCollection::_floatFormatString = "FLOAT";
 const std::string BOVCollection::_doubleFormatString = "DOUBLE";
 
 namespace {
-    size_t getFileSize(std::string filename) // path to file
-    {
-        FILE *p_file = NULL;
-        p_file = fopen(filename.c_str(),"rb");
-        fseek(p_file,0,SEEK_END);
-        size_t size = ftell(p_file);
-        fclose(p_file);
-        return size;
-    }
+size_t getFileSize(std::string filename)    // path to file
+{
+    FILE *p_file = NULL;
+    p_file = fopen(filename.c_str(), "rb");
+    fseek(p_file, 0, SEEK_END);
+    size_t size = ftell(p_file);
+    fclose(p_file);
+    return size;
 }
+}    // namespace
 
 BOVCollection::BOVCollection()
 : _time(_defaultTime), _dataFile(_defaultFile), _dataFormat(_defaultFormat), _variable(_defaultVar), _byteOffset(_defaultByteOffset), _divideBrick(_defaultDivBrick), _dataEndian(_defaultEndian),
@@ -115,7 +115,7 @@ int BOVCollection::Initialize(const std::vector<std::string> &paths)
 
         header.open(paths[i]);
         if (header.is_open()) {
-            if( getFileSize(paths[i]) > 1000 ) {
+            if (getFileSize(paths[i]) > 1000) {
                 SetErrMsg(("BOV header file larger than 1MB.  This text file shouldn't need to be larger than a few KB." + paths[0]).c_str());
                 return -1;
             }
--- a/lib/vdc/BOVCollection.cpp
+++ b/lib/vdc/BOVCollection.cpp
@@ -62,6 +62,18 @@ const std::string BOVCollection::_intFormatString = "INT";
 const std::string BOVCollection::_floatFormatString = "FLOAT";
 const std::string BOVCollection::_doubleFormatString = "DOUBLE";
 
+namespace {
+    size_t getFileSize(std::string filename) // path to file
+    {
+        FILE *p_file = NULL;
+        p_file = fopen(filename.c_str(),"rb");
+        fseek(p_file,0,SEEK_END);
+        size_t size = ftell(p_file);
+        fclose(p_file);
+        return size;
+    }
+}
+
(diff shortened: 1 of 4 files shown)


sgpearse (commented):
On OSX, my macbook hangs when loading the CF file 24Maycontrol.04001.000000.nc via the command line.

This file can be found on glade at /glade/p/cisl/vast/vapor/data/Source/CM1/Orf/24Maycontrol/24Maycontrol.04000.000000.nc

StasJ:
> On OSX, my macbook hangs when loading the CF file 24Maycontrol.04001.000000.nc via the command line.
> 
> This file can be found on glade at /glade/p/cisl/vast/vapor/data/Source/CM1/Orf/24Maycontrol/24Maycontrol.04000.000000.nc

It appears that dataset is causing the BOV reader to hang.

sgpearse:
@clyne I can write a ticket so we can move forward, but I believe that the BOV reader is the last entry to be tested.  Shouldn't the CF reader prevent the BOV from being triggered?

clyne:
> @clyne I can write a ticket so we can move forward, but I believe that the BOV reader is the last entry to be tested. Shouldn't the CF reader prevent the BOV from being triggered?

No, it should not. Per the conversation above, all of the readers are tested and if multiple readers are able to open the same file than auto-determination of the file format is not possible. Ideally, the BOV reader *should* not be hanging when confronted with an invalid file.

sgpearse:
In that case the current solution makes loading any dataset from the command line hang.  Accepting this as-is would send us backwards so I think we should fix the problem before accepting.

The BOV reader scans its input file line-by-line for tokens until it reaches the end of the file.  I'm not sure what a good strategy is for it detecting invalid files without reading the input file.  One option is to remove the BOV reader check, but I often use this feature for BOV and don't want to lose the option.

clyne:
> In that case the current solution makes loading any dataset from the command line hang. Accepting this as-is would send us backwards so I think we should fix the problem before accepting.
> 
> The BOV reader scans its input file line-by-line for tokens until it reaches the end of the file. I'm not sure what a good strategy is for it detecting invalid files without reading the input file. One option is to remove the BOV reader check, but I often use this feature for BOV and don't want to lose the option.

BOV files have to be ASCII, correct? One way is to read the first n (some reasonable number) bytes and see if they are all ASCII. @stasj can probably give guidance on testing a byte for ASCII-ness

sgpearse:
I haven't looked deeply into the problem yet, so I'm all ears for suggestions.  Rregarding ASCII, the characters could be UTF-8/16 or something else that ifstream handles.

clyne:
You might also just put a limit on how much data your parser reads, and terminate it after X number of bytes. Since are no magic numbers for these files, voodoo guesswork is all you can do.

sgpearse:
@clyne and @stasj, I [added a rule](https://github.com/NCAR/VAPOR/pull/2970/commits/a19880bcba74b070cf345106ef9d61b6c112507d) to the BOV reader to reject headers < 1MB in size.  They shouldn't ever be larger than a couple KB.

However I'm unable to get wrf, ugrid, bov, or ROMS files loaded through the command line.  Leigh's MayControl files seem to work.  I feel like this PR would move us backwards by disabling a frequently used tool (at least among developers).   Thoughts?

clyne:
> @clyne and @StasJ, I [added a rule](https://github.com/NCAR/VAPOR/pull/2970/commits/a19880bcba74b070cf345106ef9d61b6c112507d) to the BOV reader to reject headers < 1MB in size. They shouldn't ever be larger than a couple KB.
> 
> However I'm unable to get wrf, ugrid, bov, or ROMS files loaded through the command line. Leigh's MayControl files seem to work. I feel like this PR would move us backwards by disabling a frequently used tool (at least among developers). Thoughts?

@StasJ what are your thoughts on simply adding a '-ftype' option so the user can specify the format without ambiguity and we can be done with this?

StasJ:
> > @clyne and @StasJ, I [added a rule](https://github.com/NCAR/VAPOR/pull/2970/commits/a19880bcba74b070cf345106ef9d61b6c112507d) to the BOV reader to reject headers < 1MB in size. They shouldn't ever be larger than a couple KB.
> > However I'm unable to get wrf, ugrid, bov, or ROMS files loaded through the command line. Leigh's MayControl files seem to work. I feel like this PR would move us backwards by disabling a frequently used tool (at least among developers). Thoughts?
> 
> @StasJ what are your thoughts on simply adding a '-ftype' option so the user can specify the format without ambiguity and we can be done with this?

That makes sense. I believe at our meeting we decided such functionality would be added in a future PR which is why it wasn’t included here. I can add it to this PR.

sgpearse (approved):

lib/vdc/BOVCollection.cpp
```diff
@@ -103,6 +115,11 @@ int BOVCollection::Initialize(const std::vector<std::string> &paths)
 
         header.open(paths[i]);
         if (header.is_open()) {
+            if (getFileSize(paths[i]) > 1000) {
+                SetErrMsg(("BOV header file larger than 1MB.  This text file shouldn't need to be larger than a few KB." + paths[0]).c_str());
```
shaomeng, line 119: I think you meant 1KB here?

clyne: Good catch @shaomeng. Can you confirm the size here, @sgpearse, and then I'll accept.

sgpearse: Fixed.  Thanks sam.

clyne (approved):

Issue #2969, opened 2022-01-11
SliceGridAlongPlane returns incorrect grid


StasJ:
1. Slice a 3D grid using SliceGridAlongPlane

Notice that the all Z values for the Grid's coordinates are always 0
Rendering the grid with the wireframe renderer also reveals that the grid is incorrect.

sgpearse:
@clyne and @stasj - I believe the problem is that the RegularGrid class cannot currently store the coordinates of 2D grids **that are not axis aligned**.

You can see the Slice's constructor [here](https://github.com/NCAR/VAPOR/blob/eb5f3a40783b958293e9aec5eca2141ac83bdc31/lib/vdc/SliceGridAlongPlane.cpp#L239), where dimensionality and coordinate extents are defined.  The dimensions are [ sampleSize x sampleSize x 1 ] followed by the Box extents.  I believe the Grid class currently assigns a value of 0 to grids that have a dimension equal to size 1.

clyne:
@sgpearse prior to the merge of PR #2943 that was certain the case: only axis-aligned 2D grids were supported. It now *should* be possible to describe a plane that is not axis-aligned. However, it is possible that there is a bug. A couple of thoughts on how to move forward:

1. Define an arbitrary 2D coordinate system (e.g. unit normalized) in the RegularGrid, and transform the 3D world coordinates into the 2D used by the Grid. 
2. Try and run down and fix whatever problem there may be in the Grid class with 3D coordinates and a 2D grid. I have limited cycles to help out on this, but if you can provide an example that demonstrates the problem I can take a look at it. Ultimately, we should add a test case to our test suite.

sgpearse:
@clyne - Got it.  I'll update soon with findings and possibly further questions.

sgpearse:
@clyne - 

**Suggestion 1** 
I'm a little unclear.  Are you suggesting that the SliceRenderer and ContourRenderer should deduce the user coordinates of grid cells, after receiving the grid from SliceGridAlongPlane()?

**Suggestion 2** 
Update - The problem is that the RegularGrid class has a notion of _delta, which is a VAPoR::CoordType that describes the increment between vertices along the grid's xyz axes.   It is used to determine grid point locations in user-space.

[...]

Issue #2968, opened 2022-01-11
Regression: SliceGridAlongPlane() does not honor Box extents


sgpearse:
The new SliceGridAlongPlane function does not apply missing values to grid points outside of the Box.

To reproduce:
1) Load any dataset, then create and enable a Slice renderer
2) Reduce the region of the slice in the Geometry tab, and the Slice's texture will render values outside of the Box

PR #2972, opened 2022-01-12
Fix problems regarding the Slicer's Box

### LINKED ISSUE #2971: Slice crash
SliceGridAlongPlane() will crash when the Box extents of a slice renderer lie outside of the slice plane.

To reproduce:
1) Create a Slice renderer with any dataset
2) Rotate it by 90 degrees on the X axis
3) Move the box so its extents are outside of the slice plane.  Vapor will crash.

sgpearse:
Fixes #2971 and #2968.

#2969 is still a work in progress, but my work on it will be based on top of these two fixes.

### DIFF
--- a/include/vapor/SliceGridAlongPlane.h
+++ b/include/vapor/SliceGridAlongPlane.h
@@ -8,6 +8,7 @@
 class RegularGrid;
 
 struct planeDescription {
+    size_t sideSize;
     std::vector<double> origin;
     std::vector<double> rotation;
     VAPoR::CoordType boxMin;
@@ -33,8 +34,6 @@ struct planeDescription {
 VDF_API VAPoR::RegularGrid* SliceGridAlongPlane(
     const VAPoR::Grid *grid3d, 
     planeDescription description, 
-    size_t sideSize, 
-    //std::shared_ptr<float> data, 
     std::unique_ptr<float>& data, 
     std::vector<double>& windingOrder, 
     std::vector<double>& rectangle3D
--- a/include/vapor/SliceGridAlongPlane.h
+++ b/include/vapor/SliceGridAlongPlane.h
@@ -12,6 +12,8 @@ struct planeDescription {
     std::vector<double> rotation;
     VAPoR::CoordType boxMin;
     VAPoR::CoordType boxMax;
+    VAPoR::CoordType domainMin;
+    VAPoR::CoordType domainMax;
 };
 
 //! Create a 2D grid that is sampled along the orientation of a 3D grid.
--- a/include/vapor/SliceGridAlongPlane.h
+++ b/include/vapor/SliceGridAlongPlane.h
@@ -17,11 +17,16 @@ struct planeDescription {
     VAPoR::CoordType domainMax;
 };
 
-//! Create a 2D grid that is sampled along the orientation of a 3D grid.
-//! The "planeDescription" variable includes parameters that define the
-//! 2D grid's origin, rotation, and extents.  The sampling rate is defined
-//! by the "sideSize" parameter, which defines how many samples along the X
-//! and Y axes are taken along the plane inside of the 3D grid.
+//! Create a 2D RegularGrid that samples a given 3D Grid object along a plane, 
+//! through the entire domain.  The "planeDescription" variable includes parameters 
+//! that define the 2D plane's origin, rotation, and valid extents.  Points on the
+//! 2D RegularGrid that are outside of the valid extents will be assigned as
+//! missingValues.
+//! 
+//! Note: The boxMin and boxMax values within the planeDescription should come from
+//! the current RenderParams' Box class.  The domainMin and domainMax values should 
+//! come from the DataMgr::GetVariableExtents function.  Taking these values from
+//! the given 3D grid can produce erroneous results when the 3D grid's region is reduced.
 //!
 //! \param[in] grid3d A variable's 3D grid to be sampled along a plane, to generate a 2D grid object
 //! \param[in] description A set of std::vector<dobule> values that define the origin, rotation, and extents of the returned 2D grid
--- a/lib/render/SliceRenderer.cpp
+++ b/lib/render/SliceRenderer.cpp
@@ -163,13 +163,13 @@ void SliceRenderer::_resetColormapCache()
     glGenTextures(1, &_colorMapTextureID);
     glActiveTexture(GL_TEXTURE0);
     glBindTexture(GL_TEXTURE_1D, _colorMapTextureID);
-    #ifdef DEBUG
+#ifdef DEBUG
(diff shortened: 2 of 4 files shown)


include/vapor/SliceGridAlongPlane.h
```diff
@@ -8,10 +8,13 @@
 class RegularGrid;
 
 struct planeDescription {
+    size_t sideSize;
     std::vector<double> origin;
     std::vector<double> rotation;
     VAPoR::CoordType boxMin;
     VAPoR::CoordType boxMax;
+    VAPoR::CoordType domainMin;
+    VAPoR::CoordType domainMax;
```
StasJ, line 17: The 3D `Grid` passed into `SliceGridAlongPlane` already has the function `Grid::GetUserExtents` which returns the data extents. Could that just be used rather than passing around this parameter here?

sgpearse: No, and this is reason why #2968 happened.

The SliceRenderer requests a 3D grid based on the SliceRenderer's Box.  When a new Grid is created, its extents can differ from the Box extents that it is fed because some nodes get clipped off.

The fix here is to use a set of domain and box extents that are not derived from the passed 3D grid.

StasJ (approved):

lib/render/SliceRenderer.cpp
```diff
@@ -140,6 +136,16 @@ void SliceRenderer::_resetCache()
 
     _getExtents(_cacheParams.boxMin, _cacheParams.boxMax);
 
+    // clang-format off
+    _dataMgr->GetVariableExtents(_cacheParams.ts,
```
clyne, line 140: It looks like you are getting the variable extents here (_cacheParams.domainMin, _cacheParams.domainMax), and the only place you are using them is to pass them into SliceGridAlongPlane(), which also takes a grid3d argument. Why not simply call grid3d->GetUserExtents() from inside SliceGridAlongPlane()? The results should be identical, and it would simplify the code a good bit.

lib/render/SliceRenderer.cpp
```diff
@@ -173,11 +179,15 @@ int SliceRenderer::_regenerateSlice()
...
+    pd.sideSize = _textureSideSize;
     pd.origin = {_cacheParams.xOrigin, _cacheParams.yOrigin, _cacheParams.zOrigin};
     pd.rotation = {_cacheParams.xRotation, _cacheParams.yRotation, _cacheParams.zRotation};
     pd.boxMin = _cacheParams.boxMin;
     pd.boxMax = _cacheParams.boxMax;
-    RegularGrid *slice = SliceGridAlongPlane(grid3d, pd, _textureSideSize, dataValues, _windingOrder, _rectangle3D);
+    pd.domainMin = _cacheParams.domainMin;
+    pd.domainMax = _cacheParams.domainMax;
+    RegularGrid *slice = SliceGridAlongPlane(grid3d, pd, dataValues, _windingOrder, _rectangle3D);
+    if (slice == nullptr) return 0;
```
clyne, line 190: Isn't this an error condition? Shouldn't it be returning -1, not 0?

lib/vdc/SliceGridAlongPlane.cpp
```diff
@@ -163,7 +168,15 @@ void populateData(
             double y = yStart + i*delta.y;
             glm::tvec3<double, glm::highp> samplePoint = origin + x*axis1 + y*axis2;
             VAPoR::CoordType p = {samplePoint.x, samplePoint.y, samplePoint.z};
-            dataValues.get()[index] = grid->GetValue(p);
+
+            if ( p[0] <= min[0] || p[0] >= max[0] ||
```
clyne, line 172: Are you finding it necessary to explicitly compare p with the extents of the Grid object? Grid::GetValue(p) should return missingValue if p is outside of the domain.If it is not then there is a bug that should be reported.

lib/render/SliceRenderer.cpp
sgpearse: This is because I need the full extents of a variable, not the extents of the current Grid, which may have had its region reduced.

The promise that the SliceGridAlongPlane makes is that it samples a variable along a plane throughout the entire domain.  Samples that are outside of the Box are counted as missingValues.  What I'm doing here might be a little unclear, so I'll write a note below discussing the strategy.

lib/render/SliceRenderer.cpp
sgpearse: Yes. I missed cleaning this up while debugging.  Fixed.

lib/vdc/SliceGridAlongPlane.cpp
sgpearse: This is checking whether the sample point is outside of the Box, no the domain.

include/vapor/SliceGridAlongPlane.h
clyne: Got it

lib/render/SliceRenderer.cpp
clyne: Got it.

lib/vdc/SliceGridAlongPlane.cpp
clyne: Understood. Since both @StasJ and I were confused about this and explicit comment might be helpful.

A small nit, perhaps, but do you really want to (need to) exclude points on the boundary (e.g. using <= instead of <)? I doubt that we are completely consistent about this, and in most instances it may not matter, but I think elsewhere points that are on the boundary are considered "valid".

sgpearse: Yes, I agree and I'm going to write something up to put in the file.

I can go ahead and remove the '='.  These were added during debugging.  They had no effect but weren't removed.

sgpearse:
@clyne and @StasJ - here's some context to my strategy for these bugs.  I've added a summary of this in SliceGridAlongPlane.h.

The SliceRenderer was crashing when the Box and plane did not intersect.  This was happening because the returned 2D grid was only being sampled within the bounds of the current Box.  When the plane and Box did not intersect the returned grid would be invalid, causing a crash.  

One option was to add logic in the SliceRenderer to handle this special case, however this option would require every renderer using SliceGridAlongPlane() to handle it as well.  

The alternative I chose was to have SliceGridAlongPlane() return a valid grid in this case.  The returned grid would be sampled along a given plane, and would have missingValues at points that do not lie within the Box.  The Slice and Contour logic would be simplified by letting them operate on the returned grid in the same way, whether there are valid values or not.

Note that I am using the domain extents gathered from the DataMgr, and the Box extents from RenderParams.  I'm not using the extents gathered from the 3D grid because those can be in disagreement with the DataMgr and RenderParams.  This occurs due to roundoff whenever the Grid's extents are reduced, which incrementally reduces the Grid's dimension lengths.

Hope that makes sense.

clyne (approved):

PR #2974, opened 2022-01-21
ArbitrarilyOrientedRegularGrid - Issue 2969

sgpearse:
Fixes issue #2969.

Adds the new class ArbitrarilyOrientedRegularGrid to generated 2D slices from a 3D RegularGrid.

Additionally fixes a minor coordinate error in the sampling code and a backwards compatibility issue with old .vs3 files that contain SliceRenderers.

There are getter functions for the winding order of the data if it is to be used in a texture, as well as the coordinates of the outer corners.  These getters were kept after discussion with @stasj about the requirements for the ContourRenderer, which slightly differs from the requirements of the SliceRenderer.  The ContourRenderer needs the coordinates of each point that is sampled, whereas the SliceRenderer needs the extents of the entire area being sampled.

(The below comment seems to now be resolved)
~~Note: On my local repository I have removed the files SliceGridAlongPlane.cpp/.h, but they still seem to be here in the remote.  For now, I'm going to give github some time in case it needs to refresh something.~~


include/vapor/ArbitrarilyOrientedRegularGrid.h
```diff
@@ -0,0 +1,103 @@
...
+
+#include <ostream>
+#include <vector>
+#include <vapor/common.h>
+#include <vapor/RegularGrid.h>
+#include <glm/glm.hpp>
+
+// clang-format off
+
+struct planeDescription {
```
clyne, line 11: Please document the public members of this class

include/vapor/ArbitrarilyOrientedRegularGrid.h
```diff
@@ -0,0 +1,103 @@
...
+
+namespace VAPoR {
+
+//! Create a 2D RegularGrid that is rotated along an arbitrary orientation.  This new grid
+//! samples a given 3D Grid object along a plane, through the entire domain.  The "planeDescription" 
+//! variable includes parameters that define the 2D plane's origin, rotation, and valid extents.  Points on the
+//! 2D RegularGrid that are outside of the valid extents will be assigned as
+//! missingValues.
+//!
+//! Note: The boxMin and boxMax values within the planeDescription should come from
```
clyne, line 29: The need to pass in both the box and domain extents is really confusing to me. It should not be necessary. grid3d knows its extents (Grid::GetUserExtents). box{Min,Max} constrain the region of the plane that you wish to intersect with grid3d. The box may lie entirely inside, outside, or intersect grid3d's extents. The code should handle all possibilities. So I guess a question is why are "erroneous results" produced when grid3d's extents used? Sounds like a bug that we should fix, rather than complicate the API and not have the code be robust.

include/vapor/ArbitrarilyOrientedRegularGrid.h
```diff
@@ -0,0 +1,103 @@
...
+    //! \copydoc Grid::GetUserCoordinates()
+    //
+    virtual void GetUserCoordinates(const DimsType &indices, CoordType &coords) const override;
+
+    //! Retrieve the winding order that pertains to the generated array of data, if it is
+    //! to be drawn as a texture as is done in the SliceRenderer's vertex buffer.  The winding order is a series
+    //! of six 3D vertices that describe two triangles that define how the texture is drawn in space.
+    void GetWindingOrder( std::vector<double>& windingOrder ) const;
+
+    //! Retrieve the 3D coordinates of the corners of the sampled data.  This is a set of four 3D coordinates,
```
clyne, line 64: Isn't this what Grid::GetUserExtents() should return? Is there a need to introduce a new function, Get3DRectangle()

include/vapor/ArbitrarilyOrientedRegularGrid.h
```diff
@@ -0,0 +1,103 @@
...
+    std::string        GetType() const override { return (GetClassType()); }
+
+    //! \copydoc Grid::GetUserCoordinates()
+    //
+    virtual void GetUserCoordinates(const DimsType &indices, CoordType &coords) const override;
+
+    //! Retrieve the winding order that pertains to the generated array of data, if it is
+    //! to be drawn as a texture as is done in the SliceRenderer's vertex buffer.  The winding order is a series
+    //! of six 3D vertices that describe two triangles that define how the texture is drawn in space.
+    void GetWindingOrder( std::vector<double>& windingOrder ) const;
```
clyne, line 62: The "winding order" and "two triangles" confuse me. Does this class define a plane, or two adjacent triangles?

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: Fixed.

include/vapor/ArbitrarilyOrientedRegularGrid.h
```diff
@@ -0,0 +1,103 @@
...
+//! come from the DataMgr::GetVariableExtents function.  Taking these values from
+//! the given 3D grid can produce erroneous results when the 3D grid's region is reduced.
+//!
+//! \param[in] grid3d A variable's 3D grid to be sampled along a plane, to generate a 2D grid object
+//! \param[in] description A set of std::vector<dobule> values that define the origin, rotation, and extents of the returned 2D grid
+//! \param[in] dims The sampling rate that the new grid will be defined upon, along its X and Y axes
+//! \param[out] data An array of sideSize*sideSize values that contain floating point values of the sampled data.
+//!
+
+class VDF_API ArbitrarilyOrientedRegularGrid : public RegularGrid {
```
clyne, line 46: The Grid and RegularGrid class define several methods that operate on CoordType arguments. Other than GetUserCoordinates, which you have redefined, the others will produce incorrect results. We said that we would not bother to reimplement these until they are needed. But you should at lease have them throw an assert or something in case someone does try to use them.

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: No.  Get3DRectangle() returns four vertices at the corners of the sampled plane needed for the texture coordinate buffer.  

If we have a plane that is rotated 45 degrees on X, GetUserExtents() would return two 3D coordinates that describe the volume enclosing the plane, which isn't what's needed.  The 3D rectangle also cannot be inferred from this volume because a rotation of -45 degrees would return the same volume but describe a different plane.

sgpearse: PS - [See here](https://github.com/NCAR/VAPOR/blob/dd7ac4ee9d3aee8e508b6a40723e83c548ef9337/lib/render/SliceRenderer.cpp#L101) for where this structure is needed in the SliceRenderer.

include/vapor/ArbitrarilyOrientedRegularGrid.h
```diff
@@ -0,0 +1,103 @@
...
+//!
+//! Note: The boxMin and boxMax values within the planeDescription should come from
+//! the current RenderParams' Box class.  The domainMin and domainMax values should
+//! come from the DataMgr::GetVariableExtents function.  Taking these values from
+//! the given 3D grid can produce erroneous results when the 3D grid's region is reduced.
+//!
+//! \param[in] grid3d A variable's 3D grid to be sampled along a plane, to generate a 2D grid object
+//! \param[in] description A set of std::vector<dobule> values that define the origin, rotation, and extents of the returned 2D grid
+//! \param[in] dims The sampling rate that the new grid will be defined upon, along its X and Y axes
+//! \param[out] data An array of sideSize*sideSize values that contain floating point values of the sampled data.
```
clyne, line 43: Since we've take the trouble to create this class, I think we should NOT pass in a `data` argument, but instead have the class manage the space internally. Ideally, all of the Grid classes would not require a `blks` argument, but this has significant implications for DataMgr caching that do not apply here. 

Yes, this would mean that the SliceRenderer would no longer have a back door into the classes internal memory, but it probably shouldn't anyhow.

clyne (commented):
@sgpearse I think overall this is a much better design. Still a few things we need to address though.

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: It describes both.

The [texture coordinate buffer](https://github.com/NCAR/VAPOR/blob/dd7ac4ee9d3aee8e508b6a40723e83c548ef9337/lib/render/SliceRenderer.cpp#L101) needs the four coordinates that describe the corners of the plane.

We need to draw the plane as a pair of triangles.  [The vertex buffer](https://github.com/NCAR/VAPOR/blob/dd7ac4ee9d3aee8e508b6a40723e83c548ef9337/lib/render/SliceRenderer.cpp#L112) needs to know the winding order.

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: This was done last week in PR #2972, to fix #2971.  We had a bit of a conversation on it and everyone gave a thumbs up before the merge.  I'll copy/paste my response from last week below.  The PR linked above will have more context.  

Let me know if there's still something that's unclear though.

> The SliceRenderer was crashing when the Box and plane did not intersect. This was happening because the returned 2D grid was only being sampled within the bounds of the current Box. When the plane and Box did not intersect the returned grid would be invalid, causing a crash.

> One option was to add logic in the SliceRenderer to handle this special case, however this option would require every renderer using SliceGridAlongPlane() to handle it as well.

> The alternative I chose was to have SliceGridAlongPlane() return a valid grid in this case. The returned grid would be sampled along a given plane, and would have missingValues at points that do not lie within the Box. The Slice and Contour logic would be simplified by letting them operate on the returned grid in the same way, whether there are valid values or not.

> Note that I am using the domain extents gathered from the DataMgr, and the Box extents from RenderParams. I'm not using the extents gathered from the 3D grid because those can be in disagreement with the DataMgr and RenderParams. This occurs due to roundoff whenever the Grid's extents are reduced, which incrementally reduces the Grid's dimension lengths.

> Hope that makes sense.

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: My original intent was to delete all unused functions but I ran into compiler errors when deleting virtual functions.  Looking at it again, I'm seeing that I can delete functions as long as I remove the const qualifier on them.  I'm not sure why.  A deleted function can't even be called, no less modify an object.  Who knows...

I've gone ahead and deleted the other forms of GetUserCoordinates().

PS - There are dozens of other public functions that I'm not certain will function either.  I can test and delete these too but it looks like it might be a more substantial effort.

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: I thought about this too.  But unless I'm missing something, the only way to pass a Grid a blks pointer is through its constructor.  

ArbitrarilyOrientedRegularGrid's constructor receives its blks pointer and passes it to the RegularGrid constructor via the initialization list.  Without a function like Grid::SetBlks(), is it possible for ArbitrarilyOrientedRegularGrid's constructor to allocate data and configure it as is done in its parent classes?

include/vapor/ArbitrarilyOrientedRegularGrid.h
clyne: Users of this class need to be able to handle the case where the grid consists entirely of missing values. That is possible even if the plan is located   entirely within the valid domain. So I don't understand the argument that "When the plane and Box did not intersect the returned grid would be invalid, causing a crash.". It should not crash simply because the plane is outside of the grid and contains nothing but missing values.

include/vapor/ArbitrarilyOrientedRegularGrid.h
clyne: OK. So Get3DRectangle() returns the coordinates of the four corner vertices. So this is a convenience function that is equivalent to calling Grid::GetUserCoordinates() with the indices of the four corner points?

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: > It should not crash simply because the plane is outside of the grid and contains nothing but missing values.

It does not crash in this case.  Let me see if this helps:

Previous to #2971, the extents of the plane were defined according to the Box.  I would find where the plane intercepts the edges of the Box, and use these intercepts to find the Minimum Area Rectangle (MAR) enclosing all of these intercepts.

When the plane did not intersect the Box, there would be no intercepts and the MAR algorithm would crash.  One option was to return a nullptr in this case.  However this would 1) require logic in the Slicer to handle this case, and 2) require each Renderer to handle the special case where the returned grid is invalid.

Instead, we can find where the plane intercepts the edges of the domain, because the plane will always lie within the domain.  When we sample along the plane inside the domain, we can simply attribute missing values to coordinates outside of the Box.  Coordinates inside the box can still attribute MissingValues.

This way the Renderer can always rely on getting a valid grid according to the parameters its using.  No special case handling is needed.  If the Renderer gets a slice that has all missing values it operates on it the same as if there were scalar data.

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: Almost.  This is hard to explain without a picture.  Do you have time for a screen share?  If not I can try to explain here.

include/vapor/ArbitrarilyOrientedRegularGrid.h
clyne: Still confused. Polygon winding orders that I'm familiar with are typically expressed as rotations about a normal: clockwise, or counter-clockwise (CCW), not as a list of points. So maybe that is part of my confusion. Moreover, the convention for 2D Grids is that the winding order is CCW. 2D Grids are always constructed this way. Since you are constructing the grid from scratch, don't you have the ability to use a consistent winding order?

include/vapor/ArbitrarilyOrientedRegularGrid.h
clyne: I don't think you have to test them. You can assume that they're going to break if they are operating in 2D space.

include/vapor/ArbitrarilyOrientedRegularGrid.h
clyne: Unless I'm mistaken you can call `new` within the constructor member initializer

StasJ (commented):
I tested the functionality and it works perfectly! I agree with John's comments which cover most of the code review but I left a few additional minor notes.

include/vapor/ArbitrarilyOrientedRegularGrid.h
```diff
@@ -0,0 +1,114 @@
...
+    virtual void GetUserCoordinates (size_t i, size_t j, size_t k, double &x, double &y, double &z) = delete;
+
+    //! Retrieve the winding order that pertains to the generated array of data, if it is
+    //! to be drawn as a texture as is done in the SliceRenderer's vertex buffer.  The winding order is a series
+    //! of six 3D vertices that describe two triangles that define how the texture is drawn in space.
+    void GetWindingOrder( std::vector<double>& windingOrder ) const;
+
+    //! Retrieve the 3D coordinates of the corners of the sampled data.  This is a set of four 3D coordinates,
+    //! used to configure the SliceRenderer's texture coordinate buffer.
+    void Get3DRectangle( std::vector<double>& windingOrder ) const;
```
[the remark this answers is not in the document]
sgpearse: John and I talked about Get3DRectangle() and GetWindingOrder() on Friday and we decided to these functions from the VDC library.  These values are now in the SliceRenderer.  I think there is going to be more discussion on this matter going forward.

 PS - This function takes a non-const parameter, which implies that it changes.  Isn't it more efficient to take a value by reference when it's getting modified?  This is our practice with strings.  Is it not so with vectors?

include/vapor/ArbitrarilyOrientedRegularGrid.h
```diff
@@ -0,0 +1,114 @@
...
+    virtual void GetUserCoordinates (const size_t indices[], double coords[]) = delete;
+    virtual void GetUserCoordinates (const std::vector< size_t > &indices, std::vector< double > &coords) = delete;
+    virtual void GetUserCoordinates (size_t i, double &x, double &y, double &z) = delete;
+    virtual void GetUserCoordinates (size_t i, size_t j, double &x, double &y, double &z) = delete;
+    virtual void GetUserCoordinates (size_t i, size_t j, size_t k, double &x, double &y, double &z) = delete;
+
+    //! Retrieve the winding order that pertains to the generated array of data, if it is
+    //! to be drawn as a texture as is done in the SliceRenderer's vertex buffer.  The winding order is a series
+    //! of six 3D vertices that describe two triangles that define how the texture is drawn in space.
+    void GetWindingOrder( std::vector<double>& windingOrder ) const;
```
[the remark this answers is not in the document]
sgpearse: Just quoting my comment to what you noted above.  My same question applies:

> This function takes a non-const parameter, which implies that it changes. Isn't it more efficient to take a value by reference when it's getting modified? This is our practice with strings. Is it not so with vectors?

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: > Polygon winding orders that I'm familiar with are typically expressed as rotations about a normal: clockwise, or counter-clockwise (CCW), not as a list of points.

To whoever's reading this, John and I met about this last Friday.  I'm hoping I can absolve the confusion here.

The winding order does more than list the vertices of the texture according to the right-hand-rule (counter clockwise with reference to the normal).  

It lists the order of the vertices needed to draw the two triangles that compose the texture, used the [the vertex buffer](https://github.com/NCAR/VAPOR/blob/dd7ac4ee9d3aee8e508b6a40723e83c548ef9337/lib/render/SliceRenderer.cpp#L112).

@clyne - As we agreed I've moved the winding order into the Renderer library.  It now exists in SliceRenderer.cpp:211 (sorry I can't copy a perma-link for some reason).  I think this requires more discussion, because any other renderer that will need to draw a texture will need to duplicate this same ordering.

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: The ordering of the four corners of the texture has been moved from the VDC library to the SliceRenderer, at SliceRenderer.cpp:223.  I think this still may be a problematic solution, and requires more discussion.  However it does move renderer logic outside of the VDC library as we discussed.

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: @clyne - I've changed the logic to use the Grid's extents instead of the full domain extents and it's passed my tests.

clyne: Cool beans. The Doxygen documentation for the class needs to be refreshed for this change (it still discusses domainMin, domainMax).

lib/render/SliceRenderer.cpp
```diff
@@ -175,30 +176,58 @@ void SliceRenderer::_resetColormapCache()
...
+    ArbitrarilyOrientedRegularGrid* slice = new ArbitrarilyOrientedRegularGrid(
+        grid3d,
+        pd,
+        dims,
+        dataValues
+    );
+
+    CoordType corner1, corner2, corner3, corner4;
+    slice->GetUserCoordinates({0,0,0}, corner1);
+    slice->GetUserCoordinates({0,_textureSideSize,0}, corner2);
```
clyne, line 207: Shouldn't all of these be indexed by `_textureSideSize-1`, not `_textureSizeSize`? I think the only reason that it is working as that this getter will clamp the indices to be in range.

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: I missed that one, and an additional reference above.  Thanks for the catch.

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: Updated by deleting functions that reference the CoordType datatype.

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: That makes sense.  This last issue is in progress.  It's at the top of my list and should be submitted tomorrow morning barring anything unforeseen.

sgpearse: @clyne - I'm finding problems implementing this and I'm not sure it can be done without some kind of Grid::SetBlks() function.  

The fundamental thing we're trying to do is to allocate memory (blks) in ArbitrarilyOrientedRegularGrid's initialization list.

I've tried using raw and smart pointers to allocate the memory within the RegularGrid's constructor argument list.  

**Problem with smart pointers:** A smart pointer that's created in an argument list will delete the allocated data as soon as it falls out of scope, which is at the end of the RegularGrid's constructor.

```
ArbitrarilyOrientedRegularGrid::ArbitrarilyOrientedRegularGrid(
    const VAPoR::Grid *grid3d,
    planeDescription& pd,
    const DimsType& dims
) : RegularGrid(
        dims,
        {{pd.sideSize, pd.sideSize, 1}},
        {(std::unique_ptr<float>(new float[dims[0]*dims[1]*dims[2]])).get()},    // memory gets deleted after initialization list finishes
        {{0.,0.,0.}},
        {{1.,1.,1.}}
) {...}
```

**Problem with raw pointers:** Allocating memory with a raw pointer as a constructor argument will assign no lvalue that can be deleted later, so we get a memory leak.

```
ArbitrarilyOrientedRegularGrid::ArbitrarilyOrientedRegularGrid(
    const VAPoR::Grid *grid3d,
    planeDescription& pd,
    const DimsType& dims
) : RegularGrid(
        dims,
        {{pd.sideSize, pd.sideSize, 1}},
        {new float[dims[0]*dims[1]*dims[2]]},    // How do we delete this in the destructor?
        {{0.,0.,0.}},
        {{1.,1.,1.}}
) {...}
```

I've tried using private smart pointers in ArbitrarilyOrientedRegularGrid so they don't lose scope until the destructor is called (in the case of smart pointers).  This does not work because I can't allocate their memory before the RegularGrid constructor is called, which is where we need the memory.  Private raw pointers that we can manually delete will not work for the same reason - we can't allocate them before RegularGrid's constructor.

I feel like I'm running out of ideas aside from implementing a Grid::SetBlks() method, or Grid::initialize().

lib/render/SliceRenderer.cpp
sgpearse: You're correct.  Fixed.

include/vapor/ArbitrarilyOrientedRegularGrid.h
StasJ: Take a look at `PDoubleInput.cpp +6`

sgpearse: @stasj - thanks, this does work!

Followup question: I cannot get this to work with smart pointers without std::make_unique, introduced in c++14. 
 Is this the reason PDoubleInput uses raw pointers instead of smart pointers?

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: @clyne - I'm not sure why this code is missing the (outdated) label.  You can find the changes on the [current commit hash of this branch](https://github.com/NCAR/VAPOR/pull/2974/commits/7bb2296c47ce030971e08910ed522cfb931c78aa).

include/vapor/ArbitrarilyOrientedRegularGrid.h
sgpearse: Note: the (outdated) label is missing here too.  Now sure why.  Current changes can be seen [here](https://github.com/NCAR/VAPOR/pull/2974/commits/7bb2296c47ce030971e08910ed522cfb931c78aa).

lib/vdc/ArbitrarilyOrientedRegularGrid.cpp
```diff
@@ -0,0 +1,228 @@
...
+using namespace std;
+using namespace VAPoR;
+
+// clang-format off
+
+ArbitrarilyOrientedRegularGrid::ArbitrarilyOrientedRegularGrid(
+    const VAPoR::Grid *grid3d,
+    planeDescription& pd,
+    const DimsType& dims,
+    std::shared_ptr<float>& blks
```
clyne, line 20: Isn't this argument now obsolete? That was the entire point of having ArbitrarilyOrientedRegularGrid manage the space.

lib/vdc/ArbitrarilyOrientedRegularGrid.cpp
```diff
@@ -0,0 +1,228 @@
...
+    _findIntercepts(gridMin, gridMax, vertices);
+
+    // Find the minimum-area-rectangle that encloses the vertices that are along the edges
+    // of the Box.  Define this rectangle in 3D space, and in a newly projecteed 2D space.
+    std::vector<glm::tvec3<double, glm::highp>> tmpRectangle3D;
+    _getMinimumAreaRectangle( vertices );
+
+    // Pick sample points along our 2D rectangle, and project those points back into 3D space
+    // to query our 3D grid for data values.
+    _populateData( grid3d, pd, blks );
```
clyne, line 61: This should be `_myBlks`, not `blks` shouldn't it?

clyne (commented):
Just one more change (See inline) and I think we're there!

lib/vdc/ArbitrarilyOrientedRegularGrid.cpp
sgpearse: Yep.  This change was on a separate branch but not this one.  Updated and thanks for the catch.

lib/vdc/ArbitrarilyOrientedRegularGrid.cpp
sgpearse: Yep.  This change was on a separate branch but not this one.  Updated and thanks for the catch.

clyne (approved):
Looks great, @sgpearse! Nice work.

clyne:
@StasJ as soon as you approve this it can be merged and ready to go for the contour slicing work. Just a reminder to you and @sgpearse that we discussed some tweaks to the GUI for specifying the plane orientation and position. Once we've had a chance to explore that, we may need to change the slice rendering GUI.

include/vapor/ArbitrarilyOrientedRegularGrid.h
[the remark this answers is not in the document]
StasJ: In this case, it would not be more efficient to pass by reference. In the case of strings it is also not more efficient. The issue with mutable pass by reference is you have no way of knowing from syntax that the passed in value is an output in the calling code.

include/vapor/ArbitrarilyOrientedRegularGrid.h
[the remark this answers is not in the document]
StasJ: Answered above

StasJ (approved):

PR #2981, opened 2022-01-27
Fix broken sliders on Monterey

### LINKED ISSUE #2941: Sliders in FlowRendere broken
### Describe the bug
Sliders in FlowRenderer seem broken, as the screen recording shows.
Broken way 1: adjusting one slider will cause all 3 sliders to adjust too.
Broken way 2: a "range slider" as shown in rake region panel no longer shows a range. 

Note 1. Sliders in other renderers might be broken too.
Note 2. The same behavior is observed from both installer and source build.
Note 3. The test system is macOS Monterey, so it might be the new OS that's causing trouble.

### How to Reproduce
1. Open VAPOR, load VDC data: `/glade/p/cisl/vast/vapor/Bugs/2936/mydata.vdc`
2. Create a Flow Renderer. Select bx, by, bz as the advection vector field. Leave other settings with default values.
3. Enable Flow Renderer.
4. Go to `Seedings` tab and try to adjust values represented by sliders. The broken behavior is recorded below. 

https://user-images.githubusercontent.com/814885/146875757-9c8dc9fa-f526-4fa4-998d-338ae418d3e6.mov

sgpearse:
Fix #2941: broken sliders on Monterey by applying a sytleSheet to QSliders via a derived class (QMontereySlider) on OSX builds.  One caveat is that the Plotting tool's sliders will use this styleSheet because it's configured as a .ui file in QtCreator, not with our code-based widgets.

This is a Qt bug reported here: https://bugreports.qt.io/browse/QTBUG-98093

The bug report labels this issue as "closed" because it was fixed in Qt 6.2.3, but not in any version of Qt5.  The bug report also lists Qt 5.15 as an affected version, which is a commercial one.  This leaves me with some doubt whether it will be fixed in an open source version of Qt.

### DIFF
--- a/apps/vaporgui/QRangeSlider.cpp
+++ b/apps/vaporgui/QRangeSlider.cpp
@@ -1,4 +1,3 @@
-#include "QMontereySlider.h"
 #include "QRangeSlider.h"
 #include <QStylePainter>
 #include <QStyleOptionSlider>
@@ -24,7 +23,7 @@ public:
 
 QRangeSlider::QRangeSlider() : QRangeSlider(Qt::Orientation::Horizontal) {}
 
-QRangeSlider::QRangeSlider(Qt::Orientation orientation) : QMontereySlider(orientation)
+QRangeSlider::QRangeSlider(Qt::Orientation orientation) : QSlider(orientation)
 {
     _position[0] = 0;
     _value[0] = 0;
@@ -32,10 +31,27 @@ QRangeSlider::QRangeSlider(Qt::Orientation orientation) : QMontereySlider(orient
     _value[1] = QT_STOPS - 1;
     this->setRange(0, QT_STOPS);
     this->setTracking(true);
-    //this->QMontereySlider::setStyle(new QForceAbsoluteSetButtonsEnabledStyle(style()));
+    this->QSlider::setStyle(new QForceAbsoluteSetButtonsEnabledStyle(style()));
+
+#ifdef Darwin
+    this->setStyleSheet("\
+        QSlider::groove:horizontal {\
+            background: qlineargradient(x1:0, y1:0, x2:0, y2:1, stop:0 #B1B1B1, stop:1 #c4c4c4);\
+			height: 8px; \
+            margin: 2px 0;\
+		}\
+		QSlider::handle:horizontal {\
+            background: qlineargradient(x1:0, y1:0, x2:1, y2:1, stop:0 #b4b4b4, stop:1 #8f8f8f);\
+            border: 1px solid #5c5c5c;\
+            width: 18px;\
+            margin: -2px 0; \
+            border-radius: 3px;\
+		}"
+	);
+#endif
 }
 
-QSize QRangeSlider::minimumSizeHint() const { return QMontereySlider::minimumSizeHint(); }
+QSize QRangeSlider::minimumSizeHint() const { return QSlider::minimumSizeHint(); }
 
 void QRangeSlider::SetValue(float min, float max)
 {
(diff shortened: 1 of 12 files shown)


clyne:
Nice work running down a workaround for this, @sgpearse! Does it address all our GUI components as far as you know?

sgpearse:
> Nice work running down a workaround for this, @sgpearse! Does it address all our GUI components as far as you know?

Thanks and yes.  Classes that instantiate or derive from QSlider are macro'd to swap it with a QMontereySlider when building on OSX.  2 things to note:

1)  There are a few instances that are still being generated from .ui files and these needed to be "promoted" to QMontereySliders in Qt Creator parlance.  I wanted to isolate these changes to OSX but these changes will apply across all platforms unless there's some way to configure it in Creator.  This only affects the Plot utility.

2) The QRangeSlider that we use for our Box controller in the Seeding and Geometry tabs did not get fixed by inheriting QMontereySlider.  Manually applying the new stylesheet did fix the problem though.

apps/vaporgui/QRangeSlider.cpp
```diff
@@ -32,6 +32,25 @@ QRangeSlider::QRangeSlider(Qt::Orientation orientation) : QSlider(orientation)
     this->setRange(0, QT_STOPS);
     this->setTracking(true);
     this->QSlider::setStyle(new QForceAbsoluteSetButtonsEnabledStyle(style()));
+
+// Fix for Qt bug https://bugreports.qt.io/browse/QTBUG-98093
+// Apply a style sheet to QSlider to make it work on OSX Monterey
+// Note: Inheriting from QMontereySlider does not work, so we manually apply the styleSheet
+#ifdef Darwin
+    this->setStyleSheet("\
```
StasJ, line 40: Does this replace the functionality of the above setStyle?

sgpearse: If you are referring to the style changes when a handle is clicked, then yes.  The handle that is selected does not change its style.  But it functions just the same.

Would it help if I did a screen recording of what it looks like?

StasJ: The above setStyle changes the functionality of the sliders. It allows you to drag the bar rather than the controls.

sgpearse: That's always been a feature.  Do you mean something else?

StasJ: I don't quite understand, sorry

shaomeng:
Scott, it seems a clever fix, thanks for locating it! 

However, I seem to have a build error on this branch. Is it something you worked on recently? 
```
[ 25%] Building CXX object lib/vdc/CMakeFiles/vdc.dir/SliceGridAlongPlane.cpp.o
/Users/samuel/Git/VAPOR-git/lib/vdc/SliceGridAlongPlane.cpp:5:10: fatal error: 'GTE/MinimumAreaBox2.h' file not found
#include <GTE/MinimumAreaBox2.h>
         ^~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [lib/vdc/CMakeFiles/vdc.dir/SliceGridAlongPlane.cpp.o] Error 1
make[1]: *** [lib/vdc/CMakeFiles/vdc.dir/all] Error 2
make: *** [all] Error 2
```

apps/vaporgui/QRangeSlider.cpp
sgpearse: No problem - In the GIF below I'm dragging the bar with the sliders on this branch, and then I'm dragging the bar on the sliders on main.  In both cases, we can drag the bar so the functionality is unchanged.  It just looks different.

rangeSlider

StasJ: Ok great, that's what I was asking but I didn't quite understand your last reply. Thanks!

shaomeng (approved):
This is an awesome fix, thanks for bringing it into attention!

clyne:
Still need your approval on this @StasJ

StasJ (approved):

Issue #2989, opened 2022-02-02
Slice UGRID rotation broken


StasJ:
1. Open `/glade/p/cisl/vast/vapor/data/Source/UGRID/NOAA-geoflow/small`
2. Create/enable the slice renderer
3. Chang the Y rotation to 0<y<90

Notice that any value in that range will result in a 90 degree rotation.

sgpearse:
I know what you mean - This happens when we stretch the Z axis, and it becomes more noticeable the more we stretch it.  In this dataset, we stretch by nearly 4million.

It's technically the correct behavior but there may be something we can do to make it more user friendly by taking the stretch factors into account with the angles of rotation.

clyne:
I think this begs the question of which coordinate system the rotation of the plane should be applied too: before or after the scaling. I'm not sure that there is a right answer.

clyne:
Changing this to "Usability" since the behavior appears correct

PR #2992, opened 2022-02-06
3D Contours

### LINKED ISSUE #2728: Contours: Allow for rendering 3D variables

StasJ:
Adds contouring slices. Adds "offset" to slice positioning scheme. The GUI automatically calculates the possible extents of the offset along the axis perpendicular	to the slice and sets the slider range accordingly.

Fixed #2728

### DIFF
--- a/apps/vaporgui/ContourEventRouter.cpp
+++ b/apps/vaporgui/ContourEventRouter.cpp
@@ -42,7 +42,6 @@ ContourEventRouter::ContourEventRouter(QWidget *parent, ControlExec *ce) : Rende
     }));
     
     AddGeometrySubtab(new PGroup({
-        new PGeometrySubtab,
         (new PShowIf(""))->DimensionEquals(3)->Then(new PGroup({
             (new PSection("Slice Orientation", {
                 new PEnumDropdown(RenderParams::SlicePlaneOrientationModeTag, {"Rotation", "Normal"}, {(int)RenderParams::SlicePlaneOrientationMode::Rotation, (int)RenderParams::SlicePlaneOrientationMode::Normal}, "Orientation Mode"),
@@ -60,6 +59,7 @@ ContourEventRouter::ContourEventRouter(QWidget *parent, ControlExec *ce) : Rende
             (new PSliceOriginSelector)->SetTooltip("The slice plane will pass through this point. The plane can be offset from this point along the plane normal determined by the orientation."),
             (new PSliceOffsetSelector)->SetTooltip("Offset the plane from its origin along its normal (set by the orientation)."),
         })),
+        new PGeometrySubtab,
     }));
     
     AddAnnotationSubtab(new PAnnotationColorbarWidget);
--- a/apps/vaporgui/ContourEventRouter.cpp
+++ b/apps/vaporgui/ContourEventRouter.cpp
@@ -3,6 +3,7 @@
 #include "PWidgets.h"
 #include "PSliderEditHLI.h"
 #include "PConstantColorWidget.h"
+#include "PSliceOriginSelector.h"
 
 using namespace VAPoR;
 typedef ContourParams CP;
@@ -15,6 +16,7 @@ ContourEventRouter::ContourEventRouter(QWidget *parent, ControlExec *ce) : Rende
 
     AddVariablesSubtab(new PGroup({
         new PSection("Variable Selection", {
+            new PDimensionSelector,
             new PScalarVariableSelector,
             new PHeightVariableSelector,
         }),
@@ -32,9 +34,26 @@ ContourEventRouter::ContourEventRouter(QWidget *parent, ControlExec *ce) : Rende
             _minValueSlider = new PDoubleSliderEditHLI<CP>("Minimum Value", &CP::GetContourMin, &CP::SetContourMin),
             new PConstantColorWidget
         }),
+        (new PShowIf(""))->DimensionEquals(3)->Then(new PGroup({
+            new PSection("Slice", {
+                (new PDoubleSliderEdit(RenderParams::SampleRateTag, "N Samples"))->SetRange(32, 2000)
+            })
+        })),
     }));
     
-    AddGeometrySubtab(new PGeometrySubtab);
+    AddGeometrySubtab(new PGroup({
+        new PGeometrySubtab,
+        (new PShowIf(""))->DimensionEquals(3)->Then(new PGroup({
+            new PSection("Slice Rotation", {
+                (new PDoubleSliderEdit( RenderParams::XSlicePlaneRotationTag, "X"))->SetRange(-90.,90.)->EnableDynamicUpdate(),
+                (new PDoubleSliderEdit( RenderParams::YSlicePlaneRotationTag, "Y"))->SetRange(-90.,90.)->EnableDynamicUpdate(),
+                (new PDoubleSliderEdit( RenderParams::ZSlicePlaneRotationTag, "Z"))->SetRange(-90.,90.)->EnableDynamicUpdate(),
+            }),
+            new PSliceOriginSelector,
+            new PSliceOffsetSelector,
+        })),
+    }));
+    
     AddAnnotationSubtab(new PAnnotationColorbarWidget);
 
     // clang-format on
--- a/apps/vaporgui/ContourEventRouter.cpp
+++ b/apps/vaporgui/ContourEventRouter.cpp
@@ -44,13 +44,21 @@ ContourEventRouter::ContourEventRouter(QWidget *parent, ControlExec *ce) : Rende
     AddGeometrySubtab(new PGroup({
         new PGeometrySubtab,
         (new PShowIf(""))->DimensionEquals(3)->Then(new PGroup({
-            new PSection("Slice Rotation", {
-                (new PDoubleSliderEdit( RenderParams::XSlicePlaneRotationTag, "X"))->SetRange(-90.,90.)->EnableDynamicUpdate(),
-                (new PDoubleSliderEdit( RenderParams::YSlicePlaneRotationTag, "Y"))->SetRange(-90.,90.)->EnableDynamicUpdate(),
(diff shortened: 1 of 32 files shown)


apps/vaporgui/ContourEventRouter.cpp
```diff
@@ -32,9 +34,26 @@ ContourEventRouter::ContourEventRouter(QWidget *parent, ControlExec *ce) : Rende
...
+    
+    AddGeometrySubtab(new PGroup({
+        new PGeometrySubtab,
+        (new PShowIf(""))->DimensionEquals(3)->Then(new PGroup({
+            new PSection("Slice Rotation", {
+                (new PDoubleSliderEdit( RenderParams::XSlicePlaneRotationTag, "X"))->SetRange(-90.,90.)->EnableDynamicUpdate(),
+                (new PDoubleSliderEdit( RenderParams::YSlicePlaneRotationTag, "Y"))->SetRange(-90.,90.)->EnableDynamicUpdate(),
+                (new PDoubleSliderEdit( RenderParams::ZSlicePlaneRotationTag, "Z"))->SetRange(-90.,90.)->EnableDynamicUpdate(),
+            }),
+            new PSliceOriginSelector,
```
sgpearse, line 52: The origin controls seem to be no-ops.  Should they be removed?

lib/render/ContourRenderer.cpp
```diff
@@ -111,10 +145,57 @@ int ContourRenderer::_buildCache()
...
+        grid2d->GetUserCoordinates({0, pd.sideSize - 1, 0}, corner2);
+        grid2d->GetUserCoordinates({pd.sideSize - 1, 0, 0}, corner3);
+        grid2d->GetUserCoordinates({pd.sideSize - 1, pd.sideSize - 1, 0}, corner4);
+
+        _sliceQuad = {
+            glm::vec3(corner1[0], corner1[1], corner1[2]),
+            glm::vec3(corner3[0], corner3[1], corner3[2]),
+            glm::vec3(corner4[0], corner4[1], corner4[2]),
+            glm::vec3(corner2[0], corner2[1], corner2[2]),
+        };
```
sgpearse, line 190: The ordering for these vertices is also defined in SliceRenderer.cpp:217.  Are we sure this ordering (corner1, 3, 4, then 2) shouldn't be kept in one place, the ArbitrarilyOrientedRegularGrid?

lib/vdc/ArbitrarilyOrientedRegularGrid.cpp
```diff
@@ -224,3 +224,31 @@ void ArbitrarilyOrientedRegularGrid::_populateData(
...
+    glm::vec3 o(pd.origin[0], pd.origin[1], pd.origin[2]);
+    float     t0, t1;
+    glm::vec3 tMin = (dataMin - o) / n;
+    glm::vec3 tMax = (dataMax - o) / n;
+    glm::vec3 bt1 = min(tMin, tMax);
+    glm::vec3 bt2 = max(tMin, tMax);
+    t0 = max(max(bt1.x, bt1.y), bt1.z);
+    t1 = min(min(bt2.x, bt2.y), bt2.z);
+    return std::pair<float, float>(t0, t1);
+}
```
sgpearse, line 254: I think there may be a miscalculation for the range and offset.  

Case 1)
1) Load a katrina file, then create and enable a Contour renderer
2) Rotate by 20 degrees on X
3) Set the offset to its maximum value.  The contour will just barely move away from its original location.

Case 2)
With the dataset below, do the following:
1) Create and enable a contour renderer
2) Rotate it by 20 degrees on X
3) Move the offset to its minimum or maximum value.  At these values the contour should only intersect at an edge of the domain, which would render nothing.  Note in the screenshot below that the green rectangle is well within the domain, not intersecting the edge.
[256i.zip](https://github.com/NCAR/VAPOR/files/8017528/256i.zip)
Screen Shot 2022-02-07 at 10 41 51 AM

sgpearse (commented):
Thanks for doing this Stas.  This PR is much simpler than I thought it would be.  Just a few in-line questions and there may be a bug related to the offset calculation.

clyne:
@StasJ we had discussed making some GUI changes in the discussions in #2728. One of these was to allow the specification of the orientation of the slice with a plane normal. Was this implemented? 

Relatedly, I'm not clear about what the "slice origin" does and if it is needed. It seems like the orientation (via normal or rotations about the coordinate system axes) plus a plane offset should be sufficient. Can you explain? Thanks!

apps/vaporgui/ContourEventRouter.cpp
StasJ: I don't see how they are no-ops?

lib/render/ContourRenderer.cpp
StasJ: I believe this code used to be in the grid class but it was moved to the slice renderer in the PR review? They are based on the grid being CCW which should always be consistent anyway.

lib/vdc/ArbitrarilyOrientedRegularGrid.cpp
StasJ: Case 1 is correct. See #2989

Case 2 is correct. I think you meant 45 degrees, not 20.

StasJ:
> @StasJ we had discussed making some GUI changes in the discussions in #2728. One of these was to allow the specification of the orientation of the slice with a plane normal. Was this implemented?
> 
> Relatedly, I'm not clear about what the "slice origin" does and if it is needed. It seems like the orientation (via normal or rotations about the coordinate system axes) plus a plane offset should be sufficient. Can you explain? Thanks!

I added the ability to place the slice at a displacement along the plane normal. I did not realize that specifying the orientation using the normal was a requirement in itself. I will implement this.

The slice origin is the point at which the displacement is zero.

clyne:
> > @StasJ we had discussed making some GUI changes in the discussions in #2728. One of these was to allow the specification of the orientation of the slice with a plane normal. Was this implemented?
> > Relatedly, I'm not clear about what the "slice origin" does and if it is needed. It seems like the orientation (via normal or rotations about the coordinate system axes) plus a plane offset should be sufficient. Can you explain? Thanks!
> 
> I added the ability to place the slice at a displacement along the plane normal. I did not realize that specifying the orientation using the normal was a requirement in itself. I will implement this.

Cool. Sorry if it wasn't clear. It may very well be that a rotation about the coordinate axes is preferred in some use cases. So one option is to offer both (e.g. toggle between one mode and the other).
> 
> The slice origin is the point at which the displacement is zero.

Do you think this is needed/helpful? Could we simply fix the origin to the center or min extent of the domain?

StasJ:
> > The slice origin is the point at which the displacement is zero.
> 
> Do you think this is needed/helpful? Could we simply fix the origin to the center or min extent of the domain?

For example, if we were to remove it, if a user wanted to place a slice at a specific location L, they would have to look up what the center of the domain is, O, then compute the normal from O to L, the distance between O and L, and then input those values rather than just the location.

clyne:
> > > The slice origin is the point at which the displacement is zero.
> > 
> > 
> > Do you think this is needed/helpful? Could we simply fix the origin to the center or min extent of the domain?
> 
> For example, if we were to remove it, if a user wanted to place a slice at a specific location L, they would have to look up what the center of the domain is, O, then compute the normal from O to L, the distance between O and L, and then input those values rather than just the location.

OK. That use case makes sense. However, I think there may still be some things we can do to help the usability.

1. Add tool tips that clearly explain what the origin, orientation, and offset are doing and how they relate to each other.
2. There is a label under origin that states "slice origin is shown as a yellow crosshair". I don't see a cross hair, and I think it be pretty helpful to have one. Manipulating the "origin" knobs often have no visible effect. Having a crosshair would provide a visual cue as to what is going on.

apps/vaporgui/ContourEventRouter.cpp
clyne: They're not no-ops, but they can appear that way in some instances. @sgpearse see my comments [below](https://github.com/NCAR/VAPOR/pull/2992#issuecomment-1032865304).

lib/render/ContourRenderer.cpp
clyne: @sgpearse the Grid class uses CCW ordering. So I'm not sure what you are referring to when you say 'this info should be retrieved to prevent duplication"

sgpearse: My bad - I was confusing this with something else.

clyne:
@StasJ the UI changes really seem to help with the usability IMHO. Nice work. One more minor item: with the contour renderer, in particular, often nothing is drawn because the specified contour lines are outside of the range of the data in the plane. Currently a green bounding box is drawn around the plane when you are manipulating it (e.g. mouse down on offset), but when you release the button the green bounding box is no longer drawn, which may result in nothing being visible. Can we make the green bounding box always visible when the Geometry tab is active, rather than only visible when  it is being manipulated?

sgpearse:
@clyne 

> with the contour renderer, in particular, often nothing is drawn because the specified contour lines are outside of the range of the data in the plane.

IMO, this is a good argument for the histogram using the data within the plane, not the volume.

sgpearse (approved):
Thanks @stasj.  Approved.

clyne:
@sgpearse re my comments on the manipulator UI: you've approved this PR. I assume that means you are satisfied with the usability and are going to make the slice renderer manipulators consistent? Just want to confirm. Thanks.

clyne:
@stas, unfortunately once again clang-format has failed. Can you please resolve so I can merge this? Thanks

StasJ:
> @stas, unfortunately once again clang-format has failed. Can you please resolve so I can merge this? Thanks

@clyne I don't know what happened since the clang-format hook ran before the push but I manually re-ran clang-format on the bad files and that fixed it.

clyne:
> 

@StasJ, I'm not sure if it is on the VAPOR repo or Project Pythia, but I've noticed I sometimes have to run 'git commit' twice in a row to get all of the formatting done. Weird.

PR #3000, opened 2022-02-10
Merge contour3d with main

sgpearse:
@stasj and @clyne - rather than add a review comment on #2992 asking for the following, I think it may be easier/faster to submit a PR with the following changes.

This PR...
1) Merges contour3d with main and resolves conflicts
2) Adds documentation to the new RenderParams functions GetSlicePlaneOrigin(), GetSlicePlaneNormal(), and GetSlicePlaneRotation()
3) Bumps the contour renderer's rotation/orientation/offset controls to the top of the geometry tab.  These are the primary controls the user will be working with.  IMO they should be at the top rather than things like the transform table and copy region from renderer sections.


sgpearse:
Ok, I'll close this and leave a note on #2992.

PR #3005, opened 2022-02-11
Fix win2

sgpearse:


clyne (approved):

sgpearse:
@clyne I was waiting to see if checks passed before requesting review, and it looks like they're good to go.  Feel free to merge and I should be able to get installers up.

PR #3028, opened 2022-02-15
Fix issue 3015 - slice renderer missing upper portions of dataset

### LINKED ISSUE #3015: Slice renderer misses data at top of domain with VDC
The slice renderer (and contour renderer) fail to capture data in VDC files near the top of the domain.

To reproduce:
1. Load the VDC file /glade/p/cisl/vast/vapor/data/VDC/vapor-3.5.0/CAM/AdamPhillips/F2000C5_new120_025/F2000C5_new120_025-3.5.0.nc
2. Create a slice renderer and change the variable fidelity to high
3. Move the component of the slice origin to anything greater than ~900. Nothing is displayed (there are several layers over 900, and the highest is 992).

Note that the volume render is able to capture this data as can be verified by again setting the fidelity to high, and moving the minimum Z extent to a value over 900.

N.B. the correct path to use should be glade/p/cisl/vast/vapor/data/VDC/vapor-3.6.0, not glade/p/cisl/vast/vapor/data/VDC/vapor-3.5.0

The attached image shows a volume rendering with the min Z extent 950, and max 992

FV_dvr

sgpearse:
Fix #3015.

The problem was that the slice was being placed outside of the box's extents.  

At startup, a VDC will have Box classes that encompass the entire domain.  When increasing refinement, the domain will grow but the Box will not grow proportionally.

This fix adjusts the Box extents to be whatever fraction of the domain it encompassed before a refinement level change.    IE - If a Box spans between the 25% and 75% extents on an axis, it will still encompass those percentiles after refinement level changes.

### DIFF
--- a/lib/params/RenderParams.cpp
+++ b/lib/params/RenderParams.cpp
@@ -308,8 +308,32 @@ void RenderParams::SetCompressionLevel(int level) { SetValueLong(_CompressionLev
 
 void RenderParams::SetRefinementLevel(int level)
 {
+    size_t oldRefLevel = GetRefinementLevel();
+
     if (level < 0) level = 0;
     SetValueLong(_RefinementLevelTag, "Set refinement level", level);
+
+    // Adjust the Box's extents so that it encloses the same percentage of space in the new domain as it did in the old domain
+    Box* b = GetBox();
+    if (b == nullptr) return;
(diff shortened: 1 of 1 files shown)


clyne:
As discussed. This is not the fix we want to implement for this problem.

Issue #3029, opened 2022-02-15
Make GUI for slice renderer and contour renderer consistent


clyne:
The Slice and Contour renderers currently have inconsistent GUIs for manipulating the slice plane

It should be noted that the contour UI is simply an extension to the slice UI; both provide specification of orientation through rotation about the coordinate axis, and both all the specification of position with by selecting a point in the domain that the plane passes through. The contour UI adds two additional features: 1. The option to specify the orientation with a normal to the plane, and 2. the ability to specify the position of the plane as an offset along the normal from a point in space.

If the user knows exactly where they want the plane, and how they want it oriented, both UIs operate in the same way and work well. However, in the case that the user only knows apriori how they want to orient the plane, but not where they want to position it, I find the offset + position a *much* simpler interface than what is possible with the slice UI. Even in the simplest case of an axis-aligned plane, it is not at all obvious which of the three position sliders to use in the slice interface to properly move the plane. If the plane is not axis aligned, it is very difficult to sweep it through space along a normal with the slice UI.

clyne:
Input from @sgpearse 

To optimize Vapor's usability, options that serve the most common use cases should be made prevalent.  The common use cases should require little effort as possible for the user to achieve a desired result.  Less common use cases should require additional exploration of the GUI to achieve results.

We can currently position our slice/contour planes in two ways; through the origin and through the offset controls.  These two approaches serve the three use cases listed below in order of their simplicity and commonality.

Use case 1- The simplest and most common.  Requires 4 inputs (XYZ rotation + offset).
1) Rotate
2) Explore cross sections via offset parameter

Use case 2 - More complex and less common.  Needed when a user knows an exact location that the slice/contour must intersect.  Requires 6 inputs (XYZ rotation + XYZ origin).
1) Rotate
2) Set precise XYZ origin parameters

Use case 3 - Most complex and least common.  Needed when a user needs multiple slices that are offset from a point of origin.  Requires 7 inputs (XYZ rotation + XYZ origin + offset)
1) Rotate
2) Set XYZ origin parameters
3) Duplicate slice
4) Offset new slice

By presenting offset and origin controls simultaneously, we present seven parameters for our users to understand, in order to satisfy use case 3.  This is a lot, and I think we need to be sensitive to this.

The biggest challenge I have with using tools such as Vapor is that it takes too much time to produce a nice visualization (countless options to fine tune, difficult to script etc.).
-Falko Judt

[...]

PR #3033, opened 2022-02-16
#3029 Add a show/hide option for origin specification

StasJ:
#3029

ss


sgpearse (changes requested):
@stasj Now that the origin controls are optional, they should be pushed further down in the Geometry tab.  IE - the order should be Rotation, Offset, Origin.

For convenience sake, I made the change in PR #3034 if you'd like to merge it to this.

clyne (approved):
Thanks for doing this, @StasJ. I definitely think it helps with clutter/confusion.

PR #3035, opened 2022-02-16
3029 - Adopt new control scheme in SliceRenderer

sgpearse:
This PR introduces PSliceController, a widget that can be used by the both Slice and Contour renderers to position their planes.


clyne:
@sgpearse please following conventions when naming branches for PRs:

I.e. `issue_<number>`

I'm not able to checkout your branch, presumably because git doesn't like branches named with a number:

cisl-benona[240] git checkout 3029
fatal: Cannot switch branch to a non-commit '3029'

sgpearse:
@clyne - Reopened under #3037

PR #3038, opened 2022-02-16
Fix issue 3031 - Misplaced slice origin crosshair

### LINKED ISSUE #3031: Origin cursor for slice and contour manipulator displayed in wrong location
The "yellow crosshair" that is displayed by the Slice and contour origin manipulator is positioned incorrectly with some (presumably scaled) data. 

To reproduce:
1. Load Data/CAM/AdamPhillips/F2000C5_new120_025/FV_768x1152.bilinear.F2000C5_new120_025.cam.h0.2004-01.nc
2. Enable the slice renderer
3. Change to the geometry tab so that the cross hair is displayed
4. If the scene is rotated it can clearly be seen that the cross hair is displaced above the slice plane. The cross hair center should be on the plane

sgpearse:
Fixes #3031 by removing superfluous scaling logic.


StasJ (changes requested):
1. Follow the steps in #3031
2. Enable slice origin controls
3. Change the origin Z

The crosshair is not in the correct position

s

PR #3039, opened 2022-02-17
Contour plane bounding box not positioned correctly #3032

### LINKED ISSUE #3032: Contour plane bounding box not positioned correctly
The green rectangle that used to show the position of the Contour (and Slice) plane  is positioned incorrectly with some (presumably scaled) data.

To reproduce:

1. Load Data/CAM/AdamPhillips/F2000C5_new120_025/FV_768x1152.bilinear.F2000C5_new120_025.cam.h0.2004-01.nc
2. Enable the contour renderer with the "U" variable
3. Change to the geometry tab so that the green bounding rectangle is displayed
4. Move the "offset" slider (or Z component of the Origin slide) toward the max value. The visible contours will be moved, but the bounding rectangle remains in place.

StasJ:
Fix #3032

### DIFF
--- a/apps/vaporgui/VizWin.cpp
+++ b/apps/vaporgui/VizWin.cpp
@@ -944,6 +944,10 @@ void VizWin::_drawContourSliceQuad()
     ContourParams *cp = dynamic_cast<ContourParams *>(_getRenderParams());
     assert(cp);
     if (!cp) return;
+    
+    _glManager->matrixManager->PushMatrix();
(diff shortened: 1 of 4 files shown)


sgpearse (approved):
@StasJ Thank you for the fix.  I'll use this for #3031 once I can pinpoint a couple issues on my end.

clyne (approved):
Awesome. Thanks!

PR #3041, opened 2022-02-18
Add green quad to Slice Renderer's geometry mode

sgpearse:
For consistency with the ContourRenderer, this PR adds a green quad to Slice Renderer's geometry mode.

### DIFF
--- a/apps/vaporgui/VizWin.cpp
+++ b/apps/vaporgui/VizWin.cpp
@@ -667,7 +667,7 @@ void VizWin::_renderHelper(bool fast)
     if (_getCurrentMouseMode() == MouseModeParams::GetRegionModeName()) {
         updateManip();
         if (_getRenderParams() && _getRenderParams()->GetOrientable()) { _updateOriginGlyph(); }
-        if (dynamic_cast<ContourParams *>(_getRenderParams())) _drawContourSliceQuad();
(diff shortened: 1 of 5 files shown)


clyne (approved):

apps/vaporgui/VizWin.cpp
```diff
@@ -667,7 +667,7 @@ void VizWin::_renderHelper(bool fast)
     if (_getCurrentMouseMode() == MouseModeParams::GetRegionModeName()) {
         updateManip();
         if (_getRenderParams() && _getRenderParams()->GetOrientable()) { _updateOriginGlyph(); }
-        if (dynamic_cast<ContourParams *>(_getRenderParams())) _drawContourSliceQuad();
+        if (dynamic_cast<ContourParams *>(_getRenderParams()) || dynamic_cast<SliceParams *>(_getRenderParams())) _drawContourSliceQuad();
```
StasJ, line 670: You could just use isorientable now or just put it in the same if block.

sgpearse: Note - I wasn't sure what you meant by "the same if block" but I think you're saying to put the _drawContourSliceQuad() call in the same block as _updateOriginGlyph().  Updated.

PR #3045, opened 2022-02-22
Flow deltaT adjustment on thin volumes

### LINKED ISSUE #3008: Flow renderer doesn't work with UGRID data
The flow renderer does not seem to work with UGRID data. 

To reproduce:
1. Load all of the files in UGRID/NOAA-geoflow/small
2. Create a flow renderer and set the X,Y,Z component variables to v1, v2, v3
3. Enable the renderer

Nothing is displayed. Changing various flow parameters doesn't seem to help

shaomeng:
fix #3008

### DIFF
--- a/include/vapor/Advection.h
+++ b/include/vapor/Advection.h
@@ -106,6 +106,9 @@ private:
     // Returns the value after adjustment.
     float _applyPeriodic(float val, float min, float max) const;
 
+    // Print return code if it's non-zero and compiled in debug mode.
+    void _printNonZero(int rtn, const char *file, const char *func, int line) const;
+
     void        _calculateParticleIntegratedValue(Particle &p, const Particle &prev, const Field *scalarField, const bool skipNonZero, const float distScale,
                                                   const std::vector<double> &integrateWithinVolumeMin, const std::vector<double> &integrateWithinVolumeMax) const;
     static bool _isParticleInsideVolume(const Particle &p, const std::vector<double> &min, const std::vector<double> &max);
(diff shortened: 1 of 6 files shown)


clyne (commented):
@shaomeng something still isn't right here. If you set the Z component to 'none', the resulting stream lines are close to zero length. Increasing the integration count has no visible effect (other than the computation takes a lot longer). Perhaps if a velocity component is 'none' you should not be including the associated coordinate extent in the calculation of deltaT. E.g. If the Z component of velocity is set to 'none' then the Z extent of the domain should not be used in the calculation of deltaT.

StasJ (approved):

shaomeng:
> If the Z component of velocity is set to 'none' then the Z extent of the domain should not be used in the calculation of deltaT.

@clyne This review comment has been addressed in d320d1f00c7309054cbf87bdacc114a62a648924.

clyne:
Getting closer, but I still can't get the integration to proceed beyond a very short distance when "Z field" is none, can you? Can you explain why this is happening?

shaomeng:
> Getting closer, but I still can't get the integration to proceed beyond a very short distance when "Z field" is none, can you? Can you explain why this is happening?

@clyne The variables v1 and v2 have mostly tiny values centered around zero, resulting in many "sinks" or near sinks, which are areas with zero or near-zero values. When a particle enters a such area, it makes small movements in the sink but can be almost impossible to escape. With a domain size of millions, those small movements appear invisible. For this particular data set, it appears that there are more sinks than non-sinks, so most particles fall in sinks after traveling a while.

PR #3047, opened 2022-02-22
Failures converting data to VDC format #3046

### LINKED ISSUE #3046: Failures converting data to VDC format
A number of canonical data sets can not be converted to the VDC format with cfvdccreate/cf2vdc:

CAM/AdamPhillips/F2000C5_new120_025
POP/BryanASP/one_degree/daily
POP/BryanASP/one_degree/monthly
POP/BryanASP/tenth_degree/daily
POP/BryanASP/tenth_degree/monthly
ROMS/Kauffman

clyne:
Fixed #3046 

This change prohibits variables from being defined as both coordinate and data variables in the VDC. At present the VDC cannot support this because the underlying NetCDF API doesn't allow a variable to be defined twice.

### DIFF
--- a/apps/cfvdccreate/cfvdccreate.cpp
+++ b/apps/cfvdccreate/cfvdccreate.cpp
@@ -272,6 +272,8 @@ int main(int argc, char **argv)
         if (rc < 0) return (1);
 
         for (int i = 0; i < datanames.size(); i++) {
+            if (find(coordnames.begin(), coordnames.end(), datanames[i])!=coordnames.end()) continue;
+
             DC::DataVar dvar;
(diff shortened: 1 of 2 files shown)


StasJ (commented):
In the case that a variable is defined as both a coord and data var, this PR appears to only keep whichever is defined last. Wouldn’t this lead to situations where the dataset is missing data or worse missing coordinates? Should we prioritize keeping the coordinate data? We could also rename the data variable to say lat_data in this scenario.

clyne:
Yes, this is a problem. This issue has become a can of worms, and I'm not really sure what to do about it. I need to think some more. Converting this PR to a draft for now.

PR #3061, opened 2022-03-05
Fixed Crash when calculating 2D wind speed 

### LINKED ISSUE #3059: Crash when calculating 2D wind speed
To reproduce:
1) Load any dataset with 2D wind variables (katrina will work)
2) Create a wind speed variable like what's shown below.  The mismatched grids of the two variables do not matter.

```
from numpy import *
speed=sqrt(U10*U10+V10*V10)
```

Vapor will crash.

clyne:
Fixed #3059

PyEngine was still using vector<size_t> to represent grid indices.  Converted to DimsType

### DIFF
--- a/include/vapor/ConstantGrid.h
+++ b/include/vapor/ConstantGrid.h
@@ -50,7 +50,7 @@ private:
     // They do nothing and return meaningless values.
     // Do not use!
     //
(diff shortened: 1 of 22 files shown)


sgpearse (approved):
Looks good.  Thanks John.

shaomeng (approved):
Looking good, thanks John!

Issue #3050, opened 2022-02-23
Barb Renderer: Lighting is wrong


sgpearse:
The Barb Renderer sets the lighting direction along the +Z axis with a call to LegacyGL->LightDirection(lightDir), where lightDir={0,0,1,0}.  This means that in the following Map Orthographic screenshot, the light source is coming from behind the scene (we are looking down the -Z axis).  Therefore, we should not see the barb.

![Screen Shot 2022-02-23 at 11 05 33 AM](https://user-images.githubusercontent.com/9522770/155380407-6c18ac94-48df-480f-ad75-0960ab06eafc.png)

I think we should disable lighting for the barb renderer, as shown in the screenshot below.  The current lighting setup makes color-mapped barbs unrepresentative of their variable.  If we truly wanted to add lighting, we should probably support more than one light in LegacyGL.

![Screen Shot 2022-02-23 at 11 07 36 AM](https://user-images.githubusercontent.com/9522770/155380929-e7eccd58-df23-4aae-9f11-7a93936c967d.png)

To reproduce:
1) Load any dataset, create and enable a barb renderer
2) Since the light direction is coming from +Z and our default view comes from -Z, we shouldn't be seeing barbs until we orient the camera along the light direction.

PR #3067, opened 2022-03-16
Add lighting switch to Barb Renderer, tweak defaults, and promote widgets in the Appearance tab

sgpearse:
Addresses #3050 by adding a checkbox that enables/disables lighting in the Barb Renderer.  The default lighting is off so that the barbs are accurately colored by a given variable.

This PR also changes the default use of constantly-colored barbs from true to false.  IE color-mapping barbs by a variable is the default.  Reason 1) This is consistent with what we do in the Flow renderer.  Reason 2) I cannot recall seeing our barb renderer used with constant color in a visualization.  The use case for constant color certainly exists, but I don't think it's the right default.

It also promotes the barb-specific controls to the top of the Appearance tab instead of the bottom because I believe it better suits the workflow of the Barb Renderer.  Before coloring the barbs, 99% of the time users will need to change the barb orientation, location, thickness, and length.  IE - our default 10x10 array of barbs on the XY plane will almost always need to be reconfigured by the user, before they can adjust the colormap according to their needs.

Barbs with lighting
Screen Shot 2022-03-16 at 10 07 57 AM

Barbs without lighting
Screen Shot 2022-03-16 at 10 07 50 AM


StasJ (commented):
Are the image titles reversed? It seems you can’t see anything with lighting disabled. The solution to the original problem may be to have the light always point in the view direction so it has consistent lighting which is the case with the other renderers. 

As to moving the TF widget, while your argument makes sense and I am not opposed to it, we may want to consider whether it is more valuable to have the visualizer tab consistent among the renderers.

sgpearse:
@stasj and @clyne - The following changes are in:

1) The LightingEnabled tag has moved from BarbParams to RenderParams 
2) Light direction now follows the camera direction for the Barb renderer
3) Lighting is enabled by default

Note - I'm seeing unexpected behavior in the Barb renderer, in that I need to invert the light direction to get the right result.  I'm not sure if the barb normals are backwards or what.   I think we may want to consider refactoring the BarbRenderer to draw glyphs instead of mathematically drawing cylinders and arrowheads, which is hard to debug (see BarbRenderer.cpp:383).

clyne (approved):
Looks great, @sgpearse. Thanks for fixing this.

StasJ (approved):

PR #3069, opened 2022-03-16
Sliders to control Flow Rake center

sgpearse:
This adds dynamic sliders that allow the Flow Rake's center to be scanned along the XYZ axes.

The Flow Rake Center's extents are bound to the Flow Params' Box.
rakeCenter


clyne:
@sgpearse have you tried this with large data and lots of seed points? Won't that present a performance problem?

sgpearse:
@sgpearse Add dynamic update option to some sliders, but not all.  Disabled by default.

apps/vaporgui/FlowEventRouter.cpp
```diff
@@ -157,6 +167,25 @@ void FlowEventRouter::_updateTab()
     int numTS = GetActiveDataMgr()->GetNumTimeSteps();
     _pathlineLengthSlider->SetRange(0, std::max(1, numTS - 1));
     _pathlineInjectionSlider->SetRange(0, numTS);
+
+    std::vector<double> min, max;
+    FlowParams *        fp = dynamic_cast<FlowParams *>(GetActiveParams());
+    fp->GetBox()->GetExtents(min, max);
```
shaomeng, line 173: I think it's encouraged to use the API with a fixed-size array when querying box extents. Check out this one:
```
void GetExtents(VAPoR::CoordType &minExt, VAPoR::CoordType &maxExt) const;
```

lib/params/FlowParams.cpp
```diff
@@ -382,3 +382,37 @@ int FlowParams::GetSeedInjInterval() const
 }
 
 void FlowParams::SetSeedInjInterval(int val) { SetValueLong(_seedInjInterval, "What's the interval of injecting seeds into an unsteady flow advection", val); }
+
+double FlowParams::_getRakeCenter(int dim)
+{
+    Box *rakeBox = GetRakeBox();
+    if (rakeBox == nullptr) return 0.;
+    std::vector<double> minExt, maxExt;
+    rakeBox->GetExtents(minExt, maxExt);
```
shaomeng, line 391: It's encouraged to use the API with arrays.

lib/params/FlowParams.cpp
```diff
@@ -382,3 +382,37 @@ int FlowParams::GetSeedInjInterval() const
...
+    return (maxExt[dim] + minExt[dim]) / 2.;
+}
+
+void FlowParams::_setRakeCenter(int dim, double center)
+{
+    Box *rakeBox = GetRakeBox();
+    if (rakeBox == nullptr) return;
+
+    std::vector<double> minExt, maxExt;
+    rakeBox->GetExtents(minExt, maxExt);
```
shaomeng, line 401: It's encouraged to use the API with arrays.

apps/vaporgui/FlowEventRouter.cpp
sgpearse: Got it.  Fixed.

lib/params/FlowParams.cpp
sgpearse: Fixed.

lib/params/FlowParams.cpp
sgpearse: Fixed

sgpearse:
@clyne @shaomeng - This PR now adds a checkbox to enable/disable dynamic updates to all SliderEdits.  The only sliders that are currently allowed to enable/disable dynamic updates are those for orienting the flow rake center.

See FlowEventRouter.cpp:84, which calls _AllowDynamicUpdate()_ on the the X-rake-center-slider to see how this option is enabled.

SliderEdit that is allowed to enable/disable dynamic updates:
Screen Shot 2022-03-30 at 2 55 57 PM

SliderEdit that is **NOT** allowed to enable/disable dynamic updates:
Screen Shot 2022-03-30 at 2 56 16 PM

Note: Refactoring these custom menus may be needed for extensibility.  

Vapor's SliderEdit classes have the following menu options:
1) numeric formatting (Scientific notation an decimal digits)
2) min/max range setting
3) and dynamic updates

These options are currently built with the following inheritance chain:
VNumericFormatMenu::VDynamicRangeMenu::V[Double/Int]SliderEditMenu.  

A Widget may want dynamic updates (3) without numeric formatting (1) or min/max range setting (2).  This should probably be done through composition instead.

This PR follows the model that has been used in the past so it doesn't get too big, or go outside of its scope.

clyne:
I'm hading the review for this over to @StasJ. Given the widget changes Stas would be a better choice for reviewing.

shaomeng (approved):

apps/vaporgui/VIntSliderEdit.h
```diff
@@ -28,6 +31,9 @@ class VIntSliderEdit : public VSliderEditInterface {
 
     void AllowUserRange(bool allowed = true);
 
+    //! Allow the toggling of dynamic updating widgets
+    void AllowDynamicUpdate() const;
```
StasJ, line 35: The V-widget does not have a concept of params so it probably shouldn't have a function pertaining to the update frequency of params?

StasJ (changes requested):

apps/vaporgui/VIntSliderEdit.h
StasJ: I understand, but the V-Widget always outputs Changed signals and IntermediateChange signals regardless, so how does the checkbox affect it?

sgpearse:
Hmm.  I think I see what you're saying, but I'm not entirely clear.  I'll need to think about this and ping you with updates or questions.

apps/vaporgui/VIntSliderEdit.h
sgpearse: @stasj - Sorry, I'm unclear on this.  Are you saying you'd like IntermediateChange signals to be disabled unless AllowDynamicUpdate() is called?  Something like:

`void VIntSliderEdit::AllowDynamicUpdate() const { 
    _menu->AllowDynamicUpdate();
    disconnect(_slider, &VSlider::ValueChangedIntermediate, this, &VIntSliderEdit::_sliderChangedIntermediate);
}`

StasJ: AllowDynamicUpdate pertains to Dynamically Updating the Params. The V-Widgets are supposed to be unrelated to the Params and therefore should not have functions pertaining to updating the Params.

sgpearse: Please read my note above, regarding refactoring.  I'll schedule a meeting on Monday so we can discuss.

sgpearse:
@stasj I've opened #3082 to address the issues we talked about earlier this week, and have added it to my to-do list.

StasJ (approved):

PR #3073, opened 2022-03-18
Add PMetadataClasses

### LINKED ISSUE #3070: Display data set metadata
Add the ability to display (textually) data set information (metadata) panels (e.g. variable names, their coordinates, dimensions, attributes).

sgpearse:
Fixes #3070 by adding PMetadataSections to each renderer's Variables tab.  Metadata are displayed for all active variables in a renderer.

Note: The Contour Renderer seems to be setting field variables.  This will need to be addressed.

Screen Shot 2022-03-18 at 3 22 28 PM


clyne (commented):
That was fast :-). What about the global metadata (e.g. attributes and dimensions)? Maybe we don't need these?

sgpearse:
> What about the global metadata (e.g. attributes and dimensions)? Maybe we don't need these?

I don't think it can hurt.  Where can I query global attributes from?  I didn't see any accessors in the DataMgr or DataStatus.  Maybe I missed it..?

StasJ (commented):
This builds on John’s comment but a way to view the dataset metadata rather than only individual vars would be useful.

clyne:
Regarding global metadata: There 2 (maybe 3) different types to consider:

1. Dimension names and sizes: These are available from DataMgr::GetDimensionNames() and DataMgr::GetDimension(), respectively.
2. Global attributes: There are getters for these in the "DC" class, but they've not been exposed in the DataMgr. Thus wrappers for the DataMgr will need to be added (one-liners) for DC::GetAtt(), DC::GetAttNames(), DC::GetAttType(). For reference see DataMgr::GetDimensionNames()
3. variable names: For the sake of completes, providing a list of all the variables available might be helpful. These are available from DataMgr:GetDataVarNames() and DataMgr::GetCoordVarNames()

Before moving forward it would be good to consider whether **all** of the metadata information (global + variable) should be presented in one place, or if it is reasonable to split it. This PR would suggest splitting it, placing the variable metadata in the Renderer/Variables subtab, and the global data somewhere else that is TBD. Perhaps it should all be consolidated. Or maybe it is reasonable to both: have the variable metadata for the "active" variable immediately available in the Renderer/Variable subtab, but also making it available as part of a more general metadata display tool. Thoughts?

apps/vaporgui/PMetadataSection.cpp
```diff
@@ -0,0 +1,145 @@
...
+    _tree->setHeaderHidden(true);
+    _tree->header()->setStretchLastSection(false);
+    _tree->setColumnCount(2);
+    _tree->header()->setSectionResizeMode(QHeaderView::ResizeToContents);
+}
+
+void PMetadataTree::updateGUI() const
+{
+    VAPoR::RenderParams* rp = dynamic_cast<VAPoR::RenderParams*>(getParams());
+    if (rp == nullptr) MSG_ERR("Invalid RenderParams instance for acquiring variable metadata.");
```
clyne, line 37: This throws an error, but then the execution keeps proceeding, doesn't it? I.e. shouldn't you return if on a null pointer?

apps/vaporgui/PMetadataSection.cpp
```diff
@@ -0,0 +1,145 @@
...
+    qvars.removeAll(QString(""));
+
+    size_t ts = rp->GetCurrentTimestep();
+    int rLevel = rp->GetRefinementLevel();
+    int cLevel = rp->GetCompressionLevel();
+
+    _tree->clear();
+    for (auto qvar : qvars) {
+        std::vector<double> range;
+        dm->GetDataRange(ts, qvar.toStdString(), rLevel, cLevel, range);
```
clyne, line 65: GetDataRange() is an expensive operation. Can this be done lazily? I.e. only gather metadata for variables that are "open" in the display window rather than calling GetDataRange() on potentially multiple variables that the user may or may not be looking at?

apps/vaporgui/PMetadataSection.cpp
```diff
@@ -0,0 +1,145 @@
...
+        _tree->insertTopLevelItem(0,varItem);
+
+        _generateCoordVarInfo(varItem, dataVar);
+        _generateAttributeInfo(varItem, dataVar);
+    }
+}
+
+void PMetadataTree::_generateCoordVarInfo(QTreeWidgetItem* parent, const VAPoR::DC::DataVar dataVar) const {
+    VAPoR::DataMgr* dm = getDataMgr();
+    if (dm == nullptr) MSG_ERR("Invalid DataMgr instance for acquiring coordinate variable metadata.");
```
clyne, line 87: Ditto regarding not proceeding with a null pointer. Probably reasonable to simply VAssert here.

clyne (commented):
A couple of general comments:

1. The dimensions are corrupt when 2D data are displayed
2. For some renderers (e.g. 2DData) the default variable displayed never goes away. If you change the selected variable the first variable displayed doesn't get removed from the display. It's a little bit confusing.

apps/vaporgui/PMetadataSection.cpp
sgpearse: You're right.  Fixed.

apps/vaporgui/PMetadataSection.cpp
sgpearse: Fixed.

apps/vaporgui/PMetadataSection.cpp
sgpearse: I'm finding problems with this.  The QTreeWidget is cleared and regenerated on each call to PMetadataTree::updateGUI(), and the branches do not persist across params updates.

Ideally we would only reintegrate the tree when there is a change to 
1) the current time step
2) a variable's rLevel or cLevel
3) a new variable is selected.  

But because updateGUI() is const, we cannot update/refresh member variables that would track these values.

@clyne and @stasj, what do you think about the following:

Add a function to the PWidget base class, **_needsUpdate()**, that can be overridden, and includes logic that can prevent calls to a PWidget's updateGUI.  The default implementation would return true.  This would then be called at [PWidget.cpp::42](https://github.com/NCAR/VAPOR/blob/f80b3e8af6007203b3e8bd82019a6384d319dfd9/apps/vaporgui/PWidget.cpp#L42).

sgpearse:
@clyne

> The dimensions are corrupt when 2D data are displayed

Got it.  Fixed.  

Question: Should calls to DataMgr::GetDimLensAtLevel() return a set of three values for 2D variables?

> For some renderers (e.g. 2DData) the default variable displayed never goes away. If you change the selected variable the first variable displayed doesn't get removed from the display. It's a little bit confusing.

I'm not sure why, but RenderParams::GetColorMapVariableName() always returns the first variable in the list.  I think this is an external bug.

apps/vaporgui/PMetadataSection.cpp
StasJ: The dataset metadata display could be a V-Widget

apps/vaporgui/PMetadataSection.cpp
clyne: @sgpearse did you push your fix? I don't see any change here

apps/vaporgui/PMetadataSection.cpp
clyne: I will defer to @stasj on this.

apps/vaporgui/PMetadataSection.cpp
clyne: Not seeing this change either.

apps/vaporgui/PMetadataSection.cpp
sgpearse: No, not yet.  It's fixed on my local branch and this comment is for my own record keeping.  

There are lot of changes incoming.  There are several caveats I'm running into that I didn't anticipate.  I'll request re-review when ready.

clyne:
> Question: Should calls to DataMgr::GetDimLensAtLevel() return a set of three values for 2D variables?

GetDimLensAtLevel is one of a handful of remaining DataMgr methods that has not been refactored to use DimsType. Hence, the vector returned is sized by the dimensionality of the variable: 2 for 2D variables.

sgpearse:
@clyne and @stasj - Some significant changes have been made to add the requested features.  Most of the previous comments do not apply to the current version of the code.

**POpenVariableMetadataWidget** can be seen in the Variables tab for each Renderer.  This only updates when the currently open variable(s) change, or when the time step changes.  Note that there is a bug in the way RenderParams sets default variables, which can lead this to display Auxiliary variables when they're not being used.

Screen Shot 2022-04-12 at 2 04 51 PM


**PMetadataSection** displays trees for all variables in a given DataMgr at a given timestep, specified by they user.  Found at the bottom of the Scene->Viewpoint tab.  It also displays all CoordinateVariables and Global Metadata in respective trees.  These values are not updated unless the user-defined DataMgr, timestep, or variable list (as acquired from DataMgr) change.

Screen Shot 2022-04-12 at 2 05 13 PM

StasJ (changes requested):
Just the one minor bug. Looks good otherwise!

apps/vaporgui/VDoubleSliderEdit.cpp
```diff
@@ -90,7 +90,11 @@ double VDoubleSliderEdit::GetMinimum() const
 void VDoubleSliderEdit::SetMinimum(double min)
 {
     // If the new value is unchanged, or illegal, reset the menu and return
-    if (min == _slider->GetMinimum() || min >= _slider->GetMaximum() || min > _value) {
+    if (min == _slider->GetMinimum() || min > _slider->GetMaximum()) {
+        if (min < _value) {
```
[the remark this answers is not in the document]
sgpearse: Thanks, good catch.

StasJ (approved):

clyne (commented):
@sgpearse this is a great new feature, but there are still some outstanding comments that you need to address:

1. Performance: In your current implementation every single variable in the data set is read  when vapor starts. We can't do this. See my in-line comments
2. Dimensions for 2D variables are still not displayed correctly. Try loading one of the MPAS data sets. Note, for unstructured data a 2D variable will only have a single logical dimension. I believe that is what DataMgr:GetDimLensAtLevel will return.
3. Dimension names and their lengths are missing. Given that the data model in vapor based on the CF conventions, where each dimension has a name and a value, we should display these. Think of the output of ncdump.
4. It looks like you've added code to DataMgr to get global attributes, but the display (under Global Attributes) appears always empty.
5. I don't think jamming all of global information under Scene/Viewpoint is the right place for it. First,  it is certainly not intuitive to look there for a data sets metadata. Second, it makes the tab really quite long, something we've tried to avoid. How about a new, top-level "Metadata" tab to go along with "Renderers" and "Scene", or add a "Metadata browser" under the Tools menu bar along with "Data statistics"?

cc'ing @NihanthCW to give him a chance to observe the review process.

sgpearse:
> Performance: In your current implementation every single variable in the data set is read when vapor starts. We can't do this. See my in-line comments

Got it.  Branches are now only populated when opened.  The trees are reset upon timestep change.

> Dimensions for 2D variables are still not displayed correctly. Try loading one of the MPAS data sets. Note, for unstructured data a 2D variable will only have a single logical dimension. I believe that is what DataMgr:GetDimLensAtLevel will return.

I see it now.  Fixed.

> Dimension names and their lengths are missing. Given that the data model in vapor based on the CF conventions, where each dimension has a name and a value, we should display these. Think of the output of ncdump.

Do you mean we should display these in a new tree at the top of the dialog?  IE - ncdump lists the dimension names and their sizes at the top when we dump a header.  Do we want to display these at the top in the same way?

We do show the dimension names and sizes for coordinate variables, I think you're asking to make them more prominent.  Doing this would give our big metadata dialog four trees: Dimensions, Variables, Coordinate Variables, and Global Attributes.

> It looks like you've added code to DataMgr to get global attributes, but the display (under Global Attributes) appears always empty.

I think this is a bug in the implementations for DC::getAttributeNames().  I'm calling this with an empty string, which is supposed to return all the global attribute name.  This returns nothing with my POP, MPAS, and WRF data, but I do get a single global attribute for my DUKU VDC.

For example, take a look at DCMPAS.cpp:getAttNames().  This is returning after a failed call to getBaseVarInfo() instead of gathering the global attribute names.

> I don't think jamming all of global information under Scene/Viewpoint is the right place for it. First, it is certainly not intuitive to look there for a data sets metadata. Second, it makes the tab really quite long, something we've tried to avoid. How about a new, top-level "Metadata" tab to go along with "Renderers" and "Scene", or add a "Metadata browser" under the Tools menu bar along with "Data statistics"?

Sounds good.  I've made it into a dialog under the Tools menu.

clyne (approved):
This is awesome, @sgpearse. Thanks!

PR #3074, opened 2022-03-19
Issue 3062 - WRF-Fire time variable is read incorrectly 

### LINKED ISSUE #3062: WRF-Fire time variable is read incorrectly
There are two parts to this bug.

Part 1:
1) Import /glade/p/cisl/vast/vapor/data/Bugs/3062/wrfout_met_d02_2020-10-22_18:00:00
2) In the Scene/Annotation tab, enable time annotation with "Formatted date/time"

The time will be incorrect, showing 2020-10-22_17:59:28, not 2020-10-22_18:00:00.

Part 2:
1) 1) Import the following files:
    /glade/p/cisl/vast/vapor/data/Bugs/3062/wrfout_met_d02_2020-10-22_18:00:00
    /glade/p/cisl/vast/vapor/data/Bugs/3062/data/wrfout_met_d02_2020-10-22_17:59:00
2) In the Scene/Annotation tab, enable time annotation with "Formatted date/time"

The time shows the year as 2000 instead of 2020.

Incorrect: **2000**-10-22_17:59:00, 
Correct: **2020**-10-22_17:59:00.

clyne:
Fixed #3062 

This PR fixes the reported problem. It also addresses the more serious problem of incorrect time coordinate values, which is particularly problematic with unsteady flow advection. 

WRF data sets store time as a formatted string (year:month:day:hour:minute:second) that is converted to seconds since the EPOCH using UDUnits. Prior to this PR, 'seconds since the EPOCH' were stored internally as a 32-bit float, which is insufficient precision in many cases. With this PR the time coordinate is now stored as a 64-bit float. 

In addition to changing from 32-bit to 64-bit time coordinates, this PR removes a lot of dead code from the NetCDFCFCollection and NetCDFCollection classes. This was done to greatly simply the process of of adding support for 64-bit data.


sgpearse:
@clyne The first test case is fixed but the second one still gives the wrong result.  When I move to the second timestep, the annotation remains at 2020-10-22_17:59:00 instead of advancing one minute to 2020-10-22_18:00:00.

Screen Shot 2022-03-28 at 10 40 57 AM

clyne:
@sgpearse I think that is a different bug :-). I don't think the time annotation advances unless you have activated a renderer.

sgpearse (approved):

shaomeng:
Just out of curiosity, why is "seconds since the EPOCH" represented using a float/double instead of an 64-bit integer?

clyne:
Mostly because the CF conventions use the UDUNITS package for time representation, which supports the specification of time in "units" of:

<unit> since YYYY:MM:DD:HH:MM:SS, 

and the UDUNITS API will convert this to seconds since the EPOCH and return the value as a double.

shaomeng (approved):

PR #3080, opened 2022-03-29
Fixed #2840 - Show the number of cores OpenMP has available to it

### LINKED ISSUE #2840: Show the number of cores OpenMP has available to it
It would be helpful to users and developers if vapor showed the number of OpenMP cores available. Perhaps putting this information in About VAPOR?

clyne:
Fixed #2840

The number of cores available to OpenMP (as returned by omp_get_num_threads()) is now displayed under Vapor/About-Vapor. Maybe there is a better place for this?

### DIFF
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -1582,8 +1582,10 @@ void MainForm::helpAbout()
                               "Web site: http://www.vapor.ucar.edu\n"
                               "Contact: vapor@ucar.edu\n"
                               "Version: "
-                            + string(Version::GetFullVersionString().c_str()) + "\n"
-                              "OpenMP Threads: " + std::to_string(nthreads);
+                            + string(Version::GetFullVersionString().c_str())
+                            + "\n"
+                              "OpenMP Threads: "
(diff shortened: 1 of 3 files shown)


apps/vaporgui/MainForm.cpp
```diff
@@ -1565,6 +1567,11 @@ void MainForm::redo()
 
 void MainForm::helpAbout()
 {
+    int nthreads = 1;
+#pragma omp parallel
+    {
+        nthreads = omp_get_num_threads();
+    }
```
shaomeng, line 1574: If there are more than 1 thread available, the variable `nthreads` will be written simultaneously by all threads, causing a data race condition. This variable should only be written to by the 1st thread.

shaomeng (changes requested):

StasJ (commented):
Maybe this could be shown in the preferences panel next to the N Threads setting?

clyne:
> Maybe this could be shown in the preferences panel next to the N Threads setting?

Agreed. I moved it.

shaomeng (approved):

StasJ (approved):

PR #3086, opened 2022-04-14
Issue 3085 - DC::GetAttNames() not returning global attributes

### LINKED ISSUE #3085: DC::GetAttNames() not returning global attributes
To reproduce:

Make a call to DCMPAS::GetAttNames() or DCCF::GetAttNAmes() with an empty string as the argument, which should return a vector of global attribute names.  The returned std::vector<string> will be empty.

clyne:
Fixes #3085 

Accessing global attributes was never implemented by the DC* classes.

### DIFF
--- a/lib/vdc/DCCF.cpp
+++ b/lib/vdc/DCCF.cpp
@@ -21,6 +21,15 @@ using namespace std;
 
 namespace {
 
+DC::XType netcdf_to_dc_xtype(int t) {
+    switch (t) { 
+    case NC_DOUBLE: return(DC::XType::DOUBLE);
+    case NC_INT64: return(DC::XType::INT64);
+    case NC_CHAR: return(DC::XType::TEXT);
+    default: return(DC::XType::INVALID);
+    }
+}
+
 #ifdef UNUSED_FUNCTION
 // Product of elements in a vector
 //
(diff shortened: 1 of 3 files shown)


sgpearse:
@clyne thanks for the fast turnaround.  It's working for me so far but I want to test it more in #3070 before accepting it.  I'll have a better update tomorrow.

sgpearse (changes requested):
@clyne this fixes the problem, however now whenever I query DC::GetAttType() for global attributes, I always get INVALID for the DC::XType.  I can't read the global attribute value without knowing the type.  I'm seeing this on Kauffman, Katrina, MayControl, and our hist.mpas-o datasets.

clyne:
> @clyne this fixes the problem, however now whenever I query DC::GetAttType() for global attributes, I always get INVALID for the DC::XType. I can't read the global attribute value without knowing the type. I'm seeing this on Kauffman, Katrina, MayControl, and our hist.mpas-o datasets.

Fixed.

sgpearse (approved):
It works.  Thank you.

I see the changes have been pushed to the PMetadataClasses PR, #3073.  I don't mind whether we accept the changes here, or in that PR.

clyne:
> It works. Thank you.
> 
> I see the changes have been pushed to the PMetadataClasses PR, #3073. I don't mind whether we accept the changes here, or in that PR.

Oops, pushing the the PMetadataClassses PR was a mistake. I can back out those changes and submit to this PR, or leave as is. Whatever you prefer.

sgpearse:
No problem - the changes are already in my branch so it's probably easer to close #3086 and merge this one.  Same difference either way, imo.

clyne:
Closed!

PR #3091, opened 2022-04-20
add OMP support to grid range calculation

shaomeng:
null

0e1d385a6b1b0e90591bb668439ee586e453452e: [clangTidyOutput.txt](https://output.circle-artifacts.com/output/job/394ddd17-89aa-4d56-9608-ebf1fce9215c/artifacts/0/tmp/clangTidyOutput.txt)


include/vapor/OpenMPSupport.h
```diff
@@ -1,13 +1,14 @@
-#pragma once
+#ifndef OPENMPSUPPORT_H
```
clyne, line 1: Just curious. Is there a problem with using "#pragma once"? It's used elsewhere in the codebase.

clyne (commented):
@shaomeng this is great, but we need a test driver to 1: verify correctness (in this case the parallel implementation should give exact same results as serial), and 2. demonstrate performance. These could be separate or the same utilities. For correctness you might simply plug something into @sgpearse 's testGrid.cpp program, and then it would become part of our CI testing. For performance it would be helpful to have a small program that would allow us to control the grid resolution and num threads from the command line.

include/vapor/OpenMPSupport.h
shaomeng: It was giving me "multiple definition" errors.  I suspect that `#pragma once` was not helping at least.  (Commit 60dba6057f83f7d4e29aeb32fa2fa565c8070d9a finally fixed the compilation errors.)

shaomeng:
> @shaomeng this is great, but we need a test driver to 1: verify correctness (in this case the parallel implementation should give exact same results as serial), and 2. demonstrate performance. These could be separate or the same utilities. For correctness you might simply plug something into @sgpearse 's testGrid.cpp program, and then it would become part of our CI testing. For performance it would be helpful to have a small program that would allow us to control the grid resolution and num threads from the command line.

Are you able to suggest a dataset where all its values are "missing values" to test? 
I'll look into Scott's program.

clyne:
> > @shaomeng this is great, but we need a test driver to 1: verify correctness (in this case the parallel implementation should give exact same results as serial), and 2. demonstrate performance. These could be separate or the same utilities. For correctness you might simply plug something into @sgpearse 's testGrid.cpp program, and then it would become part of our CI testing. For performance it would be helpful to have a small program that would allow us to control the grid resolution and num threads from the command line.
> 
> Are you able to suggest a dataset where all its values are "missing values" to test? I'll look into Scott's program.

@shaomeng because GetRange isn't reimplemented by the various grid specializations I would think you could just use synthetic data and maybe randomly thrown in some missing values. Scott's testGrid code has utilities for generating synthetic grids. If you feel that you need to test with real data any of the ocean data sets (e.g. MOM, ROMS) should have missing values where the grid intersects the land masses.

cc @NihanthCW

include/vapor/OpenMPSupport.h
StasJ: If using include guards rather than pragma once fixes the issue the real problem is likely still lurking underneath.

StasJ (commented):
I agree with @clyne that some performance and correctness tests would be very helpful in reviewing this PR.

include/vapor/OpenMPSupport.h
shaomeng: Maybe I wasn't clear enough. My guess is that `#pragma once` is innocent for errors I run into earlier. @StasJ  rest assured: any lurking bug will be dug out and thrown away ;)

shaomeng:
All, correctness and timing code is introduced in [48c9582](https://github.com/NCAR/VAPOR/pull/3091/commits/48c9582f08d8eca482f6addac0ff386f99563b9d). It shows that the results are correct, but time is always below 1 millisecond with the current setting. @sgpearse can you bump up the test data size in one of the tests?

sgpearse (commented):
Do we want to use a single timing mechanism in Vapor?

Currently, some of our libraries use Wasp::GetTime() interface while other libraries use std::chrono.  I'd lean towards using chrono, personally.

clyne:
@shaomeng (and @sgpearse) I think we're conflating two different issues here. When we parallelize code we need to be sure the parallel version performs correctly: the parallel version should produce **approximately** the same result as the serial version (note that order of floating point operations, etc. will in most case make an **exact** result impossible). That said, in the case of GetRange(), and exact result *is* possible because there are no floating point operators, only comparisons. 

We also want know how much speedup (if any) we are getting from the parallel implementation. We can't take for granted that because some kernel is parallelized and we run it on 'n' threads it is going to run 'n' times faster. It's amazing how easy it is to come up with a parallel code that runs **slower** than the serial version.

The first part, correctness, is something we can (and probably should when it makes sense) include in our CI testing. But the CI tests need to be small and run fast. We don't have an infinite CircleCI allocation. Even if we did, we want the tests to complete quickly so that we're not waiting on PRs. 

The speedup has to be done with larger data. Much larger data. I.e. the data we use for performance tests should be large enough to give us a sense of how the parallel implementation will perform with actual data. If the data are too small, cache behavior and thread startup time will dominate the runtime. Hence, a 64^3 volume, such as Scott introduced, is not going to tell us anything. On the other hand, we probably don't need to run the performance test as part of our CI. I think it would be enough to have a tool that we can run manually, as needed, to convince ourselves we've met performance goals. Such a tool should ideally allow us to very the thread count as well as the problem size. These two factors are the hallmarks of any parallelization analysis. It might make sense to put this into the existing testGrid code. It might also make sense to have a separate drive for this.

test_apps/smokeTests/gridTools.cpp
```diff
@@ -334,6 +360,23 @@ bool RunTest(Grid *grid, bool silenceTime)
 
     PrintStats(rms, numMissingValues, disagreements, time, silenceTime);
 
+    // Test OpenMP implementation of Grid::GetRange()
+    float range_36[2] = {0.0, 1.1};
```
clyne, line 364: Why the funny initialization values (0.0, 1.1, 2.2,etc.)?

test_apps/smokeTests/gridTools.cpp
```diff
@@ -314,6 +315,31 @@ void PrintStats(double rms, size_t numMissingValues, size_t disagreements, doubl
     cout << endl;
 }
 
+// Copy of Grid::GetRange() function from VAPOR release 3.6
+//
+void GetRange_36(VAPoR::Grid* g, float range[2])
```
clyne, line 320: @shaomeng maybe we should just leave the original serial implementations inside of Grid.cpp and call it something like GetRangeSerial? I'm not sure it if is a good model that every time we parallelize a portion of the code we move the serial version into the test driver. What do you think?

test_apps/smokeTests/gridTools.cpp
```diff
@@ -334,6 +360,23 @@ bool RunTest(Grid *grid, bool silenceTime)
 
     PrintStats(rms, numMissingValues, disagreements, time, silenceTime);
 
+    // Test OpenMP implementation of Grid::GetRange()
```
clyne, line 363: Are missing values being tested? Given the logic in the code for missing value handling it seems pretty important that data with missing values gets tested.

clyne (commented):
See inline comments

test_apps/smokeTests/gridTools.cpp
shaomeng: The goal of the correctness test is to compare equality. It helps to initialize them to different values, so that they won't be equal by accident.

test_apps/smokeTests/gridTools.cpp
shaomeng: I think for complex functions, the approach you described makes sense. But finding range is something so simple that I don't think there's much value to bloat Grid with two versions of it. 

That said, I can move it back if you think it's necessary.

test_apps/smokeTests/gridTools.cpp
shaomeng: I tested using real data (MOM and ROM) you suggested earlier. Any missing values in these artificial grids are also tested.

shaomeng:
> Do we want to use a single timing mechanism in Vapor?
> 
> Currently, some of our libraries use Wasp::GetTime() interface while other libraries use std::chrono. I'd lean towards using chrono, personally.

I don't think it's a high priority to unify these 2, but if we have to, std::chrono is definitely the direction to go.

shaomeng:
> @shaomeng (and @sgpearse) I think we're conflating two different issues here. When we parallelize code we need to be sure the parallel version performs correctly: the parallel version should produce **approximately** the same result as the serial version (note that order of floating point operations, etc. will in most case make an **exact** result impossible). That said, in the case of GetRange(), and exact result _is_ possible because there are no floating point operators, only comparisons.
> 
> We also want know how much speedup (if any) we are getting from the parallel implementation. We can't take for granted that because some kernel is parallelized and we run it on 'n' threads it is going to run 'n' times faster. It's amazing how easy it is to come up with a parallel code that runs **slower** than the serial version.
> 
> The first part, correctness, is something we can (and probably should when it makes sense) include in our CI testing. But the CI tests need to be small and run fast. We don't have an infinite CircleCI allocation. Even if we did, we want the tests to complete quickly so that we're not waiting on PRs.
> 
> The speedup has to be done with larger data. Much larger data. I.e. the data we use for performance tests should be large enough to give us a sense of how the parallel implementation will perform with actual data. If the data are too small, cache behavior and thread startup time will dominate the runtime. Hence, a 64^3 volume, such as Scott introduced, is not going to tell us anything. On the other hand, we probably don't need to run the performance test as part of our CI. I think it would be enough to have a tool that we can run manually, as needed, to convince ourselves we've met performance goals. Such a tool should ideally allow us to very the thread count as well as the problem size. These two factors are the hallmarks of any parallelization analysis. It might make sense to put this into the existing testGrid code. It might also make sense to have a separate drive for this.

This PR tests correctness on all the grids, which is a good thing and doesn't need to be improved anymore. 

This PR tests performance not adequately enough, but I think it can be resolved by using a much bigger grid. The test on this much bigger grid should only be invoked manually, but not automatically on CircleCI. So if @sgpearse can provide me instructions on how to generate arbitrarily big grids, I can write a test driver.

test_apps/smokeTests/gridTools.cpp
clyne: I don't think it is necessary for  this function. Just thinking about how we deal with the general issue of verifying that  parallel versions of a function produce correct results. One way to do this is to compare with the serial implementation (assuming we accept the serial code as ground truth). I don't think copying, for example, the flow integrator into this file is something we'll want to do...

test_apps/smokeTests/gridTools.cpp
clyne: @sgpearse do we have missing values in the test grids? It would be a good idea to include if we don't. Perhaps spray some randomly into each test grid? Having a grid entirely of missing values would also be good test.

clyne:
> > Do we want to use a single timing mechanism in Vapor?
> > Currently, some of our libraries use Wasp::GetTime() interface while other libraries use std::chrono. I'd lean towards using chrono, personally.
> 
> I don't think it's a high priority to unify these 2, but if we have to, std::chrono is definitely the direction to go.

Agreed. GetTime() predates the existence of std::chrono, but I don't think it is a priority to rip out GetTime()

clyne:
> This PR tests correctness on all the grids, which is a good thing and doesn't need to be improved anymore.
> 
> This PR tests performance not adequately enough, but I think it can be resolved by using a much bigger grid. The test on this much bigger grid should only be invoked manually, but not automatically on CircleCI. So if @sgpearse can provide me instructions on how to generate arbitrarily big grids, I can write a test driver.

@shaomeng see [here](https://github.com/NCAR/VAPOR/blob/ccf6eb878c4fe112c8006c42bea0aeade9e3b207/test_apps/smokeTests/gridTools.cpp#L45) and [here](https://github.com/NCAR/VAPOR/blob/ccf6eb878c4fe112c8006c42bea0aeade9e3b207/test_apps/smokeTests/testGrid.cpp#L207)

shaomeng:
All, I've created a standalone program to test performance of this code; it can be invoked by `./test_binaries/GetRange 1024`. Let me know what you think!

test_apps/smokeTests/gridTools.cpp
sgpearse: @clyne we currently do not, so I've added them in PR #3096.  The tests include randomly assigning certain values as missing, and testing each grid type when they contain all missing values.  This seems to have turned up some bugs in the Stretched, Layered, and Curvilinear grid tests regarding RMS error.

test_apps/OpenMP/GetRange.cpp
```diff
@@ -0,0 +1,114 @@
...
+  // Create a grid to test
+  const auto blk_size = std::array<size_t, 3>{128, 128, 128};
+  auto blks = AllocateBlocks(blk_size, dims);
+  auto* grid = new VAPoR::RegularGrid(dims, blk_size, blks, {0.0, 0.0, 0.0}, {100.0, 100.0, 100.0}); 
+
+  // Fill in random values
+  std::random_device rd;
+  std::mt19937 gen(rd());
+  std::uniform_real_distribution<float> dist(-1.0, 20.0);
+  for (size_t k = 0; k < dim; k++)
```
clyne, line 86: Suggest replacing this for loop with an iterator. For larger grids (e.g. 2048^3) this initialization can take a lonnng time:

```
    VAPoR::Grid::Iterator itr;
    VAPoR::Grid::Iterator enditr = grid->end();

    itr = grid->begin();

    for (; itr != enditr; ++itr) *itr = dist(gen)
```

lib/vdc/Grid.cpp
```diff
@@ -141,24 +142,55 @@ void Grid::SetValueIJK(size_t i, size_t j, size_t k, float v)
...
+          end = this->cend();
+
+      while (*beg == missingVal && beg != end) { ++beg; }
+
+      if (beg != end) {
+        auto min = *beg;
+        auto max = *beg;
+        for (auto itr  = beg + 1; itr != end; ++itr) {
+          if (*itr != missingVal) {
+            min = std::min(*itr, min);
```
clyne, line 175: On my Mac replacing the use of the STL operators with:

```
if (*itr < min) min = *itr;
if (*itr > max) max = *itr;
```

Runs 10 to 15% faster.

clyne (commented):
Thanks for adding the benchmark. FWIW I'm seeing about a 4x speedup with 8 threads on my Mac. More importantly, the time to compute the range for a 1024^3 problem drops from about 80 seconds to about 20 seconds.

Just a couple of minor inline comments to address.

test_apps/smokeTests/gridTools.cpp
clyne: > @clyne we currently do not, so I've added them in PR #3096. The tests include randomly assigning certain values as missing, and testing each grid type when they contain all missing values. This seems to have turned up some bugs in the Stretched, Layered, and Curvilinear grid tests regarding RMS error.

Indeed it has. I'll add a comment to #3096

shaomeng:
> Thanks for adding the benchmark. FWIW I'm seeing about a 4x speedup with 8 threads on my Mac. More importantly, the time to compute the range for a 1024^3 problem drops from about 80 seconds to about 20 seconds.
I suspect that your laptop is 4 physical cores, and 8 hyperthreading cores, so you're not seeing 8X improvement.

lib/vdc/Grid.cpp
shaomeng: Interesting. Maybe it's a platform/data type specific thing, but STL algorithms such as 'std::min()' are optimized to use certain hardware instructions instead of if/else conditions, so they're supposed to be faster. I think I'll keep this one here ;)

test_apps/OpenMP/GetRange.cpp
shaomeng: That's a good point! I've made the change in 0b7c2eaa7

lib/vdc/Grid.cpp
```diff
@@ -141,24 +142,55 @@ void Grid::SetValueIJK(size_t i, size_t j, size_t k, float v)
...
-    Grid::ConstIterator enditr = this->cend();
+    const auto missingVal = GetMissingValue();
+    const auto num_vals = _dims[0] * _dims[1] * _dims[2];
+    auto num_threads = size_t{0};
+#pragma omp parallel
+    {
+      if (omp_get_thread_num() == 0)
+        num_threads = omp_get_num_threads();
+    }
+    assert(num_vals > num_threads);
```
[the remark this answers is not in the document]
shaomeng: Good point. The logic actually holds no matter the relationship between those 2 values. I've made changes in 5f4e532db.

lib/vdc/Grid.cpp
clyne: @shaomeng I think you are comparing apples to oranges :-). If the code were:

`min = a < b ? a : b`

I would agree that std::min() should be just as efficient (that would be an apple to apple comparison). But the code in question does something different: it only *conditionally* assigns a new minimum value. Thus, there are fewer assignments than with your implementation. I ran this on my may with clang12. Try it on another platform/compiler and see if you don't get similar results.

shaomeng: Interesting. I tried on cisl-vapor (Ubuntu 20, GCC 9.3) and std::min() is consistently faster. In the following screenshot, the first 3 runs with time ~2.1s are using if/else, and the following 3 runs with time ~1.8s are using std::min().
Screen Shot 2022-04-26 at 11 23 21 AM

On the other hand, I tested a super simple standalone program, and there's virtually no difference... I guess it has to be tested on the specific program one's using...

clyne: @shaomeng, just to be clear, you said you are comparing against "if/else". The code snippet I shared wasn't an "if/else", it was an "if". There is a difference because there is no "else" execution branch. So are you comparing std::min against the "if" that I shared, or the "if/else"?  Thanks.

shaomeng: Sorry, I was using "if/else" as a name for the conditional pattern, but the actual test code didn't involve an 'else' part in this case. It was a comparison between `min = std::min(val, min)` and `if (val < min) min = val;`. 

I think it ultimately comes down to if it's faster to do an assignment or to resolve a condition. It could well be a platform dependent thing...

clyne: Got it. Thanks.

clyne (approved):

StasJ (approved):

clyne:
@sgpearse , @shaomeng the range_omp_missingValues has now been merged with this branch. Although I don't see any evidence of that in this PR. Not sure how to move forward....

shaomeng:
> @sgpearse , @shaomeng the range_omp_missingValues has now been merged with this branch. Although I don't see any evidence of that in this PR. Not sure how to move forward....

I see that all the changes are here. What's appearing wrong?

sgpearse (approved):

clyne:
@sgpearse needs to resolve conflicts

PR #3093, opened 2022-04-21
Add glyphs to Particle Renderer

### LINKED ISSUE #3009: Can't change size of particle glyph
There is no way to change the size of the glyph used to render a particle, and the default size is too small to be seen (on a 2560x2600 display).

To reproduce:
1. Load the particle data in /data/Source/generateExampleDCP
2. Enable the particle render

There are no controls on the appearance tab to change the glyph size (I thought there once was, but might be wrong)

sgpearse:
Fixes #3009 by adding glyphs

d6afedd56562d735af377a2d6252f7fd710fdedd: [clangTidyOutput.txt](https://output.circle-artifacts.com/output/job/4e0a5fed-95d8-4c2b-8a4a-1272f8b21f8f/artifacts/0/tmp/clangTidyOutput.txt)


clyne (commented):
Nice work @sgpearse! I'll leave it to @stasj and @shaomeng to do a proper code review, but I have a few comments on performance and functionality. 

1. The performance for large numbers particles is not so good. @stasj may have some thoughts on how to get more out of OpenGL (eg use GLSL), but my guess is that we'll have to offer a simple point primitive as an option (like we currently have). Tweak generateExampleDCP.py so that it generates a million particles and compare the performance of the current and new implementation. The latter isn't really usable (on my laptop at lease). GLSL is probably the way to go, but @StasJ should weigh in on this.
2. Related to (1), the `stride` parameter seems to have no effect on performance with this implementation Render a million particles and change stride to 1000. Rendering speed is largely unchanged (works fine in the current implementation).
3. For consistency shouldn't we have a toggle to enable/disable lighting? 

cc @NihanthCW

lib/render/ParticleRenderer.cpp
```diff
@@ -178,4 +190,217 @@ int ParticleRenderer::_paintGL(bool)
...
+    for (int s = 0; s < nStreams; s+=2) {
+        glm::vec4 p = particles[s];
+        glm::vec4 p2 = particles[s+1];
+        const vector<flow::Particle> stream = {flow::Particle(p[0],p[1],p[2],0.,p[3]), flow::Particle(p2[0],p2[1],p2[2],0.,p2[3])};
+        sv.clear();
+        int sn = stream.size();
+
+        for (int i = 0; i < sn + 1; i++) {
+            // "IsSpecial" means don't render this sample.
+            if (i == sn || stream[i].IsSpecial()) {
```
StasJ, line 243: Why is this code here?

lib/render/ParticleRenderer.cpp
```diff
@@ -178,4 +190,217 @@ int ParticleRenderer::_paintGL(bool)
...
+    } Vertex;
+    vector<Vertex> vertices;
+    vector<int>    sizes;
+    vector<Vertex> sv;
+
+    int nStreams = particles.size();
+    for (int s = 0; s < nStreams; s+=2) {
+        glm::vec4 p = particles[s];
+        glm::vec4 p2 = particles[s+1];
+        const vector<flow::Particle> stream = {flow::Particle(p[0],p[1],p[2],0.,p[3]), flow::Particle(p2[0],p2[1],p2[2],0.,p2[3])};
```
StasJ, line 237: Do we want to convert each particle to a flow "Stream"? At the very least if we are doing this why copy and paste all the same code?

StasJ (changes requested):
To respond to @clyne comment, it seems that all of the rendering data is being generated every frame which will result in poor performance. This PR also copies the flow rendering code which, while I support code reuse, also copies the idiosyncrasies specific to the flow code which don't apply here.

lib/render/ParticleRenderer.cpp
sgpearse: Like you mentioned, an idiosyncrasy in the flow renderer.  I removed as much flow code as possible, but this section is hard to understand so I treated it like a black box.

@shaomeng, do you know if there are some more descriptive variable names I can substitute here for "sv", "sn", and "svn"?  Also, can you talk about what "IsSpecial" means?

shaomeng: Sure. In flow renderer, a "flow line" is just a sequence of particles. When you render them, you draw a line from one particle to the next. Occasionally, the flow line is disconnect, for example, it exits from one side of the volume and re-enters from the other side. In this case, I put a "special" particle in the sequence to mark that you want to skip that particular particle and restart drawing lines from the next one. Does this make sense?

lib/render/ParticleRenderer.cpp
sgpearse: > Do we want to convert each particle to a flow "Stream"?

Doing it this way made it convenient to draw the particle directions.

> At the very least if we are doing this why copy and paste all the same code?

There are several differences between ParticleRenderer::_renderParticles() and FlowRenderer::_renderStreamlines(), but many of the lines are copied.  As to why I did it this way, it wasn't just being haphazard.  

There' a rule of thumb called [the rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) that suggests refactoring take place after code has been duplicated three times.  As a rule of thumb, this is open for debate.  But if we do choose to refactor both this and the FlowRenderer, we need to agree on exactly how it should be done.

lib/render/ParticleRenderer.cpp
sgpearse: @shaomeng Yes I think so.  Thanks.  Sounds like I just need to remove the "IsSpecial" special case at line 243.

sgpearse:
@StasJ @shaomeng @clyne,

The idiosyncrasies of the Flow Renderer have been removed.  Caching has been added for performance, as well as the option to render with legacy code.  Performance comparisons are shown below, which are run on a dataset containing 1 million particles.  Note: Both the legacy Particle renderer and the one in this PR will crash when rendering 1M particles, as reported in issue #3102.  See #3102 for a modified version of generateExampleDCP.py that will generate a test file.

Uncomment ParticleRenderer.cpp:46 to print timing info on legacy and glyph rendering.

| Stride | Legacy | Glyphs |
| --- | --- | --- |
| 4* | 61397 | 122237 |
| 4 | 50888 | 66478 |
| 4 | 50637 | 66224 |
| 4 | 50592 | 66728 |
| 2* | 67615 | 206448 |
| 2 | 66607 | 135185 |
| 2 | 65748 | 133497 |
| 2 | 65849 | 133344 |

StasJ (commented):
The 1M particles without a stride works for me on the main branch.

StasJ (changes requested):
There are multiple copies of the code that load the particles from the grids. This code probably doesn't need to be duplicated.

sgpearse:
@sgpearse will validate timing on legacy particle rendering vs current again!

sgpearse:
@clyne @StasJ @shaomeng  - I've updated the table in my previous post with tests done on cisl-acworth.  I'll copy them below for convenience as well.

@StasJ - I see what you mean about the particle generator code replication.  I've added a switch so this code can be used with the legacy renderer and glyphs.  While the new code is not identical to the legacy code, it's functionally the same.

clyne (approved):
I've approved functionality/performance. 

@StasJ, @shaomeng, looking to you to review the code. Thanks.

shaomeng (commented):
Scott, this code is very difficult to understand. It uses both a `glm::vec4` and a `flow::Particle` to represent a particle, and it seems to do some conversion between the two. I don't see why `flow::Particle` is needed honestly; it seems to only complicate things...

lib/render/ParticleRenderer.cpp
```diff
@@ -52,46 +55,353 @@ using namespace VAPoR;
...
+    } Vertex;
+    vector<Vertex> vertices;
+    vector<int>    sizes;
+    vector<Vertex> sv;
+
+    int nStreams = _particles.size();
+    for (int s = 0; s < nStreams; s+=2) {
+        glm::vec4 p = _particles[s];
+        glm::vec4 p2 = _particles[s+1];
+        const vector<flow::Particle> stream = {flow::Particle(p[0],p[1],p[2],0.,p[3]), flow::Particle(p2[0],p2[1],p2[2],0.,p2[3])};
```
shaomeng, line 243: You already know that this vector has a size of 2, and the size won't change. Why not use an std::array?

StasJ (changes requested):

lib/render/ParticleRenderer.cpp
```diff
@@ -52,46 +55,353 @@ using namespace VAPoR;
...
+    glBindBuffer(GL_ARRAY_BUFFER, _VBO);
+    glBufferData(GL_ARRAY_BUFFER, sizeof(Vertex) * vertices.size(), vertices.data(), GL_STREAM_DRAW);
+    glBindBuffer(GL_ARRAY_BUFFER, 0);
+
+    _streamSizes = sizes;
+
+    _renderParticlesHelper();
+}
+
+int ParticleRenderer::_renderParticlesHelper(bool renderDirection)
```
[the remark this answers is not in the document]
sgpearse: Nope.  Thanks and fixed.

include/vapor/ParticleRenderer.h
```diff
@@ -59,7 +59,46 @@ class RENDER_API ParticleRenderer : public Renderer {
     virtual int _paintGL(bool fast);
 
 private:
+    struct {
+        size_t              ts;
+        int                 rLevel;
+        int                 cLevel;
+        std::vector<float>  tf_lut;
+        std::vector<double> tf_minMax;
```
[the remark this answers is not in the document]
sgpearse: It seems that tf_lut can have different sizes, based on some of the legacy code in the ParticleRenderer.  On line 498, there is the following check.

`assert(_cacheParams.tf_lut.size() % 4 == 0);`

I see what you mean with tf_minMax, which could be an array.  However, this variable gets populated from MapperFunction::getMinMaxMapValue() which returns a vector.  I think we would lose any performance gains by having tf_minMax be an std::array that requires an std::vector to populate it.

lib/render/ParticleRenderer.cpp
```diff
@@ -52,46 +55,353 @@ using namespace VAPoR;
...
+        _colorMapTexId = 0;
+    }
+}
 
 int ParticleRenderer::_paintGL(bool)
 {
-    int rc = 0;
+#ifdef DEBUG
+    auto start = chrono::steady_clock::now();
+#endif
```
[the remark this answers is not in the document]
sgpearse: Why do you think timing diagnostics are not good, but other kinds of diagnostics are ok?  We have DEBUG macros like this in other places in the code base, although I'm not aware of timing-debug-code being used like this.

lib/render/ParticleRenderer.cpp
sgpearse: Makes sense.  Fixed.

lib/render/ParticleRenderer.cpp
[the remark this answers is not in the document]
clyne: I have to agree with @sgpearse that timing is valuable for computationally intensive code and I don't think it should "go away". I'm not sure "ifdef DEBUG" makes sense as a macro though. "DEBUG" is often defined compiled with '-g', which is not what you want for performance evaluation. Maybe "#ifdef TIMING"?

sgpearse:
> Scott, this code is very difficult to understand. It uses both a glm::vec4 and a flow::Particle to represent a particle, and it seems to do some conversion between the two. I don't see why flow::Particle is needed honestly; it seems to only complicate things...

@shaomeng I see what you mean.  I've removed the usage of flow::Particle, and broken Legacy and Glyph codes into their own functions that don't depend on many "if (legacy)" switches.  There's just one switch now at ParticleRenderer.cpp:113.  The new functions that should make things easier to read are listed below.

_genrateParticlesLegacy()
_generateParticleGlyphs()
_renderParticlesLegacy()
_renderParticleGlyphs()

lib/render/ParticleRenderer.cpp
sgpearse: In the recent revision, I've removed the usage of flow::Particle.  There are still a handfull of lines in ParticleRenderer.cpp::_renderParticleGlyphs() that match the FlowRenderer.cpp::_renderAdvection(), but not enough commonality to share logic between the two classes, IMO.

include/vapor/ParticleRenderer.h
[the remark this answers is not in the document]
shaomeng: Ok, I was just checking. Totally fine to keep the vector structure.

lib/render/ParticleRenderer.cpp
```diff
@@ -25,6 +25,7 @@
 #include <vapor/glutil.h>    // Must be included first!!!
 
 #include <vapor/Renderer.h>
+#include <vapor/Particle.h>
```
shaomeng, line 28: Isn't it the case that `flow::Particle` is no longer used? If so, why include it here?

lib/render/ParticleRenderer.cpp
```diff
@@ -42,6 +43,8 @@
 
 using namespace VAPoR;
 
+//#define DEBUG
+
```
shaomeng, line 47: Is there any specific reason to add two lines of code that doesn't do anything?

include/vapor/ParticleRenderer.h
```diff
@@ -59,7 +59,57 @@ class RENDER_API ParticleRenderer : public Renderer {
...
+        string              varName;
+        std::vector<std::string> fieldVars;
+    } _cacheParams;
+
+    struct _vertex {
+        glm::vec3  point;
+        float      value;
+    };
+
+    //std::vector<glm::vec4> _particles;
```
shaomeng, line 82: Is there a specific reason to leave this comment line here?

shaomeng (commented):
Scott, one question for you, where can I see the clang-tidy report on this PR?

lib/render/ParticleRenderer.cpp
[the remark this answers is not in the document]
StasJ: Are you sure you included it in the commit?

lib/render/ParticleRenderer.cpp
```diff
@@ -52,96 +55,420 @@ using namespace VAPoR;
...
+    int rc = _getGrids(grid, vecGrids);
+    if (rc != 0) {
+        SetErrMsg("Could not get scalar and field grids for ParticleRenderer");
+        return rc;
+    }
+
+    bool renderGlyphs = GetActiveParams()->GetValueLong(ParticleParams::Render3DTag, true);
+    if (renderGlyphs) {
+        _generateParticleGlyphs(grid, vecGrids);
+        _renderParticleGlyphs();
```
StasJ, line 115: It appears the particles are no longer being cached.

shaomeng (approved):

lib/render/ParticleRenderer.cpp
sgpearse: Thanks, removed.

lib/render/ParticleRenderer.cpp
sgpearse: I like to have commented switches like this because I can easily turn them on/off without needing to find the macro keyword in the code.  But no biggie.  Removed.

include/vapor/ParticleRenderer.h
sgpearse: Nope.  Removed.

lib/render/ParticleRenderer.cpp
sgpearse: You're right - thanks for that and fixed.

sgpearse:
@shaomeng - You're right about clang-tidy not working.  I looked into this and it's not simple, and goes beyond the scope of this PR so I'm going to file a new ticket.

@stas - I thought I had removed "renderDirection" but I guess not.  Fixed.

lib/render/ParticleRenderer.cpp
StasJ: They're still not being cached.

StasJ (changes requested):

sgpearse:
@stasj Any thoughts on this?  Copy/pasting from above:

> They're still not being cached.
> > @StasJ - The generation of the particles is being cached. Check out line 92, where if the cache is dirty, regenerateParticles is set to true, which issues a call to _generateParticleGlyphs(). We always want to call _renderParticleGlyphs() even with a clean cache.

sgpearse:
@StasJ - The generation of the particles is being cached. Check out line 92, where if the cache is dirty, regenerateParticles is set to true, which issues a call to _generateParticleGlyphs(). We always want to call _renderParticleGlyphs() even with a clean cache.

StasJ:
> @StasJ - The generation of the particles is being cached. Check out line 92, where if the cache is dirty, regenerateParticles is set to true, which issues a call to _generateParticleGlyphs(). We always want to call _renderParticleGlyphs() even with a clean cache.

Line 269 inside _renderParticleGlyphs() uploads particle data to the GPU.

lib/render/ParticleRenderer.cpp
```diff
@@ -110,8 +110,11 @@ int ParticleRenderer::_paintGL(bool)
 
     bool renderGlyphs = GetActiveParams()->GetValueLong(ParticleParams::Render3DTag, true);
     if (renderGlyphs) {
-        if (regenerateParticles) _generateParticleGlyphs(grid, vecGrids);
-        _renderParticleGlyphs();
+        if (regenerateParticles) {
+            _generateParticleGlyphs(grid, vecGrids);
+            _renderParticleGlyphs();
+        }
+        _renderParticlesHelper();
```
StasJ, line 117: It appears `_renderParticleGlyphs()` no longer renders the particles but is only part of caching them. Could you change the name to reflect this?

sgpearse:
I agree.  I've renamed as follows:
`_renderParticleGlyphs()` to `_generateTextureData()`
`_generateParticleGlyphs()` to `_generateParticleData()`

After checking the cache, `_renderParticlesHelper()` is called regardless of cache state.

StasJ (approved):

sgpearse:
@clyne We're all set now, thanks!

PR #3095, opened 2022-04-22
add bigger grid test

sgpearse:
@shaomeng This is responding to one of the PR comments in #3091.

Is this big enough?


shaomeng:
Scott, thanks for this change, but I believe a more flexible way to test bigger grids were added to the OpenMP PR, so not need it here now.

PR #3096, opened 2022-04-25
Add missing values to smoke tests

sgpearse:
This randomly assigns missing values to our Grid tests.  It also adds a test to each tested grid type (Regular, Stretched, Layered, Curvilinear) where all values are assigned as missing-values.

This seems to break various tests for the Stretched, Layered, and Curvilinear grids.  Regular seems ok though.

Note: I also made the default arguments for the smoke test binary directory, data directory, and results directory more sensible.  Users can now specify these from their home with '~'.

### DIFF
--- a/lib/vdc/Grid.cpp
+++ b/lib/vdc/Grid.cpp
@@ -24,55 +24,42 @@ using namespace VAPoR;
 
 namespace {
 
-bool interpolate_point_on_node(
-    const std::array<float, 4> &verts, double xwgt, double ywgt,
-    float mv, float &v
-) {
+bool interpolate_point_on_node(const std::array<float, 4> &verts, double xwgt, double ywgt, float mv, float &v)
+{
     if (xwgt == 1.0 && ywgt == 1.0) {
         v = verts[0];
-        return(true);
+        return (true);
     }
     if (xwgt == 0.0 && ywgt == 1.0) {
         v = verts[1];
-        return(true);
+        return (true);
     }
     if (xwgt == 1.0 && ywgt == 0.0) {
         v = verts[2];
-        return(true);
+        return (true);
     }
     if (xwgt == 0.0 && ywgt == 0.0) {
         v = verts[3];
-        return(true);
+        return (true);
     }
 
-    return(mv);
+    return (mv);
 }
 
-bool interpolate_point_on_edge(
-    const std::array<float, 4> &verts, double xwgt, double ywgt,
-    float mv, float &v
-) {
+bool interpolate_point_on_edge(const std::array<float, 4> &verts, double xwgt, double ywgt, float mv, float &v)
+{
+    if (ywgt == 1.0 && xwgt > 0.0 && xwgt < 1.0 && verts[0] != mv && verts[1] != mv) { v = (verts[0] * xwgt) + (verts[1] * (1.0 - xwgt)); }
 
-    if (ywgt == 1.0 && xwgt > 0.0 && xwgt < 1.0 && verts[0] != mv && verts[1] != mv) { 
-        v = (verts[0] * xwgt) + (verts[1] * (1.0 - xwgt));
-    }
+    if (xwgt == 0.0 && ywgt > 0.0 && ywgt < 1.0 && verts[1] != mv && verts[3] != mv) { v = (verts[1] * ywgt) + (verts[3] * (1.0 - ywgt)); }
 
-    if (xwgt == 0.0 && ywgt > 0.0 && ywgt < 1.0 && verts[1] != mv && verts[3] != mv) { 
-        v = (verts[1] * ywgt) + (verts[3] * (1.0 - ywgt));
-    }
+    if (ywgt == 0.0 && xwgt > 0.0 && xwgt < 1.0 && verts[2] != mv && verts[3] != mv) { v = (verts[2] * xwgt) + (verts[3] * (1.0 - xwgt)); }
 
-    if (ywgt == 0.0 && xwgt > 0.0 && xwgt < 1.0 && verts[2] != mv && verts[3] != mv) { 
-        v = (verts[2] * xwgt) + (verts[3] * (1.0 - xwgt));
-    }
+    if (xwgt == 1.0 && ywgt > 0.0 && ywgt < 1.0 && verts[0] != mv && verts[2] != mv) { v = (verts[0] * ywgt) + (verts[2] * (1.0 - ywgt)); }
 
-    if (xwgt == 1.0 && ywgt > 0.0 && ywgt < 1.0 && verts[0] != mv && verts[2] != mv) { 
-        v = (verts[0] * ywgt) + (verts[2] * (1.0 - ywgt));
-    }
-
-    return(mv);
+    return (mv);
 }
 
-}
+}    // namespace
 
 Grid::Grid() { _dims = {1, 1, 1}; }
 
@@ -349,8 +336,8 @@ DimsType Grid::Dims(const DimsType &min, const DimsType &max)
     return (dims);
 }
 
-float Grid::BilinearInterpolate(size_t i, size_t j, size_t k, const double xwgt, const double ywgt) const {
-
+float Grid::BilinearInterpolate(size_t i, size_t j, size_t k, const double xwgt, const double ywgt) const
+{
     auto dims = GetDimensions();
     VAssert(i < dims[0]);
     VAssert(j < dims[1]);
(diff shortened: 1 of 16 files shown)


test_apps/smokeTests/gridTools.cpp
```diff
@@ -79,13 +79,17 @@ void MakeTriangle(Grid *grid, float minVal, float maxVal)
         for (size_t j = 0; j < y; j++) {
             for (size_t i = 0; i < x; i++) {
                 value = value == minVal ? maxVal : minVal;
-                grid->SetValueIJK(i, j, k, value);
+                if (addRandomMissingValues) {
+                    if (rand()%2) grid->SetValueIJK(i,j,k, grid->GetMissingValue());
```
shaomeng, line 83: Using rand() is considered a pretty bad programming practice (https://stackoverflow.com/questions/52869166/why-is-the-use-of-rand-considered-bad). Please consider using its better alternatives.

clyne:
This PR has uncovered some bugs in Grid::GetValueLinear() that will need to be addressed before this can be merged. Assigning myself to fix the bugs.

test_apps/smokeTests/gridTools.cpp
```diff
@@ -67,39 +68,60 @@ template<typename T> vector<T *> AllocateBlocksType(const vector<size_t> &bs, co
...
     size_t y = dims[Y];
     size_t z = dims[Z];
 
+    std::random_device rd;  //Will be used to obtain a seed for the random number engine
+    std::mt19937 gen(rd()); //Standard mersenne_twister_engine seeded with rd()
+    std::uniform_int_distribution<> distrib(0,1);
+
     float value = minVal;
+    float missingValue = grid->GetMissingValue();
+    int missing=0;
```
shaomeng, line 84: Is this variable used anywhere? If not, how about not declare it?

sgpearse: A diagnostic variable that I missed.  Fixed.

shaomeng: Do you think it's gonna be helpful to catch such instances (diagnostic variable that people forgot to delete) automatically by either GCC or clang-tidy? If so, I can open a request.

sgpearse: I don't think we need additional checks for this kind of warning, since GCC did catch it in the failed CI test.  I'm sure clang-tidy will notice this as well.  So I don't know what more could be done for compiler warnings.

shaomeng: Got it. As long as some automated mechanism is catching it, I'm good :)

shaomeng (approved):

test_apps/smokeTests/gridTools.cpp
sgpearse: Yes - the reason this one wasn't obvious is because the smoke tests were already failing.  The warning test didn't really stand out like it normally would :\

clyne:
@sgpearse the grid classes handle interpolation with missing values inconsistently. It was not obvious to me which approach was the "right" one, none of them being completely satisfying. After a bit of googling I've come to believe that being consistent with Xarray (SciPy) is probably the least bad thing to do. Though it is not documented anywhere that I can find, here is what Xarray/SciPy do with missing values (NaNs) in 1D, which also generalizes to nD. If an interpolant point b is in the range a <= b <= c, and either f(a) or f(c) is evaluates to NaN, then f(b) will interpolate to NaN. Seems simple, but one might expect that if f(a) is valid, f(c) is a NaN, and b == a, then f(b) might interpolate to f(a). Nope. That is not what happens. Moreover, now consider known values at a,b,c, where a < b < c, and interpolating a point d, where d == b. The point d might be interpolated in the range [a..b] or [b..c]. Depending on the choice, and whether f(a) is NaN and f(c) is valid, or f(a) is valid f(c) is NaN, f(d) might interpolate to NaN depending on whether the algorithm uses the range [a..b] or [a..c]. 

Bottom line, as I said making our missing value handling consistent with Xarray and SciPy is probably the best choice. The implication is that the current CompareIndexToCoords() code will need to be changed. When missing values are present the current CompareIndexToCoords() won't work with linear interpolation (nearest is fine). So either linear can be removed when there are missing values, or CompareIndexToCoords() should recognize that GetValueAtIndex() main return a valid value, while GetValue() may return a NaN. 

Does this make sense? Correcting the inconsistency as described above should be a pretty small change.

sgpearse:
> Does this make sense? Correcting the inconsistency as described above should be a pretty small change.

Not entirely, but I need to give this more than a cursory glimpse.  I'll get on this after my upcoming iteration on #3089, which I'm hoping to finish tomorrow.

sgpearse:
> ... or CompareIndexToCoords() should recognize that GetValueAtIndex() main return a valid value, while GetValue() may return a Infinity.

I'll try this route.  I think in the case where GetValueAtIndex() and GetValue() both register as missing, we can consider the value as truly missing.  When GetValueAtIndex() is non-missing and GetValue() is missing, we discard that sample.  Does that sound right?  Changes are now posted in gridTools.cpp:219.

clyne:
@sgpearse just sit tight. I'm looking at this now.

clyne:
I've made the changes to Grid::GetLinearValue so that all grid classes handle linear interpolation with missing values consistently. I've also updated gridTools.cpp to correctly assess validity when missing values are present. I've also added a test case where the are NO missing values in the data.

@sgpearse, @shaomeng, you should both review these changes. @shaomeng once you've both approved you'll need to merge this into your 'range_omp' branch. Per our conversation yesterday this branch was not created from 'main'

clyne (approved):

lib/vdc/UnstructuredGrid2D.cpp
```diff
@@ -251,7 +251,18 @@ float UnstructuredGrid2D::GetValueLinear(const CoordType &coords) const
     VAssert(face < GetCellDimensions()[0]);
 
     double value = 0;
-    for (int i = 0; i < nodes.size(); i++) { value += AccessIJK(nodes[i], 0, 0) * lambda[i]; }
+    float  mv = GetMissingValue();
+    for (int i = 0; i < nodes.size(); i++) {
+        float v = AccessIJK(nodes[i], 0, 0);
+        if (v == mv) {
+            if (lambda[i] != 0.0)
+                return (mv);
```
shaomeng, line 259: Need to free up memory block of `lamba`. Better: use a vector for lambda.

lib/vdc/UnstructuredGridLayered.cpp
```diff
@@ -224,15 +224,45 @@ float UnstructuredGridLayered::GetValueLinear(const CoordType &coords) const
 
     // Interpolate value inside bottom face
     //
+    float mv = GetMissingValue();
+
     float  z0 = 0.0;
     size_t k0 = indices[1];
-    for (int i = 0; i < lambda.size(); i++) { z0 += AccessIJK(nodes2D[i], k0) * lambda[i]; }
+    for (int i = 0; i < lambda.size(); i++) {
```
shaomeng, line 231: I don't know what kind of information `lambda` holds, but it isn't obvious to me that lambda has a small (less than 2^31) size.  Recommend giving `i` a type of `size_t`.

shaomeng (approved):

lib/vdc/Grid.cpp
```diff
@@ -22,6 +22,45 @@
...
+{
+    if (ywgt == 1.0 && xwgt > 0.0 && xwgt < 1.0 && verts[0] != mv && verts[1] != mv) { v = (verts[0] * xwgt) + (verts[1] * (1.0 - xwgt)); }
+
+    if (xwgt == 0.0 && ywgt > 0.0 && ywgt < 1.0 && verts[1] != mv && verts[3] != mv) { v = (verts[1] * ywgt) + (verts[3] * (1.0 - ywgt)); }
+
+    if (ywgt == 0.0 && xwgt > 0.0 && xwgt < 1.0 && verts[2] != mv && verts[3] != mv) { v = (verts[2] * xwgt) + (verts[3] * (1.0 - xwgt)); }
+
+    if (xwgt == 1.0 && ywgt > 0.0 && ywgt < 1.0 && verts[0] != mv && verts[2] != mv) { v = (verts[0] * ywgt) + (verts[2] * (1.0 - ywgt)); }
+
+    return (mv);
```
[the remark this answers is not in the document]
clyne: Oops. Good catch. We apparently don't have a test case that hits this code. Once again it would be great to have a unit test framework to catch these things.

lib/vdc/UnstructuredGrid2D.cpp
clyne: Thanks. Done. Replaced with smart pointers.

lib/vdc/UnstructuredGridLayered.cpp
clyne: Done.

shaomeng (approved):

PR #3097, opened 2022-04-26
Add more sample datasets

sgpearse:
This PR adds more sample datasets to our downloads page.  It also promotes the "Sample Data" section up one level so it's more visible in the main table of contents.

Changes can be build locally, or viewed [here](https://vapor.readthedocs.io/en/datasamples/downloads/downloads.html#sample-data).


shaomeng:
I think it provides little value to offer downloading a 10x10x10 synthetic data set? How about we bump it up to a bigger dimension (lower hundreds) so that people can at least find something interesting out of it?

sgpearse:
> I think it provides little value to offer downloading a 10x10x10 synthetic data set? How about we bump it up to a bigger dimension (lower hundreds) so that people can at least find something interesting out of it?

The main focus of this is to demonstrate how BOV files are composed.

I agree that it's preferable to have Pablo's LES as an example, but I'm backlogged.  It's not a good use of my time right now to hyperslab his file into a reasonably sized BOV dataset.  I think this would be a good documentation ticket that can be done at a later time.

StasJ (approved):

clyne (approved):
Thanks for doing this, @sgpearse

sgpearse:
> Understood. How about providing this file (128x128x128) to users? It's got some interesting features inside at least.
> https://github.com/shaomeng/SPERR/blob/main/test_data/wmag128.float

Great, that sounds good.  Thanks.  I have a .bov file to associate with it and I'll add it to the list.  

I'm going to also keep the simple "Brick" example and rename it, because it demonstrates a multi-variable/multi-timestep BOV dataset.

PR #3131, opened 2022-06-15
Add 'double' as a data type to raw2vdc

### LINKED ISSUE #3130: Typo in raw2vdc
raw2vdc's help menu does not state what data types can be used with the -type flag.  These types are int8, float32, and float64.  On the other hand, vdccreate accepts 'double' as well as these types.  Users end up feeding 'double' to raw2vdc after running vdccreate, and raw2vdc silently uses float32 instead.

This adds 'double' as an option to raw2vdc.

sgpearse:
Adds "double" to raw2vdc to be consistent with vdccreate.

Fixes #3130

### DIFF
--- a/apps/raw2vdc/raw2vdc.cpp
+++ b/apps/raw2vdc/raw2vdc.cpp
@@ -27,7 +27,7 @@ struct opt_t {
 } opt;
 
 OptionParser::OptDescRec_T set_opts[] = {{"varname", 1, "var1", "Name of variable"},
-                                         {"type", 1, "float32", "Primitive type of input data"},
+                                         {"type", 1, "float32", "Primitive type of input data.  Valid types are int8, float32, float64, and double"},
                                          {"lod", 1, "-1",
                                           "Compression levels saved. 0 => coarsest, 1 => "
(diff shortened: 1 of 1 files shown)


apps/raw2vdc/raw2vdc.cpp
```diff
@@ -101,7 +102,7 @@ int read_data(FILE * fp,
     if (type.compare("float32") == 0) {
         float *sptr = (float *)buffer;
         for (size_t i = 0; i < n; i++) { *dptr++ = (float)*sptr++; }
-    } else if (type.compare("float64") == 0) {
+    } else if ((type.compare("float64") == 0) || (type.compare("double") == 0)) {
         double *sptr = (double *)buffer;
         for (size_t i = 0; i < n; i++) { *dptr++ = (float)*sptr++; }
```
shaomeng, line 107: Just a sanity check, does this code mean that if the input data is in double precision, it's truncated to single precision when read in?

sgpearse: No, it just treats the "double" token the same as the "float64" token when used with the -type flag.  I'll add a link to the issue this fixes for better context.

clyne (approved):

PR #3110, opened 2022-05-25
New3rd party libraries

sgpearse:
This PR updates Vapor to use ZFP enabled HDF5 and an updated version of Ospray. Users must set the HDF5_PLUGIN_PATH to Vapor's /lib directory to allow for ZFP enabled compression.

Since not all libraries were rebuilt, the newly built libraries had to be targeted at /usr/local/VAPOR-Deps/2019-Aug again.

The new libraries can be found on [google drive](https://drive.google.com/drive/u/0/folders/0B0dQMtxB89M0Q2F2U3ZaQ2dVeWs?resourcekey=0-KjrYiMNi3FG-JnLKuDFupw). They have been named 2019-Aug-[OS]-2022.tar.xz

Links here:
[Ubuntu](https://drive.google.com/file/d/1CD3qrPuj03S8x1XshDI_3GRqCeUVLvpm/view?usp=sharing)
[CentOS](https://drive.google.com/file/d/1In3Q8fA0VCm2en1ncBQI6dzUfatLDA0y/view?usp=sharing)
[OSX](https://drive.google.com/file/d/1KnUoBfOLuF6VJMXG6KcljjKMpr4VGxQ6/view?usp=sharing)
[Windows](https://drive.google.com/file/d/1fzZ-mbY4Cek1TRsaKm79a08Od5gNCogk/view?usp=sharing)

d4810cbbaa453e7af67529a32c104f396c22f466: [clangTidyOutput.txt](https://output.circle-artifacts.com/output/job/94eb3331-1df6-43c6-8f15-087b54654043/artifacts/0/tmp/clangTidyOutput.txt)

Fixes #3179 
Fixes #3158 
Fixes #3174 
Fixes #2816

PR #3132, opened 2022-06-16
Fix Windows build for PythonAPI

sgpearse:
CMakeLists.txt is currently missing the Python libraries on configuration.

15510c73eb4ddb04c4a2bd186374d16b65f28966: [clangTidyOutput.txt](https://output.circle-artifacts.com/output/job/a27d512d-db90-425a-9f6f-c97b3bd97e2d/artifacts/0/tmp/clangTidyOutput.txt)


StasJ (commented):
Could this be fixed by just using the same version of Python in Windows as Linux/Mac in #3110?

sgpearse:
> Could this be fixed by just using the same version of Python in Windows as Linux/Mac in https://github.com/NCAR/VAPOR/pull/3110?

Hypothetically, but Windows has so many quirks that it would take some digging.  [Here are our windows build notes](https://docs.google.com/document/d/1XNBmoUvxGn9I0fy9xvB1m5PQyOI32TtdyMbwfOve0QQ/edit) which have some strange caveats.  I don't know why the Unix platforms append "m" to their installation path and Windows does not.

I can try just removing the "m" from the third party library directory, which would leave CMakeLists.txt unchanged.  While our cmake file would be cleaner, there'd be an additional weird step for building Python on windows.  Alternatively, we could rebuild Python in #3110.  I think this would take a significant amount of time, particularly on windows.

Thoughts?

StasJ:
The appended m signifies python was built with pyMalloc rather than the system malloc.

> Hypothetically, but Windows has so many quirks that it would take some digging. [Here are our windows build notes](https://docs.google.com/document/d/1XNBmoUvxGn9I0fy9xvB1m5PQyOI32TtdyMbwfOve0QQ/edit) which have some strange caveats. I don't know why the Unix platforms append "m" to their installation path and Windows does not.

I think it is the other way around, Unix uses non-m while windows uses m. It seems in your change it is explicitly requesting the non-m version on windows? Also you can specify multiple versions as valid options which the original code already does. It seems the only difference is the location of the python library within the third party libs which we should be able to have consistent between windows and non-windows?

sgpearse:
@StasJ This is a rabbit hole that I'd like to revisit with #3110.  Windows builds take ~30 minutes to test for even slight changes.  

I think it's more effective to run tests when we update the rest of the 3rd party libraries, rather than minute changes like this.  If there's something imperative that we need, then I agree we should do it.  But I think this change is more effort than it's worth, unless I'm missing something.

sgpearse:
@sgpearse will try adding the windows path to the NAMES and PATHS arguments which can take multiple values.

```        
find_library(
            Python_LIBRARIES
            NAMES python${PYTHONVERSION} python${PYTHONVERSION}m
            PATHS "${THIRD_PARTY_LIB_DIR}"
            NO_DEFAULT_PATH
        )```

**Comment by sgpearse** [comment]
@stasj I made the change to find_library() like we discussed, but it seems there are other compile issues that didn't arise for me when I first submitted this change last month.  I've taken a shot at fixing them but problems keep arising related to the "glad" library.  I've tried the following changes, which are not in this PR.  I can keep helping with this but I wanted to get your thoughts before continuing.

Add the following calls VAPOR/CMakeLists.txt around line 134:
```
include_directories("${PROJECT_SOURCE_DIR}/lib/osgl/include")
include_directories("${PROJECT_SOURCE_DIR}/lib/osgl/glad/include")
```

Additionally, make the following flags conditional in lib/osgl/glad/CMakeLists.txt at line 75:
```
if(NOT WIN32)
    target_compile_options(${LIB} PRIVATE -Wno-deprecated-declarations)
    target_compile_options(${LIB} PRIVATE -Wno-conversion-null)
endif ()
```

clyne:
@sgpearse can we close this? Other options?

sgpearse:
@clyne I don't think it should be closed until we're successfully building on windows.  The last time I was working on this, there were linker problems with the glad library.  As I understand, glad is a set of programmatically generated files that have needed some hand-manipulation.  It's something I'm pretty unfamiliar with.

clyne:
How about labeling it as postponed or make it a draft so that we don't keep discussing? Note also that there is something weird because it says 179 files changed....

sgpearse:
> How about labeling it as postponed or make it a draft so that we don't keep discussing?

Sounds good.

> Note also that there is something weird because it says 179 files changed....

That is weird.  This branch was derived from python_api and is being merged into main, but github doesn't seem to acknowledge that the python_api branch has been merged yet.  I don't know why, but the only file that's changed is the top level CMakeLists.txt.  I'm going to just close this one and open a new PR that branches off of main, and add a postponed label to that one.

PR #3144, opened 2022-07-07
Update cmake on Ubuntu

sgpearse:
Updates our CircleCI build systems to cmake's latest version provided by pip. Currently 3.22.5.

Installation log can be found [here](https://app.circleci.com/pipelines/github/NCAR/VAPOR/5780/workflows/f2fc453e-f619-4ade-a657-4c0652adec89/jobs/13713/parallel-runs/0/steps/0-103).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -616,11 +616,11 @@ workflows:
     jobs:
       - clang-tidy
       - build_ubuntu18
-      #- build_centos7
-      - test_clang_format
+      - build_centos7
+      #- test_clang_format
       #- build_win10_installer
       #- build_osx_installer
-      - build_ubuntu18_installer
+      #- build_ubuntu18_installer
       #- build_centos7_installer
       #- release_weekly_installers:
       #    requires:


sgpearse:
@StasJ I'm shaking my head at the same question.  Path of least resistance.  Here's the rabbit hole:  https://askubuntu.com/questions/829310/how-to-upgrade-cmake-in-ubuntu

StasJ (approved):
> @StasJ I'm shaking my head at the same question. Path of least resistance. Here's the rabbit hole: https://askubuntu.com/questions/355565/how-do-i-install-the-latest-version-of-cmake-from-the-command-line

You could also download the CMake binaries from KitWare however this works.

sgpearse:
> You could also download the CMake binaries from KitWare however this works.

Yes, but it involves an 10+ step procedure from a bare-bones ubuntu environment, outlined in the above link.  Even after those steps, there are errors abound.  The path of least resistance I've found is pip.  In my experience with CI and Docker, fewer commands tend to be more understandable and more reliable.

PR #3155, opened 2022-07-26
Fix 2DRenderer height offset

### LINKED ISSUE #2988: 2DRenderer has an incorrect height offset
The 2DRenderer seems to be scaled too much in the following session file.

`/glade/p/cisl/vast/vapor/Bugs/2998/2998.vs3
`
Note the difference in height between the enabled 2DRenderer, and its corresponding WireFrame renderer.

Screen Shot 2022-02-01 at 8 08 25 PM

To reproduce:
1) On Casper, load the session file /glade/p/cisl/vast/vapor/Bugs/2998/2998.vs3

Alternatively, import the NetCDF-CF file below and play with 2d and 3d renderers.

`/glade/scratch/pearse/marshall/coen/212.nc`

StasJ:
Fix #2988

### DIFF
--- a/lib/render/ImageRenderer.cpp
+++ b/lib/render/ImageRenderer.cpp
@@ -159,7 +159,7 @@ int ImageRenderer::GetMesh(DataMgr *dataMgr, GLfloat **verts, GLfloat **normals,
     // If we are terrain mapping the image or if both the image and the
     // data are geo-referenced
     //
-    double defaultZ = GetDefaultZ(dataMgr, myParams->GetCurrentTimestep());
+    double defaultZ = 0;
     if (!myParams->GetHeightVariableName().empty() || (myParams->GetIsGeoRef() && !dataMgr->GetMapProjection().empty())) {
         // Get the width and height of the image texture. These
         // will be used to set the width and height of the mesh.
--- a/lib/render/TwoDDataRenderer.cpp
+++ b/lib/render/TwoDDataRenderer.cpp
@@ -274,7 +274,7 @@ int TwoDDataRenderer::GetMesh(DataMgr *dataMgr, GLfloat **verts, GLfloat **norma
 
     VAssert(g);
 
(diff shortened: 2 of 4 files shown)


clyne (approved):
Thanks @StasJ !

sgpearse:
While the 2D and Wireframe renderers match each other now, unfortunately it seems the height offset is incorrect.  Admittedly, my bug report was in error making it hard to reproduce this.  I've updated a session file at /glade/p/cisl/vast/vapor/Bugs/2988/2988.vss.

What is shows is the 2D and Wireframe renderers matching each other's height.  However they're being drawn in the middle of the DVR when they should be drawn beneath it.  On the main branch, our wireframe renderer will be drawn at the correct height.  I believe the base of the issue is DataMgrUtils::GetExtents() not returning correct minimum Z values for the Image and 2D renderers.

main:
Screen Shot 2022-07-28 at 10 28 54 AM

issue2988:
Screen Shot 2022-07-28 at 10 29 49 AM

clyne:
@sgpearse your session file is applying a height variable, 'HGT'. Try setting to 'none'

sgpearse:
@clyne I'm not sure I follow.  We need to apply the height variable in this use case.  The height-matched renderers (2D, Image, and Wireframe) need to follow along the base of the DVR.

clyne:
@sgpearse the "height" variable specifies a displacement from the bottom of the domain, not the absolute vertical coordinate. We should probably add a tool tip to this effect. Having the ability for the "height" variable to be treated either as a displacement or an absolute coordinate might be a desirable future option. Bottom line: I believe that the behavior you are seeing is correct.

sgpearse:
@clyne - Got it.  I can't think of a use case where the height isn't depicted at the absolute vertical coordinate.  I'll open a ticket and we can talk about it next meeting.

sgpearse (approved):

PR #3165, opened 2022-08-01
fix nan comparison

### LINKED ISSUE #3087: Flow renderer crashes on NetCDF data
To reproduce:
1) Import the three NetCDF files in /glade/p/cisl/vast/vapor/Bugs/3087
2) Create  flow renderer and select Bx as the X field variable
3) Enable the renderer, vapor will crash

shaomeng:
fix

f5df70909abc081425143f3664579ce4ef35e90a: [clangTidyOutput.txt](https://output.circle-artifacts.com/output/job/838aa6f4-a1a3-45af-89bc-934217a46e8c/artifacts/0/tmp/clangTidyOutput.txt)

### DIFF
--- a/lib/flow/Advection.cpp
+++ b/lib/flow/Advection.cpp
@@ -555,9 +555,6 @@ int Advection::_advectRK4(Field *velocity, const Particle &p0, double dt, Partic
     p1.location = p0.location + dt32 / 6.0f * (k1 + 2.0f * (k2 + k3) + k4);
     p1.time = p0.time + dt;
 
-    if (!std::isnan(p0.location.x) && std::isnan(p1.location.x))
-      std::printf("Wow!\n");
-
     return 0;
 }
 
--- a/lib/flow/Advection.cpp
+++ b/lib/flow/Advection.cpp
@@ -678,7 +678,7 @@ float Advection::_applyPeriodic(float val, float min, float max) const
 
 void Advection::_printNonZero(int rtn, const char *file, const char *func, int line) const
 {
-#ifndef NDEBUG
+#ifdef VPRINT
     if (rtn != 0) {    // only print non-zero values
         printf("Rtn == %d: %s:(%s):%d\n", rtn, file, func, line);
     }
--- a/lib/flow/Advection.cpp
+++ b/lib/flow/Advection.cpp
@@ -555,6 +555,9 @@ int Advection::_advectRK4(Field *velocity, const Particle &p0, double dt, Partic
     p1.location = p0.location + dt32 / 6.0f * (k1 + 2.0f * (k2 + k3) + k4);
     p1.time = p0.time + dt;
 
+    if (!std::isnan(p0.location.x) && std::isnan(p1.location.x))
+      std::printf("Wow!\n");
+
     return 0;
 }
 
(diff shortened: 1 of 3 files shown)


lib/flow/VaporField.cpp
```diff
@@ -237,10 +237,15 @@ int VaporField::GetVelocity(double time, const glm::vec3 &pos, glm::vec3 &veloci
             missingV[i] = grid->GetMissingValue();
         }
         auto  hasMissing = glm::equal(velocity, missingV);
+        // If missing values are represented using NaN, you cannot compare equality with them!
```
clyne, line 240: One small nit: missing values should NEVER be represented as NaNs because, as you say, you can't compare them. The NaN's must be either either in the existing data or generated by the interpolation somehow.

shaomeng: I think it's from the specific data set in this case. I.e., `grid->GetMissingValue()` would return you a nan... Does this header look like it's using nan? 

```
netcdf test_bx_vapor {
dimensions:
	time = 1 ;
	x = 168 ;
	y = 93 ;
	z = 46 ;
variables:
	double Bx(time, x, y, z) ;
		Bx:_FillValue = NaN ;
		Bx:units = "G" ;
	double time(time) ;
		time:_FillValue = NaN ;
		time:axis = "T" ;
		time:units = "seconds since 2000-0101" ;
	double x(x) ;
		x:_FillValue = NaN ;
		x:axis = "Z" ;
		x:units = "m" ;
	double y(y) ;
		y:_FillValue = NaN ;
		y:axis = "Y" ;
		y:units = "m" ;
	double z(z) ;
		z:_FillValue = NaN ;
		z:axis = "X" ;
		z:units = "m" ;
}
```

lib/flow/Advection.cpp
```diff
@@ -678,7 +678,7 @@ float Advection::_applyPeriodic(float val, float min, float max) const
 
 void Advection::_printNonZero(int rtn, const char *file, const char *func, int line) const
 {
-#ifdef VPRINT
+#ifndef NDEBUG
```
StasJ, line 681: Prints hundreds of debug lines that are probably not necessary for every debug build.

lib/flow/VaporField.cpp
```diff
@@ -237,10 +237,15 @@ int VaporField::GetVelocity(double time, const glm::vec3 &pos, glm::vec3 &veloci
...
+        for (int i = 0; i < 3; i++) {
+          if (std::isnan(missingV[i]) && velocity[i])
```
StasJ, line 242: I would try and avoid implicitly converting float to bool.

StasJ (changes requested):

lib/flow/Advection.cpp
shaomeng: Well, I wasn't able to print out hundreds of lines debug output, but it's reverted in 3e4e4c7b0 .

lib/flow/VaporField.cpp
shaomeng: Good catch! It was a typo and is fixed in be4afab5c .

lib/flow/VaporField.cpp
clyne: You are correct. I thought using NaNs for missing values were prohibited by the CF conventions, but that is not the case. Unfortunately, this means that there are several other places in the code that should be changed to handle missing values that are NaNs. I will open an issue.

clyne (approved):

lib/flow/Advection.cpp
StasJ: Might've been my session file. Thanks.

StasJ (approved):

clyne:
@shaomeng will merge as soon as passing checks.

shaomeng:
> @shaomeng will merge as soon as passing checks.

Could @sgpearse suggest why circle ci is failing, and how to fix? This PR touched 2 cpp files, but circle ci fails on python.

PR #3168, opened 2022-08-02
Add documentation to setters, getters, and tags listed in VAPOR/apps/pythonapi/vapor/renderers.py

### LINKED ISSUE #3138: More extensive documentation of Params needed
The Python API generates documentation for all functions automatically from the Params classes. While it is awesome to be able to reuse documentation for  two purposes (GUI and Python API), unfortunately many of the Params class methods are not adequately documented. For example, the documentation for FlowParams. SetRenderGlyphType, which takes an int argument, simply says:

`When rendering samples, determines whether samples are rendered as circles or arrows.`

This is fine in the context of the GUI because the integer options are mapped to strings by Qt, but this documentation provides no help at all for someone trying to use the Python API. A more egregious example is FlowParams. SetSeedGenMode(). There is no documentation for this parameter at all, and thus little hope for a Python user to figure out what valid inputs might be, or how this function changes the behavior of a flow rendering. 

N.B. we were able to get a way with being lax about much of the Parameter documentation for the GUI because the results of manipulating a slider, thumb wheel, etc. were often obvious. This is not the case with a programmatic interface.

sgpearse:
Fix #3138.

I've hit a few caveats with this that I've documented [here](https://docs.google.com/document/d/11Tim67ISrWPebs2xRhxeH2DtjDOM2UOE_rQAnd1O6aE/edit).

The only caveat that I haven't figured out regards the documentation for SliceParams, which does not transfer into the python_api help() menu. Copied from the doc:

> If a Params class has only inherited functions defined in its FuncWrapperStrList in renderer.py, no documentation will be built for those functions?
> 
> IE - the Python class SliceRenderer only has three inherited functions from RenderParams.h in its FuncWrapperStrList (GetSlicePlaneRotation, GetSlicePlaneOrigin, and GetSlicePlaneNormal). It does not populate documentation for

3b9813ddedb0a412b183690b9442467b17fc3172: [clangTidyOutput.txt](https://output.circle-artifacts.com/output/job/b62057f7-b235-4cd8-a0e2-333d9aa92af9/artifacts/0/tmp/clangTidyOutput.txt)


include/vapor/BarbParams.h
```diff
@@ -54,6 +57,8 @@ class PARAMS_API BarbParams : public RenderParams {
     //! \retval double line thickness
     double GetLineThickness() const { return GetValueDouble(_thicknessScaleTag, 1.f); }
 
+    //! Set line thickness in voxels
```
clyne, line 60: Voxels? I don't think that is correct. The name of the tag, _thicknessScaleTag, would suggest that this a multiplier applied to the default line thickness

include/vapor/FlowParams.h
```diff
@@ -53,35 +53,94 @@ class PARAMS_API FlowParams : public RenderParams {
...
+    //! Sets the type of flow rendering algorithm being used.
+    //! Steady flow (streamlines) renders time-invariant trajectories that follow a vector field at a single timestep.
+    //! Unsteady flow (pathlines) render time-variant trajectories that advect through the timeseries of a loaded dataset.
+    //! \param[in] bool - Steady/streamlines = 1, Unsteady/pathlines = 0
     void SetIsSteady(bool steady);
+    
+    //! Gets the type of flow rendering algorithm being used.
+    //! Steady flow (streamlines) renders time-invariant trajectories that follow a vector field at a single timestep.
+    //! Unsteady flow (pathlines) render time-variant trajectories that advect through the timeseries of a loaded dataset.
+    //! \retval bool - Steady/streamlines = 1, Unsteady/pathlines = 0
```
clyne, line 65: For these longer setter/getter documentation strings would it make sense to make use of \copydoc to avoid replication and improve maintainability? E.g. Document the setter (or getter) with the long string, and use \copydoc for the getter (or setter).

include/vapor/FlowParams.h
```diff
@@ -53,35 +53,94 @@ class PARAMS_API FlowParams : public RenderParams {
...
     bool GetIsSteady() const;
 
+    //! Get the multiplier being applied to the flow advection algorithm.
+    //! If there happens to be a mismatch between the units of your data's domain and the units of a variable such as wind speed,
+    //! you can scale the wind field with this parameter.  IE - If your data's domain is written in kilometers but your wind
+    //! vectors are in meters, you can apply a velocity multiplyer of 1000 to correct the mismatch.
```
clyne, line 71: To convert velocity in meters to kilometers don't you want to multiply by 0.001, not 1000.0?

include/vapor/FlowParams.h
```diff
@@ -53,35 +53,94 @@ class PARAMS_API FlowParams : public RenderParams {
...
+    
+    //! Set the multiplier being applied to the flow advection algorithm.
+    //! If there happens to be a mismatch between the units of your data's domain and the units of a variable such as wind speed,
+    //! you can scale the wind field with this parameter.  IE - If your data's domain is written in kilometers but your wind
+    //! vectors are in meters, you can apply a velocity multiplyer of 1000 to correct the mismatch.
+    //! \param[in] double - Velocity field multiplier for flow rendering
     void   SetVelocityMultiplier(double);
 
+    //! Get the number of steps to advect a steady flow line (aka a streamline)
+    //! \retval long - The number of steps a steady flow line is advected
```
clyne, line 83: @shaomeng can you add some more detail to better explain the integration steps?

include/vapor/FlowParams.h
```diff
@@ -97,32 +156,48 @@ class PARAMS_API FlowParams : public RenderParams {
...
+    //! When randomly seeding flowlines with bias towards a along a chosen variable's distribution, this sets the bias strength.  Negative bias will place seeds at locations where the bias value has low values.  Positive bias will place seeds where the bias variable has high values.
+    //! \param[in] long - The bias of the seed distribution.
     void SetRakeBiasStrength(long);
 
+
     int  GetPastNumOfTimeSteps() const;
     void SetPastNumOfTimeSteps(int);
 
+    //! Returns the interval that new pathlines are injected into the scene.
+    //! \retval int - The seed injection interval.
```
clyne, line 196: What is the unit here,  Integration steps, vapor time steps?

include/vapor/FlowParams.h
```diff
@@ -171,35 +246,101 @@ class PARAMS_API FlowParams : public RenderParams {
...
+    //! When rendering samples, draw every N samples.
+    //! Applies data of type: int.
+    //! Typical values: 1 to 20.
+    //! Valid values: INT_MIN to INT_MAX.
     static const std::string RenderGlyphStrideTag;
-    //! When rendering samples, only draw the leading sample in a path
+
+    //! When rendering samples, only draw the leading sample in a path.
+    //! Applies data of type: bool.
+    //! Valid values: 0 = off, 1 = on.
```
clyne, line 283: Any reason why you've started using a different convention for specifying "Valid values" then earlier in this file? See, for example, GetSeedGenMode

include/vapor/FlowParams.h
shaomeng: I've added more explanation.

shaomeng (approved):
I've clarified a few things in FlowParams, and the rest of the document looks good!

include/vapor/FlowParams.h
```diff
@@ -53,35 +53,94 @@ class PARAMS_API FlowParams : public RenderParams {
...
+    //! you can scale the wind field with this parameter.  IE - If your data's domain is written in kilometers but your wind
+    //! vectors are in meters, you can apply a velocity multiplyer of 1000 to correct the mismatch.
+    //! \param[in] double - Velocity field multiplier for flow rendering
     void   SetVelocityMultiplier(double);
 
+    //! Get the number of steps to advect a steady flow line (aka a streamline)
+    //! \retval long - The number of steps a steady flow line is advected
     long GetSteadyNumOfSteps() const;
+    
+    //! Set the number of steps to advect a steady flow line (aka a streamline)
```
clyne:
```suggestion
    //! Set the number of integration steps to advect a steady flow line (aka a streamline). The integration step size is determined automatically and may vary over the integration
```

@shaomeng please verify that the above is correct.

include/vapor/FlowParams.h
```diff
@@ -53,35 +53,94 @@ class PARAMS_API FlowParams : public RenderParams {
...
+    //! Get the file name/path to a file containing a list of seed points to advect from.  See https://vapor.readthedocs.io/en/readthedocs/usage/flowRenderer.html#seed-distribution-settings
+    //! \retval string - A file path containing a defined list of seed points to advect from
     std::string GetSeedInputFilename() const;
-    void        SetSeedInputFilename(const std::string &);
+    
+    //! Set the file name/path to a file containing a list of seed points to advect from.  See https://vapor.readthedocs.io/en/readthedocs/usage/flowRenderer.html#seed-distribution-settings
+    //! \param[in] string - A file path containing a defined list of seed points to advect from
+    void SetSeedInputFilename(const std::string &);
 
+    //! If GetNeedFlowlineOutput() returns true, this will return the file path to the text file that data will be written to.
```
clyne, line 126: A small nit: won't this function return the output file name regardless of the value of GetNeedFlowlineOutput()?

include/vapor/FlowParams.h
```diff
@@ -171,35 +246,101 @@ class PARAMS_API FlowParams : public RenderParams {
...
     static const std::string RenderGlyphStrideTag;
-    //! When rendering samples, only draw the leading sample in a path
+
+    //! When rendering samples, only draw the leading sample in a path.
+    //! Applies data of type: bool.
+    //! Valid values: 0 = off, 1 = on.
     static const std::string RenderGlyphOnlyLeadingTag;
 
-    //! Falloff parameter for density rendering
+    //! Falloff parameter for density rendering.
```
clyne, line 291: @StasJ can you expand on this explanation? Not sure what "density rendering" means

include/vapor/FlowParams.h
```diff
@@ -171,35 +246,101 @@ class PARAMS_API FlowParams : public RenderParams {
...
+    //! Applies data of type: double.
+    //! Typical values: 0.5 to 10.0.
+    //! Valid values: DBL_MIN to DBL_MAX.
     static const std::string RenderDensityFalloffTag;
-    //! ToneMapping parameter for density rendering
+
+    //! ToneMapping parameter for density rendering.
```
clyne, line 297: @stasj ditto for this parameter. Not sure what tone mapping refers to

include/vapor/ImageParams.h
```diff
@@ -22,9 +22,8 @@ class PARAMS_API ImageParams : public RenderParams {
     //
     virtual size_t GetRenderDim() const override { return (2); }
 
-    //
-    // Get and set image file path
-    //
+    //! Set image file path
```
clyne, line 25: This is the path to the image file that will be read in and displayed, correct? Please clarify

include/vapor/ImageParams.h
```diff
@@ -36,19 +35,28 @@ class PARAMS_API ImageParams : public RenderParams {
...
 
-    //
-    // Get and set ifGeoRef
-    //
+    //! Inquire whether the currently selected image is georeferenced
+    //! Valid values: 0 = not georeferenced, 1 = georeferenced
+    //! \retval bool - State indicating whether current image is georeferenced
     bool GetIsGeoRef() const { return (GetValueLong(_isGeoRefTag, (long)true)); }
 
+    //! Set whether the currently selected image is georeferenced
```
clyne, line 47:
```suggestion
    //! If the raster image contained in the path returned by GetImagePath() has georeference information (e.g. the file is a GeoTIFF) this boolean determines whether or not the georeference information is honored
```

include/vapor/ImageParams.h
```diff
@@ -36,19 +35,28 @@ class PARAMS_API ImageParams : public RenderParams {
...
+    //! Valid values: 0 = not georeferenced, 1 = georeferenced
```
clyne, line 48:
```suggestion
    //! Valid values: 0 = do not use georeference information, 1 = use georeference information
```

include/vapor/ImageParams.h
```diff
@@ -36,19 +35,28 @@ class PARAMS_API ImageParams : public RenderParams {
...
     }
 
+    //! Set image file path
+    //! \retval string - Path to image file
     std::string GetImagePath() const;
 
-    //
-    // Get and set ifGeoRef
-    //
+    //! Inquire whether the currently selected image is georeferenced
```
clyne, line 42: See comments below for SetIsGeoRef

include/vapor/RenderParams.h
```diff
@@ -397,33 +395,48 @@ class PARAMS_API RenderParams : public ParamsBase {
     virtual bool GetOrientable() const;
 
     //! Return the renderer's 3 axis rotation for creating ArbitrarilyOrientedRegularGrids.
+    //! \retval vector<double> - Slice's rotation on X, Y, and Z axes
+    //! Valid values - -90.0 to 90.0 for each axis component
     vector<double> GetSlicePlaneRotation() const;
 
     //! Return the renderer's 3 axis origin for creating ArbitrarilyOrientedRegularGrids.
+    //! \retval vector<double> - Slice's origin on X, Y, and Z axes
```
clyne, line 403:
```suggestion
    //! \retval vector<double> - Slice's origin on X, Y, and Z axes, specified in the data's coordinate system
```

include/vapor/RenderParams.h
```diff
@@ -397,33 +395,48 @@ class PARAMS_API RenderParams : public ParamsBase {
...
+    //! Valid values - A point within the data domain
     vector<double> GetSlicePlaneOrigin() const;
 
     //! Return the renderer's 3 axis normal for creating ArbitrarilyOrientedRegularGrids.
+    //! \retval vector<double> - Slice's rotation on X, Y, and Z axes
```
clyne, line 408:
```suggestion
    //! \retval vector<double> - Slice's rotation on X, Y, and Z axes, specified in the data's coordinate system
```

include/vapor/RenderParams.h
```diff
@@ -397,33 +395,48 @@ class PARAMS_API RenderParams : public ParamsBase {
...
+    //! Valid values - A point within the data domain's X axis coordinates
     double GetXSlicePlaneOrigin() const;
 
     //! Return the renderer's origin value on the Y axis for creating ArbitrarilyOrientedRegularGrids.
+    //! \retval double - Slice's origin on the Y axis
+    //! Valid values - A point within the data domain's Y axis coordinates
     double GetYSlicePlaneOrigin() const;
 
     //! Return the renderer's origin value on the Z axis for creating ArbitrarilyOrientedRegularGrids.
+    //! \retval double - Slice's origin on the Z axis
```
clyne, line 423:
```suggestion
    //! \retval double - Slice's origin on the Z axis, specified in the data's coordinate system
```

include/vapor/RenderParams.h
```diff
@@ -484,46 +497,89 @@ class PARAMS_API RenderParams : public ParamsBase {
     static const string CustomHistogramRangeTag;
     static const string LightingEnabledTag;
 
-    //! If a renderer supports rotation about a point of origin,
-    //! this string identifies the parameter for the origin's
+    //! If a renderer supports rotation about a point of origin (IE - Slice and Contour),
+    //! this tag identifies the parameter for the origin's
```
clyne, line 501: Question for @stasj: how does the auto-python generation code handle parameters that don't have explicit setters/getters? I.e. does it generate Python setters and getters, or generate a generic Python function that accepts a tag? If the former, will this documentation make sense for Python users?

include/vapor/RenderParams.h
```diff
@@ -484,46 +497,89 @@ class PARAMS_API RenderParams : public ParamsBase {
...
+    //! If a renderer supports rotation about a point of origin (IE - Slice and Contour),
+    //! this tag identifies the parameter for the rotation
     //! about the Z axis.
+    //! Applies to data of type: double
+    //! Valid values: -90.0 to 90.0
     static const string ZSlicePlaneRotationTag;
 
-    //! If a renderer samples data points along a vector,
-    //! this string identifies the parameter for how many samples
+    //! If a renderer samples data points along a vector (IE - Slice and Contour),
```
clyne, line 542: Does sampling apply to the contour renderer? I thought that it operated on the native grid with an appropriate marching algo in all cases?

include/vapor/VolumeParams.h
```diff
@@ -25,30 +25,70 @@ class PARAMS_API VolumeParams : public RenderParams {
...
+    //! \param[in] string - Raytracing algorithm (Regular, Curvilinear, or Ospray)
     void        SetAlgorithm(std::string algorithm);
+
     void        SetAlgorithmByUser(std::string algorithm);
     bool        GetAlgorithmWasManuallySetByUser() const;
     void        SetAlgorithmWasManuallySetByUser(bool v);
-
     static std::vector<float> GetSamplingRateMultiples();
+
+    //! Get the sampling rate multiplier used with the current raytracing algorithm
```
clyne, line 41: Is this the sampling rate along the ray, or across the sampling plane? Please clarify. Also, should be "ray casting", not "ray tracing", correct?

clyne (commented):
@sgpearse this was a big effort. Thanks for doing this!

include/vapor/BarbParams.h
sgpearse: Fixed.  I don't know where voxels came from.  I was just copying the documentation from the getter

include/vapor/FlowParams.h
sgpearse: @clyne I don't think \copydoc is what we'd want here.  If SetIsSteady(bool) used used \copydoc on this getter, the first word "Gets" would be applied.  But it may be an option to add a \brief and \details section to these functions.  IE:

//! \brief Gets the type of flow rendering algorithm being used.
//! \details Steady flow (streamlines) renders time-invariant trajectories that follow a vector field at a single timestep.
//! Unsteady flow (pathlines) render time-variant trajectories that advect through the timeseries of a loaded dataset.
bool GetIsSteady() const;

//! \brief Sets the type of flow rendering algorithm being used.
//! \copydetails FlowParams::GetIsSteady()
void SetIsSteady(bool);

include/vapor/FlowParams.h
sgpearse: I felt that I should add more detail to the tags because they're just strings, and the reader can't infer what data type they correspond to.  Would you like me to add the "Valid values" and "Typical values" information to the setters and getters as well?

include/vapor/FlowParams.h
clyne: That makes sense. Maybe \sa (see also) would be an even better choice than \copydoc?

include/vapor/FlowParams.h
shaomeng: It is correct, but I feel that SetSteadyNumOfSteps() has a better explanation on this topic. I was expecting that one of the two has a more detailed documentation, which is the setter in this case, and the other one adopts \seealso or \copydoc.

include/vapor/FlowParams.h
sgpearse: Yes I suppose it does.  Fixed.

shaomeng: It is correct. I've clarified it.

include/vapor/ImageParams.h
sgpearse: Clarified.

include/vapor/ImageParams.h
sgpearse: Got it.  Fixed.

include/vapor/RenderParams.h
sgpearse: The sampling rate applies to the *deep breath* ArbitrarilyOrientedRegularGrid class, so I believe the contour renderer uses it too.  @Stasj?

include/vapor/VolumeParams.h
sgpearse: > Is this the sampling rate along the ray, or across the sampling plane?

I'm not sure.  @Stasj?

> Also, should be "ray casting", not "ray tracing", correct?

Yes, that is correct.  Fixed.

include/vapor/FlowParams.h
```diff
@@ -196,7 +200,8 @@ class PARAMS_API FlowParams : public RenderParams {
     //! \retval int - The seed injection interval.
     int  GetSeedInjInterval() const;
     
-    //! Sets the interval that new pathlines are injected into the scene.
+    //! Sets the interval w.r.t. the time steps that new pathlines are injected into the scene.
+    //! For example, 1 means that seeds are injected at every time step, and 2 means that seeds are injected at every other time step.
```
clyne, line 204:
```suggestion
    //! For example, 1 means that seeds are injected at every time step, and 2 means that seeds are injected at every other time step. Note  "time step" refers to the data set time step, not the integration time step
```

include/vapor/FlowParams.h
clyne: OK. Would like to see how the documentation appears in the two cases: functions for which there is a C++ setter/getter, and functions for which there are not C++ setter/getters.

include/vapor/RenderParams.h
StasJ: That is correct.

clyne: Maybe the caveat here is that this only applies when 3D data are sliced (for contouring or pseudo coloring)

include/vapor/RenderParams.h
StasJ: It generates python setters and getters. The setters/getters will include the tag documentation as well as specify the data type that the tag expects. We could simply omit referring to it as a "tag" if that is confusing.

clyne: I think it is more important for the documentation to be clear for the Python users. The GUI users get a lot of information by context. So, yes, I agree omitting tag would be better. E.g. something like:

"This parameter identifies the origin's location..."

would probably work for both GUI and Python users.

include/vapor/FlowParams.h
StasJ: Both of these are somewhat arbitrary parameters to the somewhat arbitrary equation used by the density paper. It's not really a rigorous algorithm, the user is intended to fiddle with them until they get something that looks good. So I don't have much to add unless it's just to try different values and see what looks good. This is easier in the GUI where you have sliders.

include/vapor/FlowParams.h
StasJ: Both of these are somewhat arbitrary parameters to the somewhat arbitrary equation used by the density paper. It's not really a rigorous algorithm, the user is intended to fiddle with them until they get something that looks good. So I don't have much to add unless it's just to try different values and see what looks good. This is easier in the GUI where you have sliders.

include/vapor/VolumeParams.h
StasJ: The sampling rate is along the ray.

include/vapor/FlowParams.h
clyne: How about including a reference to the paper and being more clear that these parameters only apply when flow render type is "density".

@sgpearse or @stasj, there are three different flow render modes (stream, particles, density). Is the API for selecting these modes documented? I don't see anything.

StasJ: I've clarified and included a reference to the paper.

Yes, it is documented under the `RenderTypeTag` which generates `SetRenderType()` and `GetRenderType()` in the python API.

include/vapor/FlowParams.h
StasJ: I've clarified and included a reference to the paper.

StasJ (commented):
> The only caveat that I haven't figured out regards the documentation for SliceParams, which does not transfer into the python_api help() menu. Copied from the doc:

>> If a Params class has only inherited functions defined in its FuncWrapperStrList in renderer.py, no documentation will be built for those functions?

>> IE - the Python class SliceRenderer only has three inherited functions from RenderParams.h in its FuncWrapperStrList (GetSlicePlaneRotation, GetSlicePlaneOrigin, and GetSlicePlaneNormal). It does not populate documentation for the inherited RenderParams

This is due to the fact that DOxygen does not propagate documentation for inherited members. I am working on a workaround.

include/vapor/FlowParams.h
sgpearse: Hmm I'm not sure how \sa would work here.  Check out FlowParams::Get/SetSteadyNumOfSteps().  I think \details and \copydetails does what we're looking for.

clyne: Agreed!

include/vapor/RenderParams.h
sgpearse: Ok, I've added the caveat you've outlined to the slice/contour tags with \details and \copydetails.

clyne (approved):

clyne:
> 

This issue is being addressed by #3175, correct? @StasJ are there any other outstanding comments on this PR? I.e. can we approve and merge?

StasJ (approved):
> This issue is being addressed by https://github.com/NCAR/VAPOR/pull/3175, correct? @StasJ are there any other outstanding comments on this PR? I.e. can we approve and merge?

Correct. Approved.

clyne:
Hooray!

PR #3175, opened 2022-08-11
Query parent classes for DOxygen docs

StasJ:
Fixes SliceParams documentation in #3168

### DIFF
--- a/apps/pythonapi/vapor/cppyyDoxygenWrapper.py
+++ b/apps/pythonapi/vapor/cppyyDoxygenWrapper.py
@@ -126,12 +126,15 @@ class CPPYYDoxygenWrapperMeta(type):
 """.strip()
 
     @classmethod
-    def __GetFunctionDocumentation(cls, pCls, func) -> str:
+    def __GetFunctionDocumentation(cls, pCls:type, func) -> str:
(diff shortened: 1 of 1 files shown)


sgpearse:
@stasj this works, thank you for the fix.

One question: 
I don't see "tag" functions in the RenderParams base class help menu, such as GetSliceOffset().  Granted, it's only relevant for the slice/contour renderers.  But is there a reason why these functions aren't being documented in the RenderParams base class?

StasJ:
> @StasJ this works, thank you for the fix.
> 
> One question: I don't see "tag" functions in the RenderParams base class help menu, such as GetSliceOffset(). Granted, it's only relevant for the slice/contour renderers. But is there a reason why these functions aren't being documented in the RenderParams base class?

Like you said, it's only relevant for the slice/contour renderers.

sgpearse:
Can these be put in the RenderParams help menu?  This would be incredibly helpful for my understanding.

StasJ:
> Can these tagged functions be put in the RenderParams help menu? This would be incredibly helpful for my understanding.

They can but that would result in, for example, the volume renderer having a NOP `GetSliceOffset()` method.

sgpearse (approved):

Issue #3178, opened 2022-08-11
Windows build currently broken


sgpearse:
PR #3152 takes some steps towards building on windows with BUILD_PYTHON=ON, but when it's off our build still fails due to ~osgl~ glad linker problems.

StasJ:
@sgpearse The windows build fails with issues unrelated to OSGL even in the branch from #3152. Could you please point me to the branch where windows is failing due to OSGL linker problems so I don't have to fix the other issues a second time?

sgpearse:
Right - I know what you mean.  The branch I've been working on is _[winPythonAPI2](https://github.com/NCAR/VAPOR/tree/winPythonAPI2)_, which is referenced in PR #3152.  Not sure if all of the changes in that PR are valid though.

PR #3180, opened 2022-08-16
Fixed #3139  - Windows crash on UGRID

### LINKED ISSUE #3139: Windows crash on UGRID data
To reproduce:
1) On windows, load our example [UGRID dataset](https://vapor.readthedocs.io/en/readthedocs/downloads/downloads.html#sample-data).

On my machine, Vapor will crash.  User mbotto has reported the following error:

```
Failure to open data set of type "ugrid"
Failed to initialize data importer
Failed to initialize netCDF data collection for reading
NetCDFSimple::Initialize(---/smallGrid.nc)
nc_get_att_long(327680, 1, _FillValue) : NetCDF: Numeric conversion not representable
------------------
Vapor 3.6.0.<debug>
OS: Windows 6.2 (9200)
```

clyne:
Fixed #3139

### DIFF
--- a/lib/vdc/NetCDFSimple.cpp
+++ b/lib/vdc/NetCDFSimple.cpp
@@ -411,15 +411,6 @@ int NetCDFSimple::_GetAtts(int ncid, int varid, vector<pair<string, vector<doubl
         return (-1);
     }
 
-    double *dblbuf = NULL;
-    size_t  dblbufsz = 0;
-
-    long * longbuf = NULL;
-    size_t longbufsz = 0;
-
-    char * textbuf = NULL;
-    size_t textbufsz = 0;
-
     for (int i = 0; i < natts; i++) {
         char namebuf[NC_MAX_NAME + 1];
         rc = nc_inq_attname(ncid, varid, i, namebuf);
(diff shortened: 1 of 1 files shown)


shaomeng (approved):
Is the nature of this bug being integers having different lengths on different platforms? If so, would it be more robust and also the code easier to understand if using fixed length integers, e.g., `int64_t` or `uint64_t`?

clyne:
> Is the nature of this bug being integers having different lengths on different platforms? If so, would it be more robust and also the code easier to understand if using fixed length integers, e.g., `int64_t` or `uint64_t`?

Agreed. But that is not what the NetCDF API supports :-(

shaomeng:
> > Is the nature of this bug being integers having different lengths on different platforms? If so, would it be more robust and also the code easier to understand if using fixed length integers, e.g., `int64_t` or `uint64_t`?
> 
> Agreed. But that is not what the NetCDF API supports :-(

Bummer...

sgpearse:
Code looks good but I'm unable to validate due to #3178, broken windows build.  We'll get around to in during bugfix testing though.

sgpearse (approved):

PR #3184, opened 2022-08-17
Fixed #3182 - Infinite loop with clang-13

### LINKED ISSUE #3182: Infinite loop with clang-13
The migration to clang-13 has turned up an infinite loop in the quadtree code on Mac OSX. 

To reproduce:

1. Attempt to load the MPAS files MPAS/FalkoJudt/dyamond_1/30km/*

VAPOR will enter an infinite loop.

clyne:
Fixed divide by zero in quadtree code.

Fixed #3182

### DIFF
--- a/include/vapor/QuadTreeRectangle.hpp
+++ b/include/vapor/QuadTreeRectangle.hpp
@@ -2,6 +2,7 @@
 
 #include <vector>
 #include <cassert>
+#include <limits>
 #include <cmath>
 #include <iostream>
 #include <cstdint>
@@ -67,7 +68,10 @@ public:
 
         // Horizontal aspect ratio
         //
-        float hAspectRatio() const { return (std::fabs((_right - _left) / (_bottom - _top))); }
+        float hAspectRatio() const {
+            if (_bottom == _top) return (std::numeric_limits<float>::quiet_NaN());
+            return (std::fabs((_right - _left) / (_bottom - _top)));
+        }
 
         friend std::ostream &operator<<(std::ostream &os, const rectangle_t &rec)
         {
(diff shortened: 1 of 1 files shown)


shaomeng:
It seems in C/C++ standard, divide by zero is "undefined behavior," or UB. One should not rely on an UB to return a particular value. It seems to me a more proper fix is to change the `rectangle.hAspectRatio()` function implementation so that it returns a special value (e.g., Infinity, NaN) when divide by zero would occur.

clyne:
Wouldn't that change the mathematical definition of `ratio` between A and B, which is A/B? The ratio of A:B should be undefined if B is zero, but it should be zero if A is zero and B is not zero. A horizontal line has zero slope, not 'no slope'.

shaomeng:
I agree that zero slop and "no slop" are two different things. I was suggesting that for the following line:
```c++
float hAspectRatio() const { return (std::fabs((_right - _left) / (_bottom - _top))); }
```
change it to be something like this:
```c++
float hAspectRatio() const { 
  if (_bottom == _top)
    return NaN;  // or any other special value
  else
    return (std::fabs((_right - _left) / (_bottom - _top))); }
```
and every caller needs to detect if the return value is `NaN` or not.

clyne:
Interesting. I've made the change, but note even cppreference.com [states](https://en.cppreference.com/w/cpp/types/numeric_limits/quiet_NaN) that a NaN can be generated by divide by zero.

shaomeng:
Yeah, it might be a common behavior, but there really isn't any guarantee of what comes out of a UB. In fact, the title of this PR suggests that clang-13 has a behavioral change.

Re: divide by zero: the [C standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) has made it clear that divide by zero is an UB (in Section 6.5.5). In practice, if an UB condition has a potential to occur (i.e., `_bottom == _top` could come true), it should have its own special handling instead of relying on the compiler to behave in a certain fashion.

shaomeng (approved):

PR #3187, opened 2022-08-19
Add conda build check on CircleCI

### LINKED ISSUE #3133: Employ CI build and test systems for Python API
A CI build/test system would
1) Document the steps for building the Python API
2) Add checks to see if PR's break the API
3) May allow for more comprehensive unit testing (no graphics cards on Circle nodes though...)

sgpearse:
Fixes #3133, by adding a compilation check for the conda .tar.bz2 package.  This tar.bz2 file, and all intermediate compilation files are stored as artifacts on CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -334,6 +334,52 @@ jobs:
             https://api.github.com/repos/NCAR/VAPOR/issues/${CIRCLE_PULL_REQUEST##*/} \
             -d "$json"
 
+  build_python_api2:
+    docker:
+      - image: sgpearse/vapor3-centos7:latest
+
+    steps:
+      - run:
+          name: install python
+          command: |
+            yum update
+            yum install -y python3
+            yum install -y python3-pip
+            pip3 install --upgrade pip
+            pip3 install scikit-build command
+
+      - run:
+          name: update cmake
+          command: |
+            #apt remove -y --purge --auto-remove cmake
+            #apt install -y libssl-dev
+            #apt install -y build-essential git
+            #git clone https://github.com/Kitware/CMake/
+            #cd CMake
+            #./bootstrap && make && make install
+
+      - run:
+          name: install miniconda
+          command: |
+            yum install -y wget
+            wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh
+            bash ~/miniconda.sh -b -p ~/miniconda 
+            rm ~/miniconda.sh
+            export PATH=~/miniconda/bin:$PATH
+            source ~/.bashrc
+
+      - checkout
+
+      - run:
+          name: conda build .
+          command: |
+            rm -rf /root/project/build
+            cd /root/project/conda
+            /root/miniconda/bin/conda install -y conda-build
+            /root/miniconda/bin/conda config --add channels conda-forge
+            nohup /root/miniconda/bin/conda build . &
+          no_output_timeout: 60m
+ 
   build_python_api:
     docker:
       - image: sgpearse/vapor3-ubuntu18:latest
@@ -684,6 +730,7 @@ workflows:
   build:
     jobs:
       - build_python_api
+      - build_python_api2
       #- clang-tidy
       #- build_ubuntu18
       #- build_centos7
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -375,9 +375,9 @@ jobs:
           command: |
             rm -rf /root/project/build
             cd /root/project/conda
-            conda install -y conda-build
-            conda config --add channels conda-forge
-            conda build .
+            /root/miniconda/bin/conda install -y conda-build
+            /root/miniconda/bin/conda config --add channels conda-forge
+            /root/miniconda/bin/conda build .
  
   build_ubuntu18:
     docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -335,10 +334,63 @@ jobs:
...
-  build_ubuntu18:
+  python_api_centos:
+    docker:
+      - image: sgpearse/vapor3-centos7:latest
+
+    resource_class: large
+
+    steps:
+      - run:
+          name: update 3rd party libraries
```
StasJ, line 345: The conda build does not use the third party libraries.

.circleci/config.yml
```diff
@@ -335,10 +334,63 @@ jobs:
...
+          command: |
+            rm -rf /usr/local/VAPOR-Deps/2019-Aug
+            pip3 install gdown --upgrade
+            gdown https://drive.google.com/uc?id=1S9DwySMnQrBuUUZGKolD__WQrjTmLgyn
+            filename="/root/project/2019-Aug-CentOS.tar.xz"
+            bsdtar -xf ${filename} -C /usr/local/VAPOR-Deps
+
+      - run:
+          name: install python
```
StasJ, line 354: Python is provided by conda.

.circleci/config.yml
```diff
@@ -335,10 +334,63 @@ jobs:
...
+          command: |
+            yum -y update || true
+            yum install -y python3
+            yum install -y python3-pip
+            pip3 install --upgrade pip
+            pip3 install scikit-build command
+
+      - run:
+          name: install miniconda
```
StasJ, line 363: You could just use the miniconda docker image.

.circleci/config.yml
```diff
@@ -335,10 +334,63 @@ jobs:
...
+            yum install -y wget
+            wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh
+            bash ~/miniconda.sh -b -p ~/miniconda 
+
+      - checkout
+
+      - run:
+          name: conda build .
+          command: |
+            yum -y install doxygen
```
StasJ, line 374: DOxygen is provided by conda.

.circleci/config.yml
```diff
@@ -335,10 +334,63 @@ jobs:
...
+            rm -rf /root/project/build/*
+            rm -f /root/project/site.local
+            cd /root/project/build
+            cd /root/project/conda
+            export PYTHONPATH=/root/miniconda/lib/python3.9
```
StasJ, line 379: This is non-portable with regards to conda and is handled by the build anyway.

.circleci/config.yml
```diff
@@ -335,10 +334,63 @@ jobs:
...
+            /root/miniconda/bin/conda install -y conda-build
+            /root/miniconda/bin/conda config --add channels conda-forge
+            /root/miniconda/bin/conda build .
+          no_output_timeout: 20m
+ 
+      - store_artifacts:
+          path: /root/miniconda/conda-bld/linux-64
```
StasJ, line 386: We only need the tarball.

.circleci/config.yml
```diff
@@ -335,10 +334,63 @@ jobs:
...
+
+  python_api_ubuntu:
```
StasJ, line 388: The conda build generates a universal linux binary.

.circleci/config.yml
```diff
@@ -350,15 +402,55 @@ jobs:
             pip3 install scikit-build command
 
       - run:
-          name: update 3rd party libraries
+          name: update cmake
```
StasJ, line 405: cmake is provided by conda.

StasJ (changes requested):

.circleci/config.yml
sgpearse: I didn't know they had one based on debian.  Sounds good.

.circleci/config.yml
sgpearse: Good to know.

.circleci/config.yml
sgpearse: CircleCI is finicky about how it uploads artifacts, in that wildcards can't be used to select files.  Only directories can be targeted.  I'll just move the tarball to an empty directory with the wildcard, and then target that directory.

.circleci/config.yml
sgpearse: Would you agree that it's still beneficial to have documented build rules for each platform?

.circleci/config.yml
sgpearse: Got it.

.circleci/config.yml
sgpearse: Fixed.

sgpearse:
@stasj - Thanks for the recommendations.  They simplify things quite a bit.  

Everything should be addressed except for the comment regarding the linux builds generating the same installer.  I'm thinking that it would be good to have build rules for all *nix platforms so we can trigger one-off builds if we need an installer.  I've added a new job for OSX (python_api_osx) for this.  The centOS and OSX builds are currently turned off, but you can see their runs succeed in commit [c112cac](https://github.com/NCAR/VAPOR/pull/3187/commits/c112cac926c6f0d64d29fb37c3311d983076530b).

.circleci/config.yml
sgpearse: Huh I thought I commented here.  Fixed.

PR #3197, opened 2022-08-25
Add -f (force) flag to gunzip in build.sh to faciliate rerunning buil…

clyne:
Add -f (force) flag to gunzip in build.sh to facilitate rerunning build.sh (if you try to run build.sh twice in a row the second instance will block on unzip).

### DIFF
--- a/conda/build.sh
+++ b/conda/build.sh
@@ -48,7 +48,7 @@ unset CMAKE_PREFIX_PATH
 SP_DIR="`python -c 'import site; print(site.getsitepackages()[0].replace(\"'$BUILD_PREFIX'\", \"'$PREFIX'\"))'`"
 
 # Our third party libs have a non-standard copy of the GTE library so it is packaged and extracted here.
-unzip -d include buildutils/GTE.zip
+unzip -f -d include buildutils/GTE.zip
 
 if false ; then
 	echo "============================================================"


StasJ (approved):
`build.sh` is intended to be run by conda-build which generates a clean build environment on every run so this issue shouldn’t arise in a standard conda build, but it can’t hurt.

clyne:
Maybe I'm not running conda build correctly. Please the instructions I added [here](https://docs.google.com/document/d/1RvbnJ_LbWvlUZzgjeutLQKwJV_M0blEHFSxsgSB7wRQ/edit?usp=sharing)
under "Building the conda image" and correct anything that looks wrong. Thanks!

StasJ:
> Maybe I'm not running conda build correctly. Please the instructions I added [here](https://docs.google.com/document/d/1RvbnJ_LbWvlUZzgjeutLQKwJV_M0blEHFSxsgSB7wRQ/edit?usp=sharing)
> under "Building the conda image" and correct anything that looks wrong. Thanks!

The instructions are correct. It is possible you had run build.sh locally at some point before in which case the extracted files would already be present in the vapor source tree.

clyne:
Gotcha. As you said, it won't hurt to include this. I seem to have run into this problem a couple of times, but I'm prone to repeating the same mistakes :-). Merging.

sgpearse:
@clyne - This branch didn't incorporate the[ python_api build check](https://github.com/NCAR/VAPOR/pull/3187) that was merged only a few hours ago.  Unfortunately this is causing our CircleCI tests to fail.  I'll try to fix this tomorrow.

sgpearse:
@clyne - The build script makes a call to `unzip`, not `gunzip`.  For `unzip`, the -f flag does as follows:

`-f  freshen existing files, create none`

I think you may have wanted to use one of the following:

`-u  update files, create if necessary`
`-o  overwrite files WITHOUT prompting`

However, I'm unable to reproduce the error you are reporting in this PR, where the build will fail if `conda build .` is called twice in succession.  [This CircleCI pipeline](https://app.circleci.com/pipelines/github/NCAR/VAPOR/5990/workflows/71736ced-2caa-405d-9242-417b0ba5aba0/jobs/14307) is making that double-call on osx, ubuntu, and centos and passes.

The PR for the python test framework is currently not running tests because it hasn't merged with main yet.  But if it did, it would fail due to the addition of the -f flag in unzip.  I am going to revert this so the test PR can merge and pass health checks.  But I'm not sure what to do about the issue you're encountering because I can't reproduce it.  Thoughts?

PR #3199, opened 2022-08-26
Issue 3167: Missing values that are NaNs are not supported

### LINKED ISSUE #3167: Missing values that are NaNs are not supported
It is possible to use NaNs as a missing value. Hence, Grid::GetMissingValue() may return a NaN. Much of the code in VAPOR that handles missing values checks for the existence of missing values in the data with something like:

`if (g->GetMissingValue() == g->GetValue())`

However, in C/C++ comparing a NaN to a NaN with an equality operator always results in False. Thus special case handling should be added throughout the code base to deal with NaN missing values. E.g.

`if ((g->GetMissingValue() == g->GetValue())) || (std::isnan(g->GetMissingValue())  && std::isnan(g->GetValue()))`

See the test data set used in #3087

clyne:
Fixed #3167 

This PR fixes #3167. However, the fix in this PR is internally replace NaN's with std::infinity. This is a much simpler fix than proposed in #3167, only touching two files, but may not prove to be as robust.

Note, this PR also fixes an unreported bug: The NetCDFCFCollection class did not handle the case where a single NetCDF
file used by the "_FillValue" and "missing_value" attribute name to indicate missing values.


sgpearse:
@clyne - Can we make this a normal PR instead of a draft at this point?

clyne:
> @clyne - Can we make this a normal PR instead of a draft at this point?

Indeed. Thanks for catching. Done.

lib/vdc/GridHelper.cpp
```diff
@@ -606,6 +606,11 @@ void GridHelper::_makeGridHelper(const DC::DataVar &var, const DimsType &roi_dim
         mv = var.GetMissingValue();
     }
 
+    // Don't allow NaNs as missing values. We replace any
+    // missing values in the data with infinity
+    //
+    if (std::isnan(mv)) mv = std::numeric_limits<float>::infinity();
```
shaomeng, line 612: I assume there won't be incoming data sets with double type and also containing missing values? 

Also, NetCDFCFCollection::_GetMissingValue() seems to represent mv as a double. So maybe there needs some consistency improvement, or better, complete coverage improvement (i.e., properly handle both float and double types)?

lib/vdc/NetCDFCFCollection.cpp
```diff
@@ -826,55 +824,24 @@ bool NetCDFCFCollection::_GetMissingValue(string varname, string attname, double
...
+    // for either, but give priority to "missing_value"
     //
     for (int d = 1; d < 5; d++) {
         vector<string> vars = NetCDFCFCollection::GetDataVariableNames(d, false);
 
         for (int i = 0; i < vars.size(); i++) {
             double mv;
-            if (_GetMissingValue(vars[i], mv_attname, mv)) { missingValueMap[vars[i]] = mv; }
+            if (_GetMissingValue(vars[i], "missing_value", mv)) { missingValueMap[vars[i]] = mv; }
+            if (_GetMissingValue(vars[i], "_FillValue", mv)) { missingValueMap[vars[i]] = mv; }
```
shaomeng, line 844: if "missing_value" has a higher priority as the comment says, shouldn't these two lines reverse order?

sgpearse (changes requested):
The fix works and I thought the code looked good, but Sam's raising a couple points that we should agree on.

lib/vdc/GridHelper.cpp
clyne: > I assume there won't be incoming data sets with double type and also containing missing values?

That is correct. The Grid class is not templated; data values are always single precision floats.

> 
> Also, NetCDFCFCollection::_GetMissingValue() seems to represent mv as a double. So maybe there needs some consistency improvement, or better, complete coverage improvement (i.e., properly handle both float and double types)?

Agreed. But I'm not sure that it is in scope for this fix.

lib/vdc/NetCDFCFCollection.cpp
clyne: You are correct. Fixed. Thanks!

sgpearse (approved):

shaomeng (approved):

PR #3217, opened 2022-09-01
fix warnings on macos

shaomeng:
This PR resolves a few compiler warnings on macos.


StasJ:
`_screenQuadVAO` and `_screenQuadVBO` are handles. They were set to `NULL` rather than `0` to convey this fact.

clyne:
@stasj you should have been a reviewer on this. My bad. We can undo/revised if needed.

PR #3224, opened 2022-09-09
Fix #3223

### LINKED ISSUE #3223: Flow GUI and params database inconsistency
Steps:
1. Insert a print statement to print the retrieved value in this line:
https://github.com/NCAR/VAPOR/blob/dc53d18f80860753c33dddc0809be22b6cea9a8c/lib/render/FlowRenderer.cpp#L884
1. Import one step of Katrina
3. Create a 2D flow renderer with vars x:U10 y:V10
4. Set the seed distribution to random
5. Enable the renderer, tt will work as expected
6. Change to random with bias. Notice that in the GUI, the bias variable is CANWAT. However, the print out message says it's P.

StasJ:


sgpearse (commented):
Quoted from the discussion on #3223:

> Thanks Stas, I really appreciate it.
> 
> I'm not sure if this fixes the problem though. If we follow the use case that Sam wrote, the user is now allowed to use the 3D variable P as a bias variable when performing 2D flow, which will break. I think you know this though as you stated above.
> 
> > I don't believe it should be the GUI's job to fix the params database.
> 
> Do you think that FlowParams should be setting its own defaults then? I can't remember if it has a notion of whether it's operating on 2D vs 3D variables, but if it does, maybe this is what's needed.

sgpearse:
@stasj I'm unclear on why we would want this to happen.  The PVariableSelector is populating the VComboBox with values from `DataMgr::GetDataVarNames(int nDim, VarType type)`.  These are the only values that should be valid.

With this PR, the VComboBox will have options that were not retrieved from the DataMgr.  If we follow Sam's steps listed in #3223, the 3D variable P will be included among the rest of the 2D variables.  But we should only be able to select 2D bias variables for 2D flow.

Can you explain the use case for giving the user options outside of variables returned by `DataMgr::GetDataVarNames(int nDim, VarType type)`?  I'll note that this PR affects all VComboBoxes in the GUI, not just our variable selectors.  If accepted, we may need to perform GUI regression tests.

StasJ:
> @StasJ I'm unclear on why we would want this to happen. The PVariableSelector is populating the VComboBox with values from `DataMgr::GetDataVarNames(int nDim, VarType type)`. These are the only values that should be valid.
> 
> With this PR, the VComboBox will have options that were not retrieved from the DataMgr. If we follow Sam's steps listed in #3223, the 3D variable P will be included among the rest of the 2D variables. But we should only be able to select 2D bias variables for 2D flow.
> 
> Can you explain the use case for giving the user options outside of variables returned by `DataMgr::GetDataVarNames(int nDim, VarType type)`? I'll note that this PR affects all VComboBoxes in the GUI, not just our variable selectors. If accepted, we may need to perform GUI regression tests.

What would you suggest as an alternative correct behavior?

sgpearse:
> What would you suggest as an alternative correct behavior?

I think PR #3226 does the trick, but you mention the following:

> There are two separate bugs here. My PR fixes a bug in the GUI while this PR fixes a bug in the Flow params.

What's the other bug in the GUI?  As far as I can tell, this bug is fixed if we set better default values when switching between 2D and 3D rendering.

StasJ:
> > What would you suggest as an alternative correct behavior?
> 
> I think PR #3226 does the trick, but you mention the following:
> 
> > There are two separate bugs here. My PR fixes a bug in the GUI while this PR fixes a bug in the Flow params.
> 
> What's the other bug in the GUI? As far as I can tell, this bug is fixed if we set better default values when switching between 2D and 3D rendering.

The bug your PR fixes is that the Flow params does not properly set its default variables. The unrelated bug that this PR fixes is that the VComboBox ignores calls to `SetValue` if the value is not already in its options.

sgpearse (approved):

PR #3225, opened 2022-09-12
Fix #3216 - Broken directional particles

### LINKED ISSUE #3216: Particle Renderer length scale gives incorrect results
1. Open the example DCP dataset
2. Create/enable a particle renderer and set the field variables
3. Enable `Show Direction`
4. Change `Length Scale`

Notice as you change the length scale, the number of particles as well as the direction of the existing particles change.

rec

sgpearse:
Fixes #3216 which found that the particles change direction when their length was adjusted.  This also combines the functions _generateTextureData() and ~_generateTextureData()~ _generateParticleData() to avoid unnecessarily iterating over all particle positions twice.

### DIFF
--- a/lib/render/ParticleRenderer.cpp
+++ b/lib/render/ParticleRenderer.cpp
@@ -408,7 +408,7 @@ void ParticleRenderer::_generateTextureData(const Grid* grid, const std::vector<
 
     glBindVertexArray(_VAO);
     glBindBuffer(GL_ARRAY_BUFFER, _VBO);
-    glBufferData(GL_ARRAY_BUFFER, sizeof(_vertex) * _particles.size(), _particles.data(), GL_STREAM_DRAW);
+    glBufferData(GL_ARRAY_BUFFER, sizeof(Vertex) * _particles.size(), _particles.data(), GL_STREAM_DRAW);
     glBindBuffer(GL_ARRAY_BUFFER, 0);
 }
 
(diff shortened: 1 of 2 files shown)


lib/render/ParticleRenderer.cpp
```diff
@@ -426,21 +375,41 @@ void ParticleRenderer::_generateParticleData(const Grid* grid, const std::vector
...
+            _particles.push_back({prep, value});
+        }
     }
+    
+    assert(glIsVertexArray(_VAO) == GL_TRUE);
+    assert(glIsBuffer(_VBO) == GL_TRUE);
+
+    glBindVertexArray(_VAO);
+    glBindBuffer(GL_ARRAY_BUFFER, _VBO);
+    glBufferData(GL_ARRAY_BUFFER, sizeof(_vertex) * _particles.size(), _particles.data(), GL_STREAM_DRAW);
```
StasJ, line 411: The struct `_vertex` follows the naming convention of a private member variable which was confusing when reviewing.

StasJ (commented):
Looks good, just one comment on naming.

lib/render/ParticleRenderer.cpp
sgpearse: Hmm.  I thought we put underscores on all private members, even definitions.  I've capitalized it to be _Vertex_ instead, which seems like what is being done in the Volume renderer, if we'd like to make that our policy.

StasJ (approved):

PR #3226, opened 2022-09-12
Alternative fix to #3223

sgpearse:
Alternative fix to PR #3223, which sets a new default rake bias variable whenever the dimensionality is changed in the flow renderer.


StasJ (approved):
There are two separate bugs here. My PR fixes a bug in the GUI while this PR fixes a bug in the Flow params.

shaomeng (approved):

PR #3231, opened 2022-09-19
Allow for saving and loading of the current camera parameters to a file

sgpearse:
This isn't intended to be included in 3.7 but it's something that I've been needing for visualization work, so I'll mark this as postponed.

> If we are working on multiple datasets that have the same domain and we want to use the same camera position among them, we need to manually enter 16 values into Vapor.  Currently I'm have a dataset like this that is run at four different resolutions.  Each time I need to change the camera perspective, I change it on one session then manually enter those values into the other three.  This means 48 manual entries per camera change, which is tedious prone to error.

This PR allows the user to save and load the camera state as an .xml file across different sessions from the Scene/Viewpoint tab.

Note: The PButtons that save or load a camera file make callbacks to ViewpointParams functions that may fail if they cannot open the targeted file.  Upon failure, we report the error, but I don't know how to capture the return code from the lambda callback.  @stasj do you have any thoughts on this?

Screen Shot 2022-09-19 at 2 56 06 PM

### DIFF
--- a/apps/vaporgui/CMakeLists.txt
+++ b/apps/vaporgui/CMakeLists.txt
@@ -180,6 +180,8 @@ set (SRCS
 	PStringDropdownHLI.h
 	PColorSelector.cpp
 	PColorSelector.h
+	PFileButton.cpp
+	PFileButton.h
 	PFileSelector.cpp
 	PFileSelector.h
 	PFileSelectorHLI.h
--- a/apps/vaporgui/PButton.h
+++ b/apps/vaporgui/PButton.h
@@ -5,9 +5,6 @@
 
 class VPushButton;
 
-typedef std::function<void(VAPoR::ParamsBase *)> Callback;
-typedef std::function<void(std::string)> Callback2;
-
 //! \class PButton
 //! \brief PWidget wrapper for VPushButton.
 //! \author Stas Jaroszynski
@@ -16,7 +13,9 @@ typedef std::function<void(std::string)> Callback2;
 //! Please don't capture in the callback.
 
 class PButton : public PWidget {
+    typedef std::function<void(VAPoR::ParamsBase *)> Callback;
     VPushButton *                                    _button;
+    const Callback                                   _cb;
     bool                                             _disableUndo = false;
 
 public:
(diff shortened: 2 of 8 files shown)


clyne:
@sgpearse I'm not sure I understand what you are trying to accomplish here. Are you running more than one VAPOR instance (unix process) that you are trying to share information between, or do you have a single VAPOR instance and you are trying to maintain the same viewpoint when you enable visualizers with different data sets?

sgpearse:
> Are you running more than one VAPOR instance (unix process) that you are trying to share information between

Yes, I have four sessions that reference four different datasets, but they share the same domain.  I need to share camera perspectives among them without manually entering the camera position, direction, up vector, and origin.

clyne:
Still struggling with the terminology. When you say "session" are you referring to a unix process or a VAPOR session (as in VAPOR session file)? If the former, you should be able to start multiple VAPOR instances (unix processes), load the same session file into each, and load different data sets. If the latter, don't the camera viewpoints get shared between data sets?

Maybe you need to do a live demo. It would be good to have a better understanding of the problem you are trying to solve.

sgpearse:
> Still struggling with the terminology. When you say "session" are you referring to a unix process or a VAPOR session (as in VAPOR session file)?

By "session" I'm referring to our .vs3 xml files.  I have four of these that reference unique datasets, and I find camera perspectives that I like with the gui for later use with the python api.  Each .vs3 session needs to share the same camera perspective, and this requires me to manually copy and paste values between sessions.  Trying to get around that with this PR.

> Maybe you need to do a live demo. It would be good to have a better understanding of the problem you are trying to solve.

Sure, if we have time during our meeting tomorrow I can demo, or whenever it's convenient.  This shouldn't take priority over 3.7 and it lets me do what I need for vis work, so we can push this back until later.

sgpearse:
Hey @stasj - I've changed the GUI in this PR so we can load/save camera files with just a button click per operation.  Before, we'd need to select a source/destination file **then** execute the save/load operation.  I felt it was too cluttered in the GUI and required too many mouse clicks.

New GUI:
Screen Shot 2022-10-18 at 2 01 48 PM

Old GUI:

StasJ (commented):
Looks good. My only note is it does not appear to handle multiple visualizers.

sgpearse:
Hmm you're correct.  There's a little bit of weirdness going on here because there isn't an individual params class that can set the camera position of all visualizers.  I think I have a reasonable workaround in function PCameraFileGroup::SetAllCameras(), which is passed as a callback to the PFileReader button.  Thoughts?

StasJ (approved):

PR #3241, opened 2022-09-22
disable clip to box

### LINKED ISSUE #3216: Particle Renderer length scale gives incorrect results

sgpearse:
Fix #3216  which regards particles rendered with direction disappearing when they exceed domain bounds, by disabling clip-to-box.


StasJ (commented):
Doesn’t this break the render region selector?

sgpearse:
> Doesn’t this break the render region selector?

It does not, because only particles from within the box extents are queried from the Grid.  Particles outside the box won't be rendered at all, and with this change, particles within the extents will have clipping disabled.

However in testing this, it's raised a new issue that seems to be in the UnstructuredGrid3D class, #3243.

sgpearse:
@stasj - I think we should be good to go here.  I can't currently re-request a review though, so I'm noting this via comment.

StasJ (approved):
Looks good to me!

PR #3247, opened 2022-09-26
Restructure particle GUI

### LINKED ISSUE #3244: Particles disappear when "Enable 3D Geometry" Disabled
1. Load DCP data and create/enable a particle renderer
1. Enable "Show direction"
1. Select appearance tab
2. Uncheck "Enable 3D Geometry"

### LINKED ISSUE #3245: Particles appearance GUI elements floating outside of sections
ss

sgpearse:
Fixes #3244 (I think)
Also fixes #3245.

Previously, the Particle gui would show/hide widgets based on certain parameters being enabled/disabled.  This could lead to a state where the "Length" widget would be hidden when rendering legacy particles.

Instead of showing and hiding widgets, this PR enables or disables them so they're always visible.


StasJ (commented):
If the GUI is disabled for 2D, how is the user to configure the 2D rendering?

sgpearse:
> If the GUI is disabled for 2D, how is the user to configure the 2D rendering?

Hmm.  I believe you're referring the the gui widgets becoming enabled or disabled according to the state of the `ParticleParams::Render3DTag` parameter.

In this context, the "Render3DTag" dictates whether we use shaders gl primitives.  As far as I'm aware, we do not support 2D particles, or we have no sample data that demonstrates it.

All this is doing is enabling parameters that are used by the shader, when we are using the shader.  Does that make sense, or is there something else you're referring to when you say "2D rendering"?

StasJ:
The particle renderer's code is heavily based on the flow renderer which has an `Enable 3D geometry` parameter. That parameter toggles between 2D and 3D shapes, not 2D/3D data.

sgpearse:
> The particle renderer's code is heavily based on the flow renderer which has an Enable 3D geometry parameter.  That parameter toggles between 2D and 3D shapes, not 2D/3D data.

Yes, however this is a place where the renderers' logic deviates.  For the Flow renderer, it determines wether 2D or 3D shaders should be used.  In the Particle renderer, it toggles between using a 3D shader, or the legacy rendering code that uses primitives.

Going back your original question:

> If the GUI is disabled for 2D, how is the user to configure the 2D rendering?

2D rendering was never supported, but a similar effect can be achieved by rendering in 3D and turning off lighting.  We could change `Enable 3D Geometry` parameter's name to something else, like `RenderLegacy` if you think that would help.

sgpearse:
@stasj - I made some additional changes to make the use of the legacy Particle renderer more like what we do for Flow.  Unlike the flow renderer, I'm not disabling the use of the legacy algorithm in Release builds, because it's been noted that it performs faster.  Let me know what you think.

New appearance tab:
Screen Shot 2022-10-20 at 9 46 10 AM

Old appearance tab:
Screen Shot 2022-10-20 at 9 49 21 AM

PR #3248, opened 2022-09-26
Add default variable logic for particle renderer

### LINKED ISSUE #3246: Particle renderer does not work without selecting velocity variables
Create/enable a particle renderer without selecting velocity variables.

```
Error performing paint event
Could not get scalar and field grids for ParticleRenderer
Variable "" on different grid from main variable
```

sgpearse:
Fixes #3246 by adding new logic for selecting default field variables.  If Vapor cannot detect X and Y variables beginning with U or V, then we search for variables that end in X, Y, and (if 3d) Z.

This applies to the use cases of 1) magnetic fields (bx, by, and bz) and our DCP standard naming convention (vel_x, vel_y, and vel_z).

### DIFF
--- a/include/vapor/RenderParams.h
+++ b/include/vapor/RenderParams.h
@@ -479,6 +479,7 @@ private:
     static const string _currentTimestepTag;
 
     string _findVarStartingWithLetter(std::vector<string> searchVars, char letter);
+    string _findVarEndingWithLetter(std::vector<string> searchVars, char letter);
 
 public:
     static const string _variableNameTag;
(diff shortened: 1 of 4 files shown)


StasJ (commented):
This is a nice feature but I don't see how it fixes the bug? The particle renderer will still fail when the velocity variables are not set, even if they are not used.

sgpearse:
I see what you mean.  I added a condition to ParticleRenderer.cpp:320 that prevents reading vector variables when direction is not enabled.  I'm not sure why git is saying I rewrote 28 lines in that file.  The condition at line 320 is what corrects the issue.

StasJ (approved):

PR #3272, opened 2022-11-09
Height variable fix

### LINKED ISSUE #3162: Height mapped variables on 2D grids
As discussed in #3155, Vapor applies height variables to 2 dimensional renderers as an offset from the bottom of the data domain.  

I believe the behavior we want is for 2D height mapped grid points to be rendered at the coordinate of the height variable.  IE - If a dataset has a minimum Z extent of 1500m, the Z coordinate of the height-mapped mesh should be HGT, not be HGT+1500.

If there's a use case for using HGT+Zmin, allowing the user to switch between the two modes would be helpful.

sgpearse:
Fix #3162 by interpreting height variables as a displacement from mean sea-level in meters.  I'm nearly certain that this is the behavior we want, but at one point we discussed allowing the user to switch between two modes with a checkbox:

1) Interpret height as a displacement from MSL, as is done in this PR
2) Interpret height as a displacement from the base of the domain

I have an implementation that allows for switching between these modes on another branch.  You can test the functionality out with [this CircleCI build artifact](https://app.circleci.com/pipelines/github/NCAR/VAPOR/6207/workflows/2f947990-e19d-4ef6-b0e1-a043ead0047e/jobs/15031) or see it in the gif below.  Here is [a good data file](https://drive.google.com/file/d/1jZDNZPBNCdmWhzwCXdPv_iQgVNWO8PD8/view?usp=share_link) to use for testing.

However switching between these modes adds logic to several of our renderers that I don't think should be added unless necessary.  This functionality may warrant a greater overall refactoring to make the handling of height variables consistent among all 2D renderers.  Before we do this, we should decide if this a use case that we truly need to support.  Would it be ok for us to require that users just translate their renderers that need be positioned as in the rarely-used requirement 2?

IE - I believe the switching between modes is unnecessary.  Height-mapped 2D renderers should behave as shown below, when the checkbox is off.  The user can manually translate the renderer if needed.  @clyne I'm requesting review to make sure we have the right requirements, not for code review or testing.
Nov-09-2022 10-55-22

### DIFF
--- a/apps/vaporgui/PVariableSelector.cpp
+++ b/apps/vaporgui/PVariableSelector.cpp
@@ -1,4 +1,5 @@
 #include "PVariableSelector.h"
+#include "PCheckbox.h"
 #include <vapor/RenderParams.h>
 #include <assert.h>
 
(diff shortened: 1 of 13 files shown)


clyne (approved):
Scott, is the text for the label, "add height values to the bottom of the domain", accurate? To me this implies that they are added to the minimum Z coordinate (which may be correct). Or are the height values added to the Z coordinate at each respective horizontal coordinate? 

In either case, a more helpful label might be "height values are relative", and provide a tooltip say that in the relative case the displacement values are added to bottom of the domain (or Z coordinate of the bottom layer of the grid at the associated X,Y, coordinate)

sgpearse:
> Scott, is the text for the label, "add height values to the bottom of the domain", accurate? To me this implies that they are added to the minimum Z coordinate (which may be correct). Or are the height values added to the Z coordinate at each respective horizontal coordinate?

Yes, this is accurate.  The HGT variable gets added to the minimum Z coordinate when the checkbox is enabled.  I'll change the label and add a tooltip if this is the way we want to go, but I don't think there's a use case for this, at least that I have seen.

This PR **does not include** the checkbox.  I have another branch that does, which I can submit a PR for.  I don't think need it when the user can just translate the renderer for this rare use case.  This checkbox would introduce some ugly conditionals in the renderers because none of them handle height variables in the same way, and clutter the gui.

StasJ (changes requested):
Open DUKU and create a 2D data renderer and a 2D wireframe renderer. Set both their height variables to HGT. Set both their Z Scales to 10. Notice they are at different heights.

b2

Open DUKU which is a mountainous area. Create an image renderer and set its height to HGT. Notice the image renderer is both lower and higher than the image renderer which should be at "zero"

b1


This PR does not make any changes to the Contour renderer which supports height offsets.

Finally, and partially unrelated to this PR, I noticed that the 2D Flow renderer does not support height offsets. Should it?

sgpearse:
@stasj Try to make the two renderers' origins match.  For whatever reason, they mismatch by default.

sgpearse:
Accident.

StasJ (changes requested):
@sgpearse did you mean to request a re-review?

With regards to your comment on setting origins, this PR is supposed to, as you state, have renderers interpret height variables as a displacement from mean sea-level in meters. If two renderers are set to the same height, yet they have different resulting heights, clearly one of them is incorrect.

sgpearse:
> @sgpearse did you mean to request a re-review?

@stasj - Yes.  Maybe you missed my comment.  

This is due to a separate issue, #3277.  The wireframe's default origin is 10k when it should be 0, like the TwoDData renderer's.   Set both origins to 0 and they will match height.

Nov-16-2022 09-34-58

StasJ (approved):

NihanthCW (approved):

PR #3273, opened 2022-11-10
Issue3162 v1

sgpearse:
This PR adds a checkbox that adds the minimum Z extent to 2D renderers using a height variable.  Is there a use case for this???

This is the alternative to PR #3272, which simply represents height variables as an absolute coordinate.  I think that's what we want, not this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -932,15 +932,15 @@ workflows:
   build:
     jobs:
       #- clang-tidy
-      - build_ubuntu18
-      - build_centos7
-      - build_python_api_ubuntuDebug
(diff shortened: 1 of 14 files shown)


NihanthCW:
Hi Scott, just to clarify: So, in the end, we want to choose either #3272 or #3273 right?

sgpearse:
> @sgpearse , just to clarify: So, in the end, we want to choose either https://github.com/NCAR/VAPOR/pull/3272 or https://github.com/NCAR/VAPOR/pull/3273 right?

Yes, that's correct.

PR #3275, opened 2022-11-14
Restructure smoketest functions

### LINKED ISSUE #3094: Mess up smoke test headers won't break build.
Funny fact: mess up smoke test headers won't break build. Is it a feature or a bug? 

Example: 
If I change the function signature of this line: https://github.com/NCAR/VAPOR/blob/339086b671be773c86a97307e003d4e5c4edc879/test_apps/smokeTests/gridTools.h#L40

to be:
```c++
bool RunTest(int haha_I_am_a_new_param, VAPoR::Grid *grid, bool silenceTime);
```
there will be no build failures at all, which alludes, can these smoke test results be trusted?

sgpearse:
Fixes [#3094](https://github.com/NCAR/VAPOR/issues/3094) by re-ordering smoke test functions. Some functions were being defined before being called resulting in multiple signatures. This resulted in modifications to the the header creating new function definitions, which wouldn't be called in lieu of the source's definition.

This also re-orders the smoke test function definitions to match the order of their header declarations.

### DIFF
--- a/test_apps/smokeTests/dataMgrTools.cpp
+++ b/test_apps/smokeTests/dataMgrTools.cpp
@@ -3,6 +3,25 @@
 #include "gridTools.h"
 #include "dataMgrTools.h"
 
+int TestDataMgr(const std::string &fileType, size_t memsize, size_t nthreads, const std::vector<std::string> &files, const std::vector<std::string> &options, bool silenceTime)
+{
+    VAPoR::DataMgr dataMgr(fileType, memsize, nthreads);
+    int            rc = dataMgr.Initialize(files, options);
+    if (rc < 0) {
+        cerr << "Failed to intialize WRF DataMGR" << endl;
+        return -1;
+    }
+
+    PrintDimensions(dataMgr);
+    PrintMeshes(dataMgr);
+    PrintVariables(dataMgr);
+    TestVariables(dataMgr, silenceTime);
+    PrintCoordVariables(dataMgr);
+    PrintTimeCoordinates(dataMgr);
+
+    return 0;
+}
+
 void PrintDimensions(const VAPoR::DataMgr &dataMgr)
 {
     vector<string> dimnames;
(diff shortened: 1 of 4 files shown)


shaomeng:
> Fixes [#3094](https://github.com/NCAR/VAPOR/issues/3094) by re-ordering smoke test functions. Some functions were being defined before being called resulting in multiple signatures.

I have 2 questions Scott:
1. Isn't "being defined before being called" the expected way of writing C++ functions?
2. Multiple signatures sound like a problem need to be addressed. I'd expected that one or more signatures to be deleted to address it. Could you explain why "re-ordering smoke test functions" can solve it?

sgpearse:
> Isn't "being defined before being called" the expected way of writing C++ functions?

Yes.

> Multiple signatures sound like a problem need to be addressed. I'd expected that one or more signatures to be deleted to address it. Could you explain why "re-ordering smoke test functions" can solve it?

This solves that problem by ordering the functions in a way that we only have one declaration in the header and one definition in the source.  Before, we had a declaration in the header and an extra declaration+definition in the source due to the order.

IE - The code will successfully call the signature of myFunc() that is declared+defined in the source.  We can modify the header declaration and it will not break the build.

foo.h

`void myFunc( int dummyVar );`

foo.cpp

```
#include "foo.h"

void myFunc() { std::cout << "myFunc()" << std::endl; }

void caller() {
    myFunc();
}
```

shaomeng:
Honestly I don't follow very well. One of the big advantages of C++ over C is that one don't need to worry about function declaration ordering anymore. More specifically, when writing C code, one needs to be careful about ordering often, and C++ eliminates this requirement, thus reducing chances to shoot your feet.

Could the smoke test program take advantage of this wonderful capability provided by C++? Or maybe explain why it cannot take advantage of it? My thinking here is, if the code is written in a way that function ordering matters, then it's quite possible that the ordering is messed up again later down the road.

shaomeng:
Scott, the file `gridTools.cpp` is fairly complicated with more than 700 lines of code, and it could further grow with more tests added. I suggest that you take advantage of C++ classes and write these tests in a way that relative ordering of functions doesn't matter. By doing so, we not only fix bug #3094, but also eliminate the chance of a further change breaking the ordering. See my previous comment on why it's a good idea to not require a strict ordering. 

One example of ordering matters in the current form: 
If you move the function `bool RunTest(Grid *grid, bool silenceTime)` from line 488 to line 275, the code won't compile anymore.

shaomeng (changes requested):
Given that the smoke test program is more than 700 lines of code and more than 2 dozen functions implemented inside, it should be written following objective oriented paradigm instead of functional paradigm to prevent similar issues from occuring again.

PR #3315, opened 2023-01-20
Python API documentation overhaul

### LINKED ISSUE #3304: Documentation is wrong about the vector field multiplier
In the documentation, the following is stated:

Vector field multiplier: Allows the user to multiply their vector field values by a scalar. This is useful in the situation
where the data domain and the vector field are in different units. For example, if the data domain is in units of km, and
the vector field is in units of m/s, the user may **multiply their vector field by 1000** so that all units are congruent with
each other.

This should read "**multiply their vector field by 0.001**"

I had some very weird results until I figured this out.

- [ ] Did VAPOR crash?
- [ X] Did you get wrong results?

**Impact**
- [X ] High - User productivity significantly degraded
- [ ] Medium - User productivity partially degraded

### To Reproduce

I have a model with a domain in km and velocity data in m/s just like the "for example" above, but 1000 should be 1/1000! To reproduce, just follow the advice above!

sgpearse:
Fixes #3304

This PR makes the following changes to our documentation

1) Switch themes from sphinx-rtd-theme to sphinx-book-theme
2) Adds Python API installation instructions, examples, and class reference
3) Moves documentation content into three primary sections, 1) Data formatting requirements, 2) Vapor application reference, and 3) Python API reference
4) Many organizational changes and naming changes
5) readTheDocs is now configured to build using a conda environment

The large number of files changed (447) are due to 1) the Python API having many classes that are documented in this PR, and 2) Renaming directories changes all contained files.

I suggest comparing the current docs with the ones in this PR via the following links:
[Current website](https://vapor.readthedocs.io/en/readthedocs/#)
[New website in this PR](https://vapor.readthedocs.io/en/pythonhelp/index.html)


NihanthCW (approved):
This looks good to me.
The actual python api reference can be polished in appearance but as we've discussed earlier, we can address that is successive iterations. I think this is a very good starting point. Thanks Scott!

StasJ (changes requested):
## Requested Changes

I believe it is standard practice not to include generated files in git repos since a small change in one of the config files may result in a commit with tens of thousands of changes. 

We probably don't need to expose internal python modules like `vapor.cppyyDoxygenWrapper` or `vapor.renderer.FuncWrapperRegex` since those are not relevant to the end user. In some cases they are the majority which could make it hard to navigate. For example, below there is only one relevant module:

s1

## Comments

Would it be possible to include some sort of formatting to make it more readable? For example in [xarray's documentation](https://docs.xarray.dev/en/stable/generated/xarray.plot.FacetGrid.html) the formatting makes it easier to distinguish different elements while in Vapor's doc everything except the class name is plain text.

StasJ (approved):

NihanthCW (approved):

shaomeng (approved):

PR #3322, opened 2023-02-02
Fix Deleting a Python variable plotted in a renderer causes Vapor to crash #3293

### LINKED ISSUE #3293: Deleting a Python variable plotted in a renderer causes Vapor to crash
### Describe the bug
When an active renderer is showing a user-derived variable, and the variable is deleted in the Python derived variable form, the renderer crashes to shell

- [X] Did VAPOR crash?

**Impact**
- [X] High - User productivity significantly degraded

### To Reproduce
1. Go to Python derived variables
2. Create a derived variable
3. Create a renderer and plot the variable according to type
4. Enable the renderer
5. Open Python derived variables and delete variable
6. Vapor crashes to shell

### Expected behavior
I expect Vapor to warn the user the variable is used and can't be deleted, or the renderer to be disabled until a new variable is set.

### Desktop
 - OS and version: [Windows 11 21H2]
 - Version: [Vapor 3.7]

StasJ:
Fix #3293

### DIFF
--- a/lib/common/VAssert.cpp
+++ b/lib/common/VAssert.cpp
@@ -16,6 +16,6 @@ void Wasp::_VAssertFailed(const char *expr, const char *path, const unsigned int
 #ifdef NDEBUG
     exit(1);
 #else
-    raise(SIGTERM);
+    raise(SIGINT);
 #endif
 }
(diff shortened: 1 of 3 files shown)


sgpearse (changes requested):
The problem is fixed, but I'm curious what lib/osglCMakeLists.txt has to do with the fix.  It's still fixed when I checkout main's version.  Was this an accident or is there something I'm missing?

StasJ:
> The problem is fixed, but I'm curious what lib/osglCMakeLists.txt has to do with the fix. It's still fixed when I checkout main's version. Was this an accident or is there something I'm missing?

Not an accident, just a minor thing that didn't warrant it's own PR.

sgpearse (approved):

PR #3330, opened 2023-02-08
fix #3318

### LINKED ISSUE #3318: Flow deltaT calculation on mostly zero fields
Flow integration relies on a step size, noted as `deltaT` in the code base. The current code samples velocity from 1000 positions in the volume, and use the maximum sampled velocity to estimate an initial `deltaT`. This approach is mostly sufficient, however, when there's only a tiny region containing non-zero velocity, these 1000 samples can easily miss the non-zero regions.

To reproduce, download and extract the following dataset [tube_data.tar.gz](https://github.com/NCAR/VAPOR/files/10504709/tube_data.tar.gz). Use the seed in `listOfSeeds.txt` to generate a flow line, and you'll see a straight line, indicating that `deltaT` is too big.  After manually scaling down the initial `deltaT` by 400, a more plausible flowline is produced, as the screenshot shows.

A possible solution: provide a knob to users to manually scale up/down the initial step size. This knob will have a default value of 1.0, and will remain 1.0 for most use cases.

Note: this bug was originally reported on [discourse](https://vapor.discourse.group/t/trouble-with-the-flow-renderer/260/6).

Screenshot from 2023-01-25 16-33-29

shaomeng:
Follow steps described in #3318 , one initially gets a straight line.

After manually setting `First Step Size Multiplier` to `0.01`, then a spiral flow line is produced.

### DIFF
--- a/apps/vaporgui/FlowEventRouter.cpp
+++ b/apps/vaporgui/FlowEventRouter.cpp
@@ -43,6 +43,7 @@ FlowEventRouter::FlowEventRouter(QWidget *parent, ControlExec *ce) : RenderEvent
                 _pathlineInjectionSlider = new PIntegerSliderEdit(FP::_seedInjInterval, "Injection Interval"),
             }),
             new PDoubleInput(FP::_velocityMultiplierTag, "Vector Field Multiplier"),
+            new PDoubleInput(FP::_firstStepSizeMultiplierTag, "First Step Size Multiplier"),
             new PCheckbox(FP::_xPeriodicTag, "X axis periodicity"),
             new PCheckbox(FP::_yPeriodicTag, "Y axis periodicity"),
             new PCheckbox(FP::_zPeriodicTag, "Z axis periodicity"),
--- a/apps/vaporgui/FlowEventRouter.cpp
+++ b/apps/vaporgui/FlowEventRouter.cpp
@@ -42,8 +42,6 @@ FlowEventRouter::FlowEventRouter(QWidget *parent, ControlExec *ce) : RenderEvent
                 _pathlineLengthSlider = new PIntegerSliderEdit(FP::_pastNumOfTimeSteps, "Pathline Length"),
                 _pathlineInjectionSlider = new PIntegerSliderEdit(FP::_seedInjInterval, "Injection Interval"),
             }),
-            new PDoubleInput(FP::_velocityMultiplierTag, "Vector Field Multiplier"),
-            new PDoubleInput(FP::_firstStepSizeMultiplierTag, "First Step Size Multiplier"),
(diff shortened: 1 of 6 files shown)


sgpearse (changes requested):
@shaomeng - Can we add our new widget to a new PSection entitled "Advanced Options" at the very bottom of the Seeding tab?  Also, something this obscure probably deserves a descriptive tooltip.  See [line 69](https://github.com/NCAR/VAPOR/blob/6fafb33aa0f9c1b4e7e518288b57c19aa4bcf2b8/apps/vaporgui/FlowEventRouter.cpp#L69) for an example.

shaomeng:
> @shaomeng - Can we add our new widget to a new PSection entitled "Advanced Options" at the very bottom of the Seeding tab? Also, something this obscure probably deserves a descriptive tooltip. See [line 69](https://github.com/NCAR/VAPOR/blob/6fafb33aa0f9c1b4e7e518288b57c19aa4bcf2b8/apps/vaporgui/FlowEventRouter.cpp#L69) for an example.

Good idea to add a tooltip. Question regarding "Advanced Options": do you know how to hide/show it?

sgpearse:
That can be done with PShowIf() which has a few examples in FlowEventRouter.cpp.  I'm guessing we only want to show this for steady flow?  

You just need to feed PShowIf() the tag of a parameter and it will show/hide its contents.  Line 68 is as follows:

```
 68   (new PShowIf(FP::_seedGenModeTag))->Equals((int)FlowSeedMode::LIST)->Then({
 69       (new PFileOpenSelector(FP::_seedInputFilenameTag, "List of seeds file"))
 70         ...
```

shaomeng:
> That can be done with PShowIf() which has a few examples in FlowEventRouter.cpp. I'm guessing we only want to show this for steady flow?
> 
> You just need to feed PShowIf() the tag of a parameter and it will show/hide its contents. Line 68 is as follows:
> 
> ```
>  68   (new PShowIf(FP::_seedGenModeTag))->Equals((int)FlowSeedMode::LIST)->Then({
>  69       (new PFileOpenSelector(FP::_seedInputFilenameTag, "List of seeds file"))
>  70         ...
> ```

Never mind, I think it's a control that applies to both steady/unsteady. Finally question, could you show me a sample code on how to create an advanced section and place a specific widget in it?

sgpearse:
It would look something like this, as the last item in the _seedingTab item after the "Write flow lines to file" section:

```
new PSection("Advanced Options", {
    new PDoubleInput(FP::_firstStepSizeMultiplierTag, "First Step Size Multiplier")->SetToolTip("...")
})
```

sgpearse (approved):
@shaomeng - good idea of adding the velocity multiplier to Advanced Options.

PR #3336, opened 2023-02-22
Change notifier URL to one with an improved repository name

sgpearse:
The repository that hosts Vapor's landing page at https://www.vapor.ucar.edu/ is named `Vapor-SupportPage `because it was previously used as a user support form years ago.  Since it is no longer used for that, it would be less confusing if it had a more descriptive name.

We also use this repository to store Vapor's notifications.  This PR updates the URL used by the notifier to direct to a repository named `VAPOR-LandingPage`, instead of `VAPOR-SupportPage`.

Once this is merged, we can change the actual repository name without breaking the notifier.

### DIFF
--- a/apps/vaporgui/CheckForNotices.cpp
+++ b/apps/vaporgui/CheckForNotices.cpp
@@ -91,7 +91,7 @@ void CheckForGHNotices(std::function<void(const std::vector<Notice> &)> callback
 #ifdef TESTING_API
     req.setUrl(QUrl("http://localhost:8000/list.json"));
 #else
-    req.setUrl(QUrl("https://api.github.com/repos/NCAR/VAPOR-SupportPage/contents/notices"));
+    req.setUrl(QUrl("https://api.github.com/repos/NCAR/VAPOR-LandingPage/contents/notices"));
 #endif
     manager->get(req);
 }


NihanthCW (approved):
Looks good. I guess you were anyway planning to do this but I would just double check after this is merged to ensure everything is working.

StasJ (changes requested):
This will break the notifier for all current and previous versions of Vapor.

sgpearse:
@stasj - Do you think there is an ideal time to do this, or that it should not be done at all?  Some things to consider are 1) whether we plan on having any notifications before 3.8.1's release, and 2) Vapor's release notification system does not use this repository, and would be unaffected by this.

The last time we changed the VAPOR-SupportPage repo's name, it did indeed break the notifier.  Conversely, it seems that [github should be redirecting references](https://github.com/orgs/community/discussions/22669#discussioncomment-3237667) to the old repo to the new one.  I don't know why that didn't happen last time.

StasJ:
@sgpearse I believe last time it was changed by accident. Let's discuss in person tomorrow.

sgpearse:
Per today's discussion, we will opt not to change the repo's name and live with this quirk so that we don't break the notifier in previous Vapor versions.

PR #3351, opened 2023-03-08
improve out-of-bound detection

### LINKED ISSUE #3325: Flow: Pathlines seem to never stop advecting
flow

Pathlines seem to get "stuck" during their advection, leading to increasing rendering times as more pathlines are injected into the scene.

To reproduce:
1. Download the attached .zip file that contains a BOV files, a binary data file, and a session file.
2. Load the session file and point it to the attached data.  Failing that, load the data and create a flow renderer as follows:
    - Pathlines
    - X, Y, and Z Field Variables = a
    - Pathline Length = 1
    - Injection Interval = 1
    - X, Y, and Z axis seeds = 10
5. Enable image capture and run the time series

The images on my computer and Casper are generated quickly at first but then slow down.  It only takes 5 time steps for the pathlines to completely exit the domain.  Rendering time should theoretically plateau after 5 time steps.

Many renderings captured in the first minute at the beginning of the time series:
Screen Shot 2023-02-02 at 11 04 00 AM

One capture per 2 minutes near timestep 100:
Screen Shot 2023-02-02 at 11 04 53 AM

[flowBug3325.zip](https://github.com/NCAR/VAPOR/files/10572883/flowBug3325.zip)

shaomeng:
This PR would fix #3325.

### DIFF
--- a/lib/flow/Advection.cpp
+++ b/lib/flow/Advection.cpp
@@ -90,6 +90,7 @@ int Advection::AdvectSteps(Field *velocity, double deltaT, size_t maxSteps, ADVE
             }
 
             if (rv == SUCCESS) {
+                // Bookmark_1
                 // The new particle *may* be the same as the old particle in case
                 // there's a sink, meaning the velocity is zero.
                 // In that case, we mark p1 as "special" and terminate the current stream.
@@ -104,6 +105,7 @@ int Advection::AdvectSteps(Field *velocity, double deltaT, size_t maxSteps, ADVE
                     numberOfSteps++;
                 }
             } else if (rv == MISSING_VAL) {
+                // Bookmark_2
                 // This is the annoying part: there are multiple possiblities.
                 // 1) past0 is really located at a missing value location;
                 // 2) past0 is inside the volume, but really close to the boundary,
@@ -198,6 +200,9 @@ int Advection::AdvectTillTime(Field *velocity, double startT, double deltaT, dou
         if (p0.time < startT)      // Skip this stream if it didn't advance to startT
             continue;
 
+        if (p0.IsSpecial())
+            continue;
+
         size_t thisStep = 0;
 
         while (p0.time < targetT) {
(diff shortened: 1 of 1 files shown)


shaomeng:
@sgpearse, could you test this branch on the real data set, and see if there's still "stuck" effect as shown in your original bug report?

sgpearse:
@shaomeng I ran some tests over the weekend and I do not see pathlines getting stuck, but the time it takes to perform a rendering is still increasing over time.  This can be seen with the screenshots in /glade/p/cisl/vast/vapor/Bugs/3325/mar10, which take tens of seconds at the start, gradually increasing to ~2 minutes at the end.

To reproduce, you can look at my real dataset session file /glade/p/cisl/vast/vapor/Bugs/3325/3325_v2.vs3.  Alternatively I also increased the size of the synthetic dataset to 10,000 timesteps which can be imported from the directory /glade/p/cisl/vast/vapor/Bugs/3325/bovFiles.

I'd be ok approving this PR because it ameliorates the issue, but keeping the original issue open.  Thoughts?  Let me know if there's anything I can do to help troubleshoot.

sgpearse:
> Hi Scott, glad to know that there's no "stuck" pathlines anymore. I consider that's a successful fix.

Sounds good.  Let's accept this fix and if it resurfaces we can re-open.

> That means the number of total pathlines is 2X at step two, 3X at step three, and eventually 10,000X after 10,000 time steps. Though "out-of-bound" pathlines are not advected anymore, they are still visited and tested every time, taking some computational time.

Can you talk about why this is necessary?  Once a pathline has gone out of bounds, why keep on processing it?  I haven't yet tried pushing the upper limits with the pathline renderer because of the current issue at hand (#3325).  I plan on using pathlines heavily for future work.

sgpearse (approved):

sgpearse:
> Even out-of-bound lines are not advected, they're visited and tested.

But why?  Maybe this is best discussed tomorrow, but once the ball is out of play, tennis players and line judges disregard it.  Can't we treat pathlines in the same way by ignoring them once they've exited the domain?

sgpearse:
Accepting, merging, and closing.  #3353, which regards optimizations for the Pathline algorithm can be discussed in the future.

PR #3395, opened 2023-05-22
Refactor renderer GUI

StasJ:
o

- New design is more in line with other 3D software
- More compact allowing it to present additional useful information such as the primary rendered variable directly in the list view
- Visibility is represented by a standard eye/crossed eye rather than the previous checkbox so that it no longer requires a header to explain the functionality
- Renderers are grouped under their dataset
- Custom elements are macOS dark-mode aware
- Adds ability to inspect datasets
- Adds context menus to each renderer
- Adds ability to rename renderers
- Includes incremental refactor of the tab manager
- Includes standardized widgets for elements used in higher level GUI, e.g. scroll areas, routers, etc.


shaomeng (approved):
The renderer list looks much better, and feels much easier to use now! Great job Stas!

sgpearse (changes requested):
Looks great Stas.  I think I found a bug:

1) Import any wrf file and create a Barb and Slice renderer.  Enable them.
2) Delete the Barb renderer
3) Edit->Undo

The Barb renderer will come back into the RenderHolder, but I'm not able to get it to paint.

sgpearse (approved):

PR #3414, opened 2023-07-07
Relative dataset path support

### LINKED ISSUE #3389: Vapor does not handle session files with relative dataset paths
These session files can be created when vapor loads datasets from the command line.

### LINKED ISSUE #3399: Relative, not absolute paths in VAPOR session files
I would like to share data including session files with other users so they can open up a shared dataset and see everything the way I do. Currently, session files contain absolute paths to the data, obviating sharing session files with people on different computers. My request is that Vapor output relative paths, not absolute paths. I edited out all of the full paths in a Vapor session file such that they were relative, and it worked fine, so this is currently working, just not implemented by Vapor itself.

**Impact**
- [ ] High - User productivity significantly improved
- [x] Medium - User productivity partially improved
- [ ] Low - User productivity trivially improved

StasJ:
- Adds support for loading/saving dataset paths relative to session files
- When saving a session file, vapor now checks if all of the open datasets are contained within the directory containing the session file. If so, it will save dataset paths as relative to the session file. Otherwise it will use absolute paths.

Expected behavior when sharing sessions:
- Sharing session pointing to data on shared HPC drive in /glade will be absolute.
- Sharing a folder containing data and session will be relative.

Fix #3399
Fix #3389

### DIFF
--- a/include/vapor/FileUtils.h
+++ b/include/vapor/FileUtils.h
@@ -16,6 +16,7 @@ COMMON_API std::string ReadFileToString(const std::string &path);
 COMMON_API std::string HomeDir();
 COMMON_API std::string Basename(const std::string &path);
 COMMON_API std::string Dirname(const std::string &path);
+COMMON_API std::string Realpath(const std::string &path);
 COMMON_API std::string Extension(const std::string &path);
 COMMON_API std::string RemoveExtension(const std::string &path);
 COMMON_API std::string POSIXPathToWindows(std::string path);
@@ -26,11 +27,13 @@ COMMON_API bool        IsPathAbsolute(const std::string &path);
 COMMON_API bool        Exists(const std::string &path);
 COMMON_API bool        IsRegularFile(const std::string &path);
 COMMON_API bool        IsDirectory(const std::string &path);
+COMMON_API bool        IsSubpath(const std::string &dir, const std::string &path);
 COMMON_API FileType    GetFileType(const std::string &path);
 COMMON_API std::vector<std::string> ListFiles(const std::string &path);
 
 //! @code JoinPaths({"home", "a/b"}); @endcode
-COMMON_API std::string JoinPaths(std::initializer_list<std::string> paths);
+COMMON_API std::string JoinPaths(const std::initializer_list<std::string> &paths);
+COMMON_API std::vector<std::string> SplitPath(std::string path);
 
 COMMON_API int MakeDir(const std::string &path);
 
--- a/include/vapor/STLUtils.h
+++ b/include/vapor/STLUtils.h
@@ -18,9 +18,13 @@ template<typename T> vector<T> Filter(const std::vector<T> &v, std::function<boo
     return v2;
 }
 
+template<typename T> bool BeginsWith(const T &str, const T &match) {
+    return str.size() >= match.size() && equal(match.begin(), match.end(), str.begin());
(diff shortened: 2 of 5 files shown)


shaomeng:
I think that this PR addresses a less-common use case, while leaving a more-common use case unaddressed. It concerns if a user is more used to save the session file in the same folder/in a subfoler as the data, or a parent folder of the data. Here's an example. 

Imagine a user visualizes data saved in his `~/tornado_data/` folder, I suppose that it's natural to save the session file inside of `~/tornado_data/` or even a subfoler: `~/tonado_data/vapor_session/`. The session file should use relative path in this case. However, that's not what this PR addresses. Instead, this PR will only use relative path if the user is saving the session file at `~`. 

I'm curious to know what other members consider to be a common use case.

StasJ:
> I think that this PR addresses a less-common use case, while leaving a more-common use case unaddressed. It concerns if a user is more used to save the session file in the same folder/in a subfoler as the data, or a parent folder of the data. Here's an example.
> 
> Imagine a user visualizes data saved in his `~/tornado_data/` folder, I suppose that it's natural to save the session file inside of `~/tornado_data/` or even a subfoler: `~/tonado_data/vapor_session/`. The session file should use relative path in this case. However, that's not what this PR addresses. Instead, this PR will only use relative path if the user is saving the session file at `~`.
> 
> I'm curious to know what other members consider to be a common use case. @sgpearse @NihanthCW

@shaomeng This PR already does what you describe, except for sessions in subfolders because that would make examples such as the one below relative.

```
~/output.nc
~/Documents/session.vs3
```

shaomeng:
I think it's easiest to discuss the desired behavior using the following examples:

A user is looking at data residing at `/home/sam/tornado_data/`. Now they want to save a session file. There are a few places to save it:
1.  `/home/sam/tornado_data/vapor_sessions/`
2. `/home/sam/tornado_data/`
3. `/home/sam/`
4. `/home/sam/vapor_sessions/`
5. `/home/stas/`

Respectively, in my opinion, the saved session file should be using: 
1. relative path
2. relative path
3. absolute path
4. absolute path
5. absolute path

I'd like know what other people think about these use cases.

NihanthCW:
> I think it's easiest to discuss the desired behavior using the following examples:
> 
> A user is looking at data residing at `/home/sam/tornado_data/`. Now they want to save a session file. There are a few places to save it:
> 
> 1. `/home/sam/tornado_data/vapor_sessions/`
> 2. `/home/sam/tornado_data/`
> 3. `/home/sam/`
> 4. `/home/sam/vapor_sessions/`
> 5. `/home/stas/`
> 
> Respectively, in my opinion, the saved session file should be using:
> 
> 1. relative path
> 2. relative path
> 3. absolute path
> 4. absolute path
> 5. absolute path
> 
> I'd like know what other people think about these use cases.

Can we simply enforce that users need to save the session file at a certain location for relative paths to work?
or 
present an option (eg. a checkbox somewhere) to specify whether to save the file path in absolute or relative terms? Otherwise, what's stopping them from saving the session file in a random subfolder/parent folder. I am not sure how we might address all those use cases without complicating the logic. I suspect that would also make debugging more difficult in the future.

sgpearse (approved):

NihanthCW:
Open a new issue for the checkbox option.

PR #3445, opened 2023-08-24
Fix #3440: Moving domain bug

### LINKED ISSUE #3440: Moving domain bug
To reproduce, download the dataset [in this zip file](https://drive.google.com/file/d/1Y89Ij1lHURIZ69z2AkvQg0l3MTjUfSoz/view?usp=sharing).

1) Load the data
2) Create and enable an image and volume renderer
3) Advance to timestep 2
4) Disable the volume renderer
5) Create and enable a flow renderer

The attached session file should show the following screenshot:
Screen Shot 2023-08-21 at 7 25 13 AM

StasJ:
Fix #3440

### DIFF
--- a/lib/params/RenderParams.cpp
+++ b/lib/params/RenderParams.cpp
@@ -158,7 +158,8 @@ int RenderParams::Initialize()
     // look for others
     //
     string varname = GetVariableName();
-    size_t ts = 0;
+    size_t ts = GetCurrentTimestep();
+    size_t tsInit = ts;
     int    ndim = _maxDim;
     bool   foundVar = false;
     if (!varname.empty()) {
@@ -175,7 +176,10 @@ int RenderParams::Initialize()
             bool ok = DataMgrUtils::GetFirstExistingVariable(_dataMgr, 0, 0, ndim, varname, ts);
             if (ok) {
                 foundVar = true;
-                break;
+                if (_dataMgr->VariableExists(tsInit, varname, 0, 0)) {
+                    ts = tsInit;
+                    break;
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


sgpearse (changes requested):
Follow the steps in the bug report, then disable the Volume and Image renderers.  The Flow region is incorrect.

Screen Shot 2023-08-29 at 5 12 06 PM

StasJ:
> Follow the steps in the bug report, then disable the Volume and Image renderers. The Flow region is incorrect.

I get correct results when following your directions:

ss

sgpearse:
That's really strange.  When I load domain.vs3 in the zip file from the bug report, and disable all renderers but flow, something goes awry.  I'm definitely on issue3340.

Screen Shot 2023-09-01 at 8 18 04 AM

StasJ:
> That's really strange. When I load domain.vs3 in the zip file from the bug report, and disable all renderers but flow, something goes awry. I'm definitely on issue3340.

1. Loading `domain.vs3` is not one of the steps in the bug report:
> Follow the steps in the bug report, then disable...

2. That session file was generated using a bug. The bug was fixed, but it does not retroactively repair bad session files, especially since we never released a version containing this bug.

StasJ:
> That's fair, however when not using the session file I see the following which is the incorrect behavior from my screenshot 3 days ago :(. The domain is too long on the Y axis and the northern region of the flow renderer is truncated.
> 
> I'm not sure how we're seeing two different things. Would it be useful to do a screen share?

Just a sanity check: Are you using the pre-release installer or building from this branch for your tests?

sgpearse:
> Just a sanity check: Are you using the pre-release installer or building from this branch for your tests?

I'm building on issue3440 in debug mode with all other defaults applied.  But I don't think there's urgency since we've agreed to postpone this edge case until after release.  I'd like to do a screen-share when we meet on Tuesday though, just for sanity.

sgpearse (approved):

PR #3484, opened 2023-10-24
Add cross compilation rules for macos arm64/x86 and optimize circleci/config.yml

### LINKED ISSUE #3463: MacOS Gen1 and Gen2 being deprecated on CircleCI soon
Our x86 MacOS build protocols use xcode versions that will soon be deprecated.  These builds should be migrated to M1 systems to reduce their build times, and should have their xcode versions updated.

https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718

sgpearse:
Fixes #3463.  With CircleCI's deprecation of macOS/x86 executors, we need to start cross-compiling on arm64.

This PR also introduces reusable commands to reduce code repetition, and layer caching to save on compile time on the expensive M1 executors.


CMakeLists.txt
```diff
@@ -162,7 +157,6 @@ else()
     endif()
 endif()
 
-include_directories ("${PROJECT_SOURCE_DIR}/include")
```
StasJ, line 165: Is this not required?

CMakeLists.txt
```diff
@@ -22,20 +22,15 @@ if(NOT CMAKE_BUILD_TYPE)
...
         message("Building on macOS M1 architecture (arm64)")
-        # Set a flag or perform actions specific to M1 architecture
-        set(M1_ARCHITECTURE TRUE CACHE BOOL "Building on macOS M1 architecture")
         set(CMAKE_OSX_DEPLOYMENT_TARGET "12.0" CACHE STRING "Minimum OS X deployment version" FORCE)
-
+        set(CMAKE_OSX_ARCHITECTURES "arm64")
     else()
         message("Building on macOS x86 architecture")
         set(CMAKE_OSX_DEPLOYMENT_TARGET "10.15" CACHE STRING "Minimum OS X deployment version" FORCE)
+        set(CMAKE_OSX_ARCHITECTURES "x86_64" CACHE STRING "macOS build architecture" FORCE)
```
StasJ, line 33: We should target the host architecture by default.

StasJ (changes requested):

CMakeLists.txt
sgpearse: I'm perplexed on how the test builds were successful across multiple platforms.  I've re-added this line.  Thanks for the catch.

CMakeLists.txt
sgpearse: Ok, native build platforms are now targeted by default.  We can now target specific architectures by specifying CMAKE_OSX_ARCHITECTURES.

StasJ (approved):

CMakeLists.txt
StasJ: As am I...

CMakeLists.txt
StasJ: Great, thank you!

PR #3490, opened 2023-11-03
Fix Particle Lighting seems weird #3214

### LINKED ISSUE #3214: Particle Lighting seems weird.
The particle lighting seems weird, if not wrong. The test data is this one (/glade/p/cisl/vast/vapor/data/Source/DCP) and the following screenshots explains what's weird.

Note: the reference lighting effect pictures are from this website: https://www.cs.toronto.edu/~jacobson/phong-demo/.

##  Specular lighting seems weird
To illustrate the effect of specular lighting, let's turn off every other lighting and turn specular lighting to its full level, as the following screenshot shows. 
Screen Shot 2022-08-31 at 11 50 34 AM
However, this looks more like ambient lighting. A proper specular lighting on a sphere would look something like this:
Screen Shot 2022-08-31 at 11 52 16 AM

## Shineness is not strong enough. 
Let's try to use the strongest shineness, as the following screenshot shows. However, each sphere is barely shining. 
Screen Shot 2022-08-31 at 11 57 01 AM

A properly lighted sphere, however, would give very obvious shining effect, as the following screenshot shows.  
Screen Shot 2022-08-31 at 11 58 22 AM

StasJ:
Fix #3214

### DIFF
--- a/lib/params/ParticleParams.cpp
+++ b/lib/params/ParticleParams.cpp
@@ -53,7 +53,7 @@ void ParticleParams::_init()
     SetValueDouble(PhongAmbientTag, "", .4);
     SetValueDouble(PhongDiffuseTag, "", .8);
     SetValueDouble(PhongSpecularTag, "", 0.);
-    SetValueDouble(PhongShininessTag, "", 2.);
+    SetValueDouble(PhongShininessTag, "", 0.5);
 
     RenderParams::SetDefaultVariables(3, true);
 }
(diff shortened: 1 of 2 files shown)


shaomeng:
This PR changes the behavior of the lighting, but I wouldn't say it moves in the right direction.
Specifically, regarding the two issues reported in #3214 , 
1) the specular lighting remains the same
2) the Shineness is not improved, but appears more like fuzziness after this PR.

PR #3491, opened 2023-11-07
Fix issue3403 - reinstate smoke tests on SUSE

### LINKED ISSUE #3403: build_ubuntu18_installer tests need to be migrated to ubuntu20
With the upcomming deprecation of Ubuntu 18, we need to migrate all build, warning, and smoke tests to Ubuntu 20.

~~This will include the removal of the extra CMake options added in PR #3417.~~ Those options are still needed due to differences between Ubuntu20 and 22.

sgpearse:
This reinstates the smoke tests that were previously employed on Ubuntu18, which was deprecated.

I chose SUSE because I believe we will be supporting it for however long Derecho is running it.  We supported CentOS7 for ~9 years, which is longer than Ubuntu22's lifetime.

Fixes #3403.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,6 +106,8 @@ commands:
                 tar --no-same-owner -xf /tmp/smokeTestData.tar.gz -C /smokeTestData
                 chown -R root:root /smokeTestData
                 chmod -R 777 /smokeTestData
+            else
+                echo "Data already acquired"
             fi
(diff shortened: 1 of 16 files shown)


shaomeng:
I guess the only way to test it out is to merge it and then see how CircleCI handles it, right?

shaomeng (approved):
This looking good, thanks Scott!

PR #3529, opened 2024-01-25
Fix moving particle domain support and single particle crash

### LINKED ISSUE #3526: DCP dataset containing a single particle crashes vapor
[particles_000.nc.zip](https://github.com/NCAR/VAPOR/files/14042872/particles_000.nc.zip)

### LINKED ISSUE #3527: Track renderer regions does not properly adjust the region if the dataset extents get larger without moving
This might be the case in a particle dataset

StasJ:
Fix #3527
Fix #3526


NihanthCW (approved):
I tested the single particle crash and it's working. Didn't have a dataset to test the moving domain bug. We can get this confirmation from David .

PR #3539, opened 2024-02-16
Flow cleanup, Google Test

shaomeng:
This PR cleans up some flow code, including fixing a pointer consistency issue.


NihanthCW:
Remove Google Tests integration for now until further discussion.

shaomeng:
> Remove Google Tests integration for now until further discussion.

This PR is ready to be reviewed without GTest.

lib/flow/VaporField.cpp
```diff
@@ -575,4 +562,14 @@ const VAPoR::Grid *VaporField::_getAGrid(size_t timestep, const std::string &var
     return grid;
 }
 
-void VaporField::ReleaseLockedGrids() { _recentGrids.clear(); }
+
+void VaporField::ReleaseLockedGrids()
+{
+    // Release locked grids by giving the cache a bunch of nullptrs with unique invalid keys.
```
StasJ, line 568: This seems like a hack.

StasJ (changes requested):

StasJ (approved):

include/vapor/VaporField.h
```diff
@@ -1,28 +1,29 @@
...
     std::string           _varName;
-    uint64_t              _timestep;
-    int32_t               _refLev, _compLev;
-    std::array<double, 3> _ext_min, _ext_max;
+    VAPoR::CoordType      _ext_min = {0.0, 0.0, 0.0};
+    VAPoR::CoordType      _ext_max = {0.0, 0.0, 0.0};
 
 public:
-    void Reset(uint64_t, int32_t, int32_t, std::string, const std::vector<double> &, const std::vector<double> &);
+    void Reset(uint32_t, int32_t, int32_t, std::string, VAPoR::CoordType, VAPoR::CoordType);
```
sgpearse, line 26: Why switch from passing const references to non-const CoordType values?

include/vapor/ptr_cache.hpp
```diff
@@ -0,0 +1,106 @@
...
+// This is an implementation of a least-recently-used (LRU) cache that keeps
+// raw pointers pointing to big structures (e.g., grids, quadtrees).
+//
+// This cache has two execution policies:
+// 1) only insertion counts as `recently used`, and
+// 2) both insertion and query count as `recently used`.
+//
+// Given this design, this cache is expected to keep the ownership of these
+// structures once they're put in the cache, and all other codes will not
+// need to manage these structures. In other words, once an object is evected,
```
sgpearse, line 11: Typo: Do you mean evicted instead of evected?

sgpearse (commented):
Interesting stuff.  Just a question and I think I found a typo.

include/vapor/VaporField.h
shaomeng: In this case, I was attempting to make the function signature simpler (i.e., without using const and references).

Const references avoid a copy, which can be expensive if 1) the copy happens on the heap, and 2) the copy is big. In this case, after changing the signature, a copy happens on the stack and quite small with just 3 values, so it's not expensive anyways. I think the simplicity wins in this case.

include/vapor/ptr_cache.hpp
shaomeng: Good catch; should be fixed!

include/vapor/VaporField.h
sgpearse: Gotcha.  I was suprised to see this but if there's not a big performance hit then readability wins.

sgpearse (approved):

include/vapor/VaporField.h
shaomeng: Yeah, references have its own overheads, so for small amounts of data (debatable how small is small), simply make a copy seems easiest for everyone (including the compiler).

shaomeng:
This PR is approved, I'll merge it to main so that Leigh can try it out!

PR #3541, opened 2024-02-20
Improve particle renderer performance 12,500x

StasJ:
This PR makes the particle renderer over 12,500 times faster. Below Vapor is rendering over 20,000,000 particles in realtime. This PR also drastically decreases both its CPU and GPU RAM requirements.

performance


sgpearse (approved):
Interesting use of static functions for openGL setup.  That makes sense.  Interesting use of goto as well.  I avoid it because of stigma but the new code does read easily with it.

PR #3555, opened 2024-03-03
fix a mutex use in ptr_cache

shaomeng:
This PR fixes an incorrect use of a mutex in `ptr_cache` that resulted in crashes.


StasJ (changes requested):
This file uses a combination of 2-space and 4-space tab formatting which is inconsistent with itself and the rest of the codebase.

StasJ (approved):

PR #3562, opened 2024-03-13
fix broken windows build

### LINKED ISSUE #3561: Broken Windows Build
Our windows build is currently broken at Grid.cpp:140

`nDims = std::max(nDims, 1UL); // Otherwise will return 0 dim for single particle`

sgpearse:
Fix #3561 by using size_t instead of UL.

@stasj are there any implications for this fix?  Windows is being fussy about the size_t/UL type mismatch.


StasJ (approved):
> @StasJ are there any implications for this fix? Windows is being fussy about the size_t/UL type mismatch

No, it’s the same thing

PR #3571, opened 2024-04-01
Fix UnstructuredGrid2D broken

### LINKED ISSUE #3566: UnstructuredGrid2D broken
To reproduce:

1) import /glade/p/cisl/vast/vapor/data/Source/MPAS/FalkoJudt/*

Vapor will crash.  This is also causing our smoke tests to fail.

StasJ:
Fix #3566


StasJ:
Vapor considers length 1 topological data-dimensions as not contributing to the grids topological dimension. This can lead to certain valid scenario's which are considered invalid such as single particle datasets.

PR #3608, opened 2024-05-01
Use DataMgrUtils instead of CE

### LINKED ISSUE #3575: UGRID/large image renderer empty

sgpearse:
@stasj, I looked into using RenderParams to query a random 3D variable but there does not seem to be a way to do this.  The RenderParams gets instantiated with a dimensionality, and has no evident way to query variables outside of their given dimension.  The only path forward I found was to use DataMgrUtils, shown here.

Also, the CE method used in PR #3584 does in fact find the domain extents by taking the union of all variables in the dataset.  Not taking the union of the extents could break the PRegionSelector if we don't land on a variable that encloses the entire domain, which could happen.  WRF can contain soil and fire grids that lie outside or within the meteorological grids that we typically deal with.  

Feel free to pick whichever solution you like.

Fixes #3575 and #3583.

### DIFF
--- a/apps/vaporgui/PRegionSelector.cpp
+++ b/apps/vaporgui/PRegionSelector.cpp
@@ -1,6 +1,8 @@
+#include <string>
 #include "PRegionSelector.h"
 #include "QRangeSliderTextCombo.h"
 #include <vapor/RenderParams.h>
+#include <vapor/DataMgrUtils.h>
 #include <assert.h>
 #include "POrientationSelector.h"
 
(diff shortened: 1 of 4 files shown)


StasJ (changes requested):
Thanks, this solution requires much fewer changes!

Like I mentioned in the original PR review, however, this does not fix the underlying issue causing the crash in #3583, which is that the image renderer will crash for any zero width region.

sgpearse:
@stasj I chose to return from functions that were VAsserting when the texture size was too small.  I don't think it's necessary to report an error to the user when the shrink the region to something that won't render anything.

StasJ (approved):
Looks good, thanks!

PR #3617, opened 2024-06-01
relax dimension requirements

### LINKED ISSUE #3588: VAPOR can't read the WRF outputs from the CONUS404 reanalysis data set
VAPOR is unable to read the CONUS404 data set hosted on the RDA [repository](url). The outputs are a reanalysis product, written in WRF format. When trying to import any of the files in the collection VAPOR complains about the XLONG variable

shaomeng:
This PR removes a check in the WRF reader, appearing to fix #3588 .

### DIFF
--- a/lib/vdc/DCWRF.cpp
+++ b/lib/vdc/DCWRF.cpp
@@ -991,6 +991,10 @@ int DCWRF::_InitHorizontalCoordinates(NetCDFCollection *ncdfc)
         SetErrMsg("Coordinate variables XLONG and XLAT having different dimensions!\n");
         return (-1);
     }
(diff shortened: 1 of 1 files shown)


clyne (commented):
Not sure that I would completely eliminate the test. How about checking that the number of dimensions is either 2 or 3, since we believe the logic handles those two cases?

shaomeng:
> Not sure that I would completely eliminate the test. How about checking that the number of dimensions is either 2 or 3, since we believe the logic handles those two cases?

Great suggestion! The dimension check is back now with an either 2 or 3 possible values.

clyne (approved):

PR #3623, opened 2024-06-12
Error check and dump metadata function

ifranda:
I thought I'd open a small PR to test run making changes to the Python api. I've made two changes:
1. When RGB colors are converted to HSV, make a check to verify the list is of length 3. I think this stops the kernel from crashing when using incorrect inputs with `SetColorRGBList`
2. Added a `DumpVarMetadata` function for datasets. Some of the examples printed out the metadata manually, and I thought it would be nice to have it integrated into the package. 

I've tested it out and I think it's working as intended, but let me know if there's anything else I should watch out for!

### DIFF
--- a/apps/pythonapi/vapor/dataset.py
+++ b/apps/pythonapi/vapor/dataset.py
@@ -71,8 +71,8 @@ class Dataset(SmartWrapper, wrap=link.VAPoR.DataMgr):
             print(f"  {var}")
             print(f"    Dimensionality:", self.GetVarGeometryDim(var))
             print(f"    Coordinates:", self.GetVarCoordVars(var, True))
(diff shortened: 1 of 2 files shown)


NihanthCW:
Great! You can assign Stas as a reviewer and discuss any questions tomorrow.

StasJ (commented):
Looks good, thanks! I would suggest implementing `DumpVarMetadata` as `__repr__` and include dimensions and coordinates, similar to Xarray. Also please address Circle tests.

ifranda:
Just made the change! We can definitely talk more about what's being returned here if we have other ideas.

StasJ (approved):
Looks good, thanks.

PR #3624, opened 2024-06-18
Utils organization

ifranda:
This PR creates a directory for open source utility functions. The organization of this is definitely up for discussion! If we structure it in this way, we can import utils like this:

`from vapor.utils import general_utils, animation_utils`
or like this:
`from vapor.utils.general_utils import ShowMatPlotLibHistogram`


StasJ (commented):
The layout looks good to me. Two minor suggestions:
- Since the files are already in the utils submodule, they don't need the `_utils` postfix.
- Since there will be files for different types of utils, the current `general_utils.py` pertains to the histogram so it could be renamed to `utils/histogram.py`

StasJ (changes requested):
Examples that utilize the original utils module need to be updated as they still point to the old location.

ifranda:
Good catch, I think these changes should do it

PR #3629, opened 2024-07-01
Update LICENSE

NihanthCW:
replaced custom license with BSD3 license which is used across NSF NCAR. 
Doing it now for NUMFocus


StasJ (approved):
~~The original LICENSE.txt needs to be removed.~~
Wrote this, then just removed it myself, then went to approve and it saved my original message.

shaomeng (approved):

NihanthCW:
rename licence to licence.txt

PR #3630, opened 2024-07-02
Auto check/enable VGL on casper

StasJ:
- Checks for use of VGL on Casper
- If not present, show user dialog with option to auto-relaunch vapor using VGL
- Add menu option for enabling/disabling this feature which only shows on Casper

s1
s2


StasJ:
TODO: Clarify that `Relaunch` button will use VGL

sgpearse (approved):

Issue #3626, opened 2024-06-21
CAM datasets no longer supported


StasJ:
Attempting to import a CAM dataset (like `glade/data/Source/ROMS/Kauffman/b.n20.B1850CN.f09_g16.09.roms.ha.0001-03.nc`) will result in an empty visualizer and no data variables.

shaomeng:
- Reason no data variable:
  - This particular ROMS file isn't really CF compliant. As a work around, VAPOR implements some logic so the variables can still be read. Unfortunately, this logic is broken, and the variables cannot be read anymore.

- Nature of the work around:
  - code:  https://github.com/NCAR/VAPOR/blob/31fdd7881bd18d5ed08d98fcc161e6bcad4c75ce/lib/vdc/NetCDFCFCollection.cpp#L310-L357
  -  It appears that the work around is based on some logic to match certain variable names, but it's not clear how this logic really works. 
  - The code author clearly states that this file isn't CF compliant, and the code is just a work around.

- My question + thinking
  - Is it common for *modern* ROMS files to be CF compliant or not? Note that this file in question was produced in 2012.
  - I normally avoid using guess work to just make things work, especially things that are not standard compliant themselves.
  - Are there newer ROMS files to examine?

shaomeng:
Hi @StasJ , you mentioned that release 3.9.2 was having this dataset imported correctly and variables identified correctly. Could you verify if it's really the case? In my test this file appeared broken since at least 2022.

StasJ:
I tested `/glade/campaign/cisl/vast/vapor/data/Source/ROMS/jsmall` using the latest 3.9.2 release on Casper and it works.

shaomeng:
Do you mind testing the file actually report in the issue?

shaomeng:
Hey @StasJ , I tested it again, and it seems if you include the grid file in the same `Kauffman` folder (`NEP4_grid.nc`) then the dataset loads correctly, using the current main branch. Do you mind giving it a try?

StasJ:
Hey @shaomeng, you are correct, I apologize for the wild goose chase. I knew about the auxiliary grid file but I must've forgotten to import it when I reported the bug but remembered when I retested the issue. Maybe it would be worthwhile to have an error message for this scenario although I know that is easier said than done due to the heuristic nature of importing these datasets...

shaomeng:
[...]

PR #3634, opened 2024-07-13
raise a more informative message when the coordinate file is missed

shaomeng:
This PR doesn't fix, but provides a more detailed message on the coordinate file missing case, as described in #3626 .

### DIFF
--- a/lib/vdc/DCCF.cpp
+++ b/lib/vdc/DCCF.cpp
@@ -68,13 +68,17 @@ int DCCF::initialize(const vector<string> &paths, const std::vector<string> &opt
 
 int DCCF::initialize(const vector<string> &paths, const std::vector<string> &options, NetCDFCFCollection *ncdfc)
 {
-    if (_ncdfc) delete _ncdfc;
+    if (_ncdfc) {
+      delete _ncdfc;
+      _ncdfc = nullptr;
+    }
     _paths = paths;
     
     // Initialize the NetCDFCFCollection class.
     //
     int rc = ncdfc->Initialize(paths);
     if (rc < 0) {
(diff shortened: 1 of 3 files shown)


lib/vdc/DCCF.cpp
```diff
@@ -68,13 +68,17 @@ int DCCF::initialize(const vector<string> &paths, const std::vector<string> &opt
...
+      delete _ncdfc;
+      _ncdfc = nullptr;
+    }
     _paths = paths;
     
     // Initialize the NetCDFCFCollection class.
     //
     int rc = ncdfc->Initialize(paths);
     if (rc < 0) {
+        delete ncdfc;
```
StasJ, line 81: The memory leak you are fixing here is due to the function being improperly called. This will cause a segfault if it is used correctly.

StasJ (changes requested):

lib/vdc/DCCF.cpp
shaomeng: This is a downside of using raw pointers. Do you suggest to revert my change here, or to spend some effort to clean up related code in a proper manner?

StasJ: I would at least revert the change as a small memory leak is better than a segfault, however, I believe this function is currently only used in one location so it should be very easy to fix properly.

shaomeng: Good idea. Do you mind reviewing again?

StasJ (approved):

lib/vdc/DCCF.cpp
StasJ: Looks good, thanks

PR #3635, opened 2024-07-19
Fix Vapor crashes when variable names have invalid XML characters #3633

### LINKED ISSUE #3633: Vapor crashes when variable names have invalid XML characters
When variable names have invalid XML elements, vapor will crash instead of issuing a warning when creating a new renderer.

To reproduce:
1) Load NetCDF-CF file in /glade/campaign/cisl/vast/vapor/Bugs/3633
2) Create a new renderer

[User reported this issue here.](https://vapor.discourse.group/t/creating-render-causes-vapor-to-crash/391/3)

StasJ:
Support for variables with invalid XML characters cannot be added without major refactoring. The dataset in question is not CF complaint as the spec does not allow the characters used (CF Conventions § 2.3. Naming Conventions).

I added a check for CF-compliant names in the data loader.

Fix #3633

### DIFF
--- a/include/vapor/XmlNode.h
+++ b/include/vapor/XmlNode.h
@@ -501,7 +501,8 @@ public:
     // I don't know how to export an operator<< !
     static ostream &streamOut(ostream &os, const XmlNode &node);
 
-    static bool IsValidXMLElement(string s);
+    static bool IsValidXMLElement(const string &s, int *badIdx=nullptr);
+    static string SanitizeXMLTag(string s);
 
 private:
     static vector<long>   _emptyLongVec;    // empty elements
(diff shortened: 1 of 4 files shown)


shaomeng:
Hi @StasJ , I'm a little confused by the logic in this PR, and frankly what the old logic is. Do you mind providing a bit more information on why the old code caused a crash, and how the new logic prevents the crash? Thanks!

StasJ:
The crash was caused by params entries being created for variables whose names are invalid XML tags. This PR rejects datasets containing such variable names.

shaomeng (approved):
Looking good, thanks Stas!

Issue #3483, opened 2023-10-20
OSPray on M1


StasJ:
OSPRay now supports M1 builds and needs to be added to the third party library.

Issue #3269, opened 2022-11-01
Provide a native build for Apple M1 architecture


shaomeng:
The current Apple M1 installer is still targeting the X86 architecture. Relying on Apple's virtualization/instruction translation technology must have a pretty big performance hit. Providing a native build targeting the M1 architecture must bring a pretty significant performance boost. 

Note: I don't have concrete numbers on how big the actual performance difference is. It'd be interesting to use an existing software with both X86 and M1 builds to quantify the performance difference.

shaomeng:
A runtime comparison using a command-line utility [SPERR](https://github.com/NCAR/SPERR). The same software is compiled on both x86 and Arm64 macbooks, and the runtime is collected on an M1 macbook.  Runtime comparison with a series of lossy compression error tolerance levels is in the following graph. It shows that the performance difference is in the 20% to 35% range. 

![Runtime Comparison on Apple M1 System](https://user-images.githubusercontent.com/814885/204644746-e825040d-8d8f-47a0-aa83-ba888a8c54ea.png)

The raw numbers are kept in this [google sheet](https://docs.google.com/spreadsheets/d/12yc4torG_eWNbLlYMYo54KmBFXnBJQ62RS2e6fIb-zk/edit?usp=sharing).

shaomeng:
> @shaomeng Is CPU based data compression the right metric to be comparing? 
Not sure. This is the best data point I have right now.

shaomeng:
I'm not sure if the current offering of M1 build is native or not. @sgpearse can help clarify, and close the issue if it's already native M1 build.

PR #3639, opened 2024-07-25
Codesigning vapor on Apple Silicon

### LINKED ISSUE #3638: Weekly installer broken on M1
Vapor will crash after attempting to open a dataset. This only happens on a clean system.

sgpearse:
Makes changes to support a native arm64 build.  These changes are currently macro'd in order to keep supporting our current x86 build.  The macros will be removed when the x86 third party libraries have been updated to support codesigning.

After an installer is generated, scripts/codesignMacOS.sh can be run to sign all binaries, libraries, and the bundle itself.  Notarization and stapling then takes place after codesigning.  All of these steps require the use of our Developer Application ID certificate.

A large portion of this work included modifying our third party libraries.  The new libraries can be found on our AWS bucket here: https://vaporawsbucket.s3.us-west-2.amazonaws.com/2024-Jul-AppleSilicon.tar.xz

@stasj, OSPRay was upgraded to version 3, which deprecated a few of the old API calls.  Can you take a close look at what they've been replaced with?  I am seeing that the OSPRay volume renderer only renders during fast mode while moving the scene.

Fixes #3638, #3483, and #3269.

A test installer can be found here:
https://drive.google.com/file/d/1JkpBhv1WcgNvNeNuUH1HT1fU3ztARzcf/view?usp=drive_link


CMakeLists.txt
```diff
@@ -312,10 +329,11 @@ elseif (APPLE)
 	endif ()
 
     if (BUILD_PYTHON)
-	    set (CMAKE_INSTALL_RPATH "@loader_path")
-	else ()
-	    #set (CMAKE_INSTALL_RPATH "@executable_path/../lib")
-	    set (CMAKE_INSTALL_RPATH "@executable_path/../Frameworks")
+        set (CMAKE_INSTALL_RPATH "@loader_path")
+    elseif ("${CMAKE_OSX_ARCHITECTURES}" STREQUAL "arm64")
+        set (CMAKE_INSTALL_RPATH "@executable_path/../Frameworks;@executable_path/../Resources")
```
StasJ, line 334: RPATH is for finding libraries/frameworks which for an AppBundle should exist inside the Frameworks directory. Is there a reason we also need it to link to the Resources directory?

From the macOS docs:
```
Frameworks: Contains any private shared libraries and frameworks used by the executable
Resources: data files that live outside your application’s executable file. Resources typically consist of things like images, icons, sounds, nib files, strings files, configuration files, and data files
```

CMakeLists.txt
```diff
@@ -463,9 +474,22 @@ if (APPLE)
...
+            if ("${CMAKE_OSX_ARCHITECTURES}" STREQUAL "arm64")
+                file (GLOB INSTALL_GUI_FRAMEWORKS ${PYTHONDIR})
+                set (PYTHON_DESTINATION "${INSTALL_LIB_DIR}/../Resources")
+            else()
+                file (GLOB INSTALL_GUI_FRAMEWORKS ${PYTHONPATH})
+                set (PYTHON_DESTINATION "${INSTALL_LIB_DIR}")
+            endif()
 		    install (
-		    	DIRECTORY ${INSTALL_GUI_FRAMEWORKS}
-		    	DESTINATION ${INSTALL_LIB_DIR}
```
StasJ, line 468: Installed libraries should go in `${INSTALL_LIB_DIR}`

lib/common/ResourcePath.cpp
```diff
@@ -119,12 +121,16 @@ std::string Wasp::GetPythonDir()
 #ifdef WIN32
     return GetPythonPath();
 #endif
-
-    string path = GetResourcePath("");
-
-    string exists = FileUtils::JoinPaths({path, PYTHON_INSTALLED_PATH});
-
-    if (!FileUtils::Exists(FileUtils::JoinPaths({path, PYTHON_INSTALLED_PATH}))) path = string(PYTHON_DIR);
+    string path = string(PYTHON_DIR);  // Try our third-party-library directory first
```
StasJ, line 124: We want to use the installed python dir first, and only have the developer libs as a fallback.

StasJ (commented):
I notice that a bunch of our library paths are unique for the macOS arm64 build, as well as some  libraries like OSPRay have incompatible version differences. Is this temporary or permanent?

CMakeLists.txt
sgpearse: Yes, it's a codesigning requirement.  Frameworks cannot contain non-binary files like .py modules, so Python needs to reside in Resources.

CMakeLists.txt
sgpearse: I'm not sure what your implying.  The destination is indeed ${INSTALL_LIB_DIR}.

lib/common/ResourcePath.cpp
sgpearse: This is a fairly obscure part of our codebase.  I'd recommend revisiting it after 3.9.3 but we can postpone the release if it's that important.

sgpearse:
> I notice that a bunch of our library paths are unique for the macOS arm64 build, as well as some libraries like OSPRay have incompatible version differences. Is this temporary or permanent?

Definitely temporary and noted in the PR text:

> These changes [for arm64] are currently macro'd in order to keep supporting our current x86 build. The macros will be removed when the x86 third party libraries have been updated to support codesigning.

CMakeLists.txt
sgpearse: Just a followup with a reference.  This is also how Blender distributes Python.

> Store Python, Perl, shell, and other script files and other non-Mach-O executables in your app's Contents/Resources directory. While it's possible to sign such executables and store them in Contents/MacOS, this is not recommended.

https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG13

StasJ: I see the issue and I am fine keeping it this way.

To respond to your followup:

> Store Python, Perl, shell, and other script files and other **non-Mach-O executables** in your app's Contents/Resources directory. While it's possible to sign such executables and store them in Contents/MacOS, this is not recommended.

Blender's macOS app does not follow the specification at all.

CMakeLists.txt
StasJ: This line is deleted and replaced with something else.

lib/common/ResourcePath.cpp
StasJ: I'm not sure what you mean by obscure, this determines the version of python that is used by Vapor so it is rather critical. With this change it will be impossible for a developer to test an installed version of Vapor without first removing their third party library installation.

StasJ:
> Definitely temporary and noted in the PR text:

Fair point, my apologies. I read this but must've forgot about it by the time I was wrapping up the review after our meeting.

lib/common/ResourcePath.cpp
sgpearse: I understand and totally see you point.  IMO, I don't think it's critical to fix now, but it's critical to fix soon because installers are not often tested on dev computers.  This is why we do platform tests on clean (non-dev) systems.

If this solution didn't work as in this PR, I would agree to modify it as you've suggested.

This file is obscure because:
1) It heavily relies on macro'd functions that cannot be plugged into a debugger like lldb.  GetResourcePath() can't be debugged without adding print statements to see what happens
2) PYTHON_INSTALLED_PATH is also macro'd
3) CMake code injections are furthermore specified from lib/common/CMakeConfig.cpp.in.  This is as hard to debug as macro's, but the changes are coming from yet another compile time modification
4) The #ifdef WIN32 macro for GetPythonPath() should be in-lined with GetPythonDir().  What's the difference between a PythonPath and a PythonDir?Apparently it's windows is dir, unix is path?  Path and Dir are synonyms so this makes no sense.

CMakeLists.txt
sgpearse: Ok, thanks, but for the record:

> Blender's macOS app does not follow the specification at all.

Blender puts all of its libraries in the Resources directory.  Could you elaborate if I'm mistaken?

CMakeLists.txt
sgpearse: I see what you mean.  This is for the same reason in one of our previous comments, which is that Python's modules cannot be signed in the /Frameworks directory.

Alternatively, I believe we could put the libpython in /Frameworks and set PYTHONHOME to the module path.  I'm not sure if keeping all libraries in /Frameworks is worth the extra complexity and environment modification though.

CMakeLists.txt
StasJ: Correct, which directly contradicts the specification.

lib/common/ResourcePath.cpp
StasJ: I use the installers to test on my dev machine frequently when I am unsure if an issue is related to new code or my dev environment. This invalidates such a test. It also creates the situation where if someone has a different incompatible version of the third part libs installed (older or newer) vapor may try to load that version and crash. A few months down the line we may forget about this behavior and have to figure out why vapor is crashing. Since this is only going to affect developers not end users I'm fine with not holding up the release over this. Also, read obscure as *not important or well known*.

sgpearse:
I see what you're saying, but does this need to be done imminently?

> A few months down the line we may forget about this behavior

I agree and suggest that we add a new issue and handle it for the next release.  I want to emphasize that debugging this code requires a chain of events that include recompilation, codesigning, and notorization.  Adding a single print statement to debug this code is inordinately time consuming.  I've tested this with confidence but it's a huge time consuming pain.
 
This process works right now, but I agree that we can also improve it with your suggestions.  If we're going to do rapid releases, we should heavily consider flexibility.

StasJ:
> I see what you're saying, but does this need to be done imminently?

I said

> Since this is only going to affect developers not end users I'm fine with not holding up the release over this.

however, the reason why it would be preferable to do this now is once vapor 3.9.3 is released with this behavior, that version will permanently have this issue when we use it to test for regressions.

PR #3660, opened 2024-08-23
Signal to docs repository on release

ifranda:
Workflow that triggers our documentation workflow in our other repo on release. This will automate one more step of the release process.

### DIFF
--- a/.github/ISSUE_TEMPLATE/workflows/release.yml
+++ b/.github/ISSUE_TEMPLATE/workflows/release.yml
@@ -1,0 +1,22 @@
+name: Trigger Documentation Update
+
+on:
+  release:
+    types: [published]
+  push:
+    branches:
+        - Actions
+
+jobs:
+  trigger-doc-update:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout VAPOR repo
+        uses: actions/checkout@v4
+
+      - name: Trigger documentation update
+        run: |
+          curl -X POST -H "Authorization: token ${{ secrets.DISPATCH_TOKEN }}" \
+          -H "Accept: application/vnd.github.v3+json" \
+          https://api.github.com/repos/NCAR/VaporDocumentationWebsite/dispatches \
+          -d '{"event_type":"update-docs"}'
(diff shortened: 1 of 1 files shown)


sgpearse:
@ifranda Hey Ian, I thought the VaporDocumentationWebsite watched this repository for new releases in order to trigger the docs rebuild.  Is that not true?  If true, why is this needed?

ifranda:
Unfortunately I think the way actions are set up means that an action can only watch for a trigger in its own repository, so if we want to run the documentation workflow automatically we'll need a workflow here to pass along the signal. 

But automating this doesn't save that much time -- we can always run the documentation workflow manually just by clicking a button and it takes just ~5 minutes to run. And if we want to manually review the PR it creates it'll take some manual work anyway. What are your thoughts on this?

sgpearse:
> Unfortunately I think the way actions are set up means that an action can only watch for a trigger in its own repository, so if we want to run the documentation workflow automatically we'll need a workflow here to pass along the signal.

Interesting.  I thought the docs were auto-generated in [this commit](https://github.com/NCAR/VaporDocumentationWebsite/commit/16de6b95b9f47cd3ca1a3b204687d5e5d4adc6bd) by the GHA[bot] since you were on PTO.  So I guess this was manually generated by yourself?

> But automating this doesn't save that much time -- we can always run the documentation workflow manually just by clicking a button and it takes just ~5 minutes to run. And if we want to manually review the PR it creates it'll take some manual work anyway. What are your thoughts on this?

I'd say always automate stuff like this.  If the work is a one-off job then sometimes doing it manually is easier, but we will be needing to do this repeatedly on each release.  We'll still need to add an action item to verify the website, so I'll do that now.

Another helpful contribution Ian.  Thank you :)

ifranda:
> Interesting. I thought the docs were auto-generated in [this commit](https://github.com/NCAR/VaporDocumentationWebsite/commit/16de6b95b9f47cd3ca1a3b204687d5e5d4adc6bd) by the GHA[bot]

I manually went in to trigger the workflow and merge the PR for this but it only takes a few minutes. 

On a side note, we can also manually trigger the documentation workflow to generate the docs when we make a documentation update! There might be cases where this is beneficial since it's generated on a remote linux machine and always uses the latest conda releases for the sphinx packages.

PR #3671, opened 2024-10-10
Fix #3574 - Slice renderer not rendering until after timestep changed

### LINKED ISSUE #3574: Slice renderer not rendering until after timestep changed
1. On casper, open ROMS/his
2. Create/enable a slice renderer

Notice nothing is drawn

3. Step forward a timestep
4. Step back to the original timestep

Notice data has appeared

sgpearse:
The ROMS/his dataset was setting a default variable that was all 0's, causing a problem for the Slice shader's color lookup table.  This PR artificially adds .1 to the maximum data value maximum so that there is some kind of an actual range in the case where the data's minimum==maximum.

Fixes #3574

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -636,7 +636,7 @@ workflows:
       #- build_python_api_ubuntuDebug
       #- build_python_api_ubuntu
       #- build_python_api_osx
-      # - build_win10_installer
+      #- build_win10_installer
       #- build_macOS_installers
       #- build_appleSilicon_installer
       #- build_macOSx86_installer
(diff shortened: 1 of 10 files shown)


StasJ (changes requested):
This should be fixed in the slice and contour renderers rather than having DataMgr::GetDataRange return an incorrect value.

StasJ (commented):
This will present the user with incorrect information since it is writing the hacked value back into the params database. This may also still break for values >2.1M, although this is an unlikely scenario.

sgpearse:
Ok I see what you mean.  In the current state of this PR, the range modification does not affect the params database, and I'm using FLT_EPSILON instead of .1.

StasJ (commented):
Thanks for addressing this, Scott.
Some notes:
- Using FLT_EPSILON will make this not work for values > 1
- I noticed that the contour renderer does not have this issue in main.
- The 2D renderer does have this issue.

sgpearse:
Will choose an arbitrary value instead of adding a delta.

StasJ (approved):
Looks great, thanks Scott.

Issue #3677, opened 2024-11-15
Broken Windows Build


sgpearse:
PR #3676 broke the windows build.  Build link and errors below:

```
...
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\objidl.h(14893,62): error C2872: 'byte': ambiguous symbol (compiling source file C:\Users\circleci\project\lib\common\FileUtils.cpp) [C:\Users\circleci\project\build\lib\common\common.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\objidl.h(15138,62): error C2872: 'byte': ambiguous symbol (compiling source file C:\Users\circleci\project\lib\common\FileUtils.cpp) [C:\Users\circleci\project\build\lib\common\common.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\objidl.h(15157,62): error C2872: 'byte': ambiguous symbol (compiling source file C:\Users\circleci\project\lib\common\FileUtils.cpp) [C:\Users\circleci\project\build\lib\common\common.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\oaidl.h(563,26): error C2872: 'byte': ambiguous symbol (compiling source file C:\Users\circleci\project\lib\common\FileUtils.cpp) [C:\Users\circleci\project\build\lib\common\common.vcxproj]

    2 Warning(s)
    102 Error(s)
```

https://app.circleci.com/pipelines/github/NCAR/VAPOR/7796/workflows/c61fe807-0d6d-4d21-bc67-f6fa152b2286/jobs/19092

PR #3684, opened 2024-11-19
Fix broken windows build

sgpearse:
Fix issue #3677, which reports a broken windows build.  This also enables our build_win10_installer test on PR's since windows frequently breaks our build and it would be nice to find out about this at PR submission time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -651,7 +651,7 @@ workflows:
       #- build_python_api_ubuntuDebug
       #- build_python_api_ubuntu
       #- build_python_api_osx
-       #- build_win10_installer
+      - build_win10_installer
       # - build_macOS_installers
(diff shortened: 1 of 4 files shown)


StasJ (commented):
Thanks Scott!

Also I noticed that this PR enables the win10 installer for all pushes, I don't know if you mean to do that.

sgpearse:
Hey, no problem.

> Also I noticed that this PR enables the win10 installer for all pushes, I don't know if you mean to do that.

Yes, I commented on this in the main PR.  I feel like Windows keeps tending to bite us after a PR is merged.  Our CI usage has gone down so I enabled it here to save us this kind of headache in the future, but let's deliberate at today's meeting or via email/here.

StasJ (approved):

PR #3698, opened 2025-01-29
Draft for Issue1123

### LINKED ISSUE #1123: Transfer function editor improvements

sgpearse:
Draft to fix #1123 by creating new PTFEditors with an expand-window button, shown in the top left corner of the editor.

I have two questions that I'm uncertain on:

1) ParamsMgr.cpp:1315 asserts that _groups.size() > 0, but then returns in the following statement under the same condition.  Why is it throwing an assert here?  Does the return statement suffice?

2) I need to set the sizePolicy for TFMapWidgets that contain TFColorMaps and TFIsoValueMaps.  I'm doing this in the TFMapWidget constructor with dynamic casts to catch the TFColorMap/TFIsoValueMap instances.  Is there a better way to do this?

I'm currently reaching out to leigh for usability feedback.

Note that the Volume Renderer's colormapped variable transfer function is currently not expandable.  I plan on coming back to it hopefully once the currently reported regressions have been fixed.

Screen Shot 2025-01-29 at 10 09 55 AM


apps/vaporgui/PTFEditor.cpp
```diff
@@ -28,22 +29,24 @@ template class PTFMapWidget<TFIsoValueWidget>;
...
+PTFEditor::PTFEditor(const std::string &tag, const std::set<Element> elements, const std::string &label, bool expandable) : PWidget(tag, _section = new VSection(label.empty() ? tag : label)), _expandable(expandable)
 {
     _maps = new TFMapGroupWidget;
     _histogram = new TFHistogramMap(tag);
     _opacityMap = new TFOpacityMap(tag);
     _colorMap = new TFColorMap(tag);
     _isoMap = new TFIsoValueMap(tag);
     _range = new TFMappingRangeSelector(tag);
+    _elements = elements;
+    _label = label;
```
StasJ, line 41: The label parameter should be held by the VSection rather than in the subclass. Also, it is optional which is not being handled.

apps/vaporgui/PTFEditor.cpp
```diff
@@ -134,4 +142,57 @@ void PTFEditor::updateGUI() const
...
+    Update(getParams(), getParamsMgr(), getDataMgr());
+}
+
+void PTFEditor::closeExpandedPTFEditor() {
+    _expandedPTFEditor->close();
+    _expandedPTFEditor=nullptr;
+}
+
+void PTFEditor::closeEvent(QCloseEvent* event) {
+    emit closed();
```
StasJ, line 194: Why not simply call `closeExpandedPTFEditor`

apps/vaporgui/PWidget.h
```diff
@@ -55,6 +54,8 @@ class PWidget : public UWidget {
     void     setToolTip(const QString &) = delete;
 
 protected:
+    const std::string  _tag;
```
StasJ, line 57: This should remain private.

apps/vaporgui/TFMapWidget.cpp
```diff
@@ -199,6 +201,11 @@ TFMapWidget::TFMapWidget(TFMap *map)
     AddMap(map);
     setContextMenuPolicy(Qt::CustomContextMenu);
     connect(this, SIGNAL(customContextMenuRequested(const QPoint &)), this, SLOT(_showContextMenu(const QPoint &)));
+
+    // Is there a better way or place to set the sizePolicy of this object?
```
StasJ, line 205: You could do it in the subclass constructor

apps/vaporgui/VSection.cpp
```diff
@@ -75,3 +99,28 @@ void VSection::SettingsMenuButton::paintEvent(QPaintEvent *event)
...
+    // This function is overridden to prevent Qt from drawing its own dropdown arrow
+    QStylePainter p(this);
+
+    QStyleOptionToolButton option;
+    initStyleOption(&option);
+    option.subControls = QStyle::SC_ToolButton;
+    option.features = QStyleOptionToolButton::None;
+    p.drawComplexControl(QStyle::CC_ToolButton, option);
+}
+
```
StasJ, line 126: `ExpandSectionButton` duplicates the code of `SettingsMenuButton` apart from changing the icon path.

lib/params/ParamsMgr.cpp
```diff
@@ -1312,7 +1312,8 @@ void ParamsMgr::PMgrStateSave::EndGroup()
 
     if (!GetEnabled()) return;
 
-    assert(_groups.size());
+    // If we are going to return in the next statement, why also assert for the same condition?
+    //assert(_groups.size());
```
StasJ, line 1316: This assert needs to remain here. I think the intention was to have it be a VAssert so that it would only crash in debug builds and silently fail on release builds.

apps/vaporgui/VSection.h
```diff
@@ -19,11 +19,13 @@ class VSection : public QTabWidget {
     Q_OBJECT
 
     class SettingsMenuButton;
+    class ExpandSectionButton;
 
 public:
     VSection(const std::string &title);
     QVBoxLayout *layout() const;
     void         setMenu(QMenu *menu);
+    void         setExpandedSection();
```
StasJ, line 28: Method is called `set...` but nothing is being set.

StasJ (changes requested):
Crash:
1. Create/enable a volume renderer
2. Expand TF editor
3. Create/enable a 2d renderer
4. Delete the first renderer
5. Edit the first renderers TF in the expanded view
6. Vapor will crash
---
The resolution of the histogram needs to be limited/performance improved as it becomes noticeably slow at fullscreen retina.

sgpearse:
Thanks @stasj.  I'll address/reply to your comments shortly.

apps/vaporgui/PTFEditor.cpp
sgpearse: We need to specifically capture the expanded PTFEditor's closeEvent so the containing PTFEditor can then set the expanded to nullptr.

If we don't, then the checks for _expandedPTFEditor==nullptr will fail, particularly in PTFEditor::showExpandedPTFEditor().

apps/vaporgui/TFMapWidget.cpp
sgpearse: The subclass isn't a QWidget though.  It needs to be done in the TFMapWidget, not the TFMap :\

lib/params/ParamsMgr.cpp
sgpearse: I'm seeing this happen in other places in the code, so I'll file issue #3717.

apps/vaporgui/VSection.h
sgpearse: QTabWidget::setCornerWidget() is being called though(?).

StasJ: For example, the method above it called "setMenu" sets the menu to the passed parameter. Maybe `enableExpandedSection` or `setExpandable(bool)` would be more appropriate?

apps/vaporgui/TFMapWidget.cpp
StasJ: TFMap could have a virtual `GetSetSizePolicy` method.

apps/vaporgui/PTFEditor.cpp
StasJ: So you're saying when PTFEditor receives its own `close` event in PTFEditor::closeEvent, it needs to then re-emit different `closed` event which it can then capture again in PTFEditor::closeExpandedPTFEditor where it tells the expanded window to close?

StasJ (commented):
Thanks Scott. Just reconsider the comment in PTFEditor::closeEvent, the other two are optional suggestions. Finally, I noticed in macOS dark mode, the icon is barely visible (admittedly, the settings icon is not great either)

screenshot

apps/vaporgui/PTFEditor.cpp
sgpearse: No, the PTFEditor is not receiving its own close event.  When the PTFEditor's _expanded instance_ closes, the containing PTFEditor must to set the _expanded instance_ to nullptr.  See the connection being made at PTFEditor.cpp:175.

apps/vaporgui/TFMapWidget.cpp
sgpearse: Ok, that's a good idea.

sgpearse:
@stasj all comments are addressed.  The dark mode issue will work as a cross platform fix but I'm curious if you have other thoughts on it.  The PTFEditor::closeEvent signal is necessary for this PR to work.  Here's my comment from the conversation:

> So you're saying when PTFEditor receives its own close event in PTFEditor::closeEvent, it needs to then re-emit different closed event which it can then capture again in PTFEditor::closeExpandedPTFEditor where it tells the expanded window to close?
> 

> > No, the PTFEditor is not receiving its own close event. When the PTFEditor's _expanded instance_ closes, the containing PTFEditor must to set the _expanded instance_ to nullptr. See the connection being made at PTFEditor.cpp:175.

apps/vaporgui/PTFEditor.cpp
StasJ: I see. There are two instances of the PTFEditor with slightly different behavior sending signals to each other. Could the window container behavior be extracted into its own class?

StasJ:
Responded in the original thread.

apps/vaporgui/PTFEditor.cpp
sgpearse: I've extracted the signal into a new subclass, ExpandedPTFEditor.

StasJ (approved):
It doesn't need to be a subclass of PTFEditor since the closing behavior is not a part of it, just a VContainer would do but this is fine.

PR #3719, opened 2025-03-05
Restructure the GUI - Fixes Issue3603

### LINKED ISSUE #3603: Export options warrant a space on the toolbar
Exporting (Capture) the image/ animation is a key-step in the VAPOR visualization workflow. I've noticed users often enquire about how to do this, which makes me wonder if we need to place these options in the toolbar. This might be worth discussing in the context of key framing enhancement.

sgpearse:
Fixes #3603 by changing the GUI to what is shown below.

Two considerations:
1) PWidget::ShowBasedOnParam(string, string) now can accept strings for comparing parameter values.  I am finding that I need to also set the ShowBasedOnParam(string, int) to a non-zero value for things to work.   PWidget::_enableBasedOnParamValue is getting a default applied when constructed, which it should not to my understanding (shouldn't it be garbage?).

2) MainForm has several utilities for importing data and capturing imagery.  I've made the necessary functions public, but this means that PImportData and PCaptureWidget need to hold pointers to the MainForm.  

image


apps/vaporgui/RendererInspector.cpp
```diff
@@ -27,6 +27,9 @@ void RendererInspector::Update()
 {
     ParamsMgr *paramsMgr = _ce->GetParamsMgr();
     GUIStateParams *guiParams = (GUIStateParams *)paramsMgr->GetParams(GUIStateParams::GetClassType());
+    bool noDatasetLoaded = _ce->GetParams<GUIStateParams>()->GetOpenDataSetNames().empty();
+    if (noDatasetLoaded) return;
```
StasJ, line 31: If there is no dataset loaded, this should not be updated in the first place.

apps/vaporgui/RenderersPanel.cpp
```diff
@@ -18,10 +18,12 @@ RenderersPanel::RenderersPanel(VAPoR::ControlExec *ce)
     _splitter->setChildrenCollapsible(false);
     _inspectorRouter->Add(_renInspector = new RendererInspector(_ce));
     _inspectorRouter->Add(_dataInspector = new DatasetInspector(_ce));
+    setEnabled(false);
 }
 
 void RenderersPanel::Update()
 {
+    setEnabled(true);
```
StasJ, line 26: Again, this was already being handled and will this leads to crash documented in main comment.

apps/vaporgui/VLabel.h
```diff
@@ -6,10 +6,16 @@
...
+protected:
     QLabel *_ql;
 
 public:
     VLabel(const std::string &text = "");
     void SetText(const std::string &text);
     void MakeSelectable();
 };
+
+class VHyperlink : public VLabel {
```
StasJ, line 18: This deserves its own file

apps/vaporgui/VLineItem.cpp
```diff
@@ -28,6 +28,10 @@ VLineItem::VLineItem(const std::string &label)
...
     layout()->setContentsMargins(_LEFT_MARGIN, _TOP_MARGIN, _RIGHT_MARGIN, _BOTTOM_MARGIN);
 
-    QLabel *qLabel = new QLabel(label.c_str());
-    layout()->addWidget(qLabel);
+    _qLabel = new QLabel(label.c_str());
+    layout()->addWidget(_qLabel);
+}
+
+void VLineItem::SetLabelText(const std::string &text) {
+    _qLabel->setText(QString::fromStdString(text));
```
StasJ, line 36: Line Items are a widget and a label describing that widget. For your use case here, you need to use two labels in a horizontal layout rather than repurposing the line time.

include/vapor/STLUtils.h
```diff
@@ -42,7 +42,7 @@ COMMON_API std::string ToLower(std::string str);
 COMMON_API std::vector<std::string> Split(std::string str, const std::string &delimeter);
 COMMON_API std::string Join(const std::vector<std::string> &parts, const std::string &delimeter);
 COMMON_API std::string ReplaceAll(std::string source, const std::string &oldSegment, const std::string &newSegment);
-
+COMMON_API std::string GetUserTime();
```
StasJ, line 45: Name could be confused for user data time.

lib/params/GUIStateParams.cpp
```diff
@@ -46,6 +46,8 @@ const string GUIStateParams::_flowDimensionalityTag = "_flowDimensionalityTag";
 const string GUIStateParams::BookmarksTag = "BookmarksTag";
 const string GUIStateParams::MovingDomainTrackCameraTag = "MovingDomainTrackCameraTag";
 const string GUIStateParams::MovingDomainTrackRenderRegionsTag = "MovingDomainTrackRenderRegionsTag";
+const string GUIStateParams::SelectedImportDataTypeTag = "SelectedImportDataTypeTag";
+const string GUIStateParams::SelectedImportDataFilesTag = "SelectedImportDataFilesTag";
```
StasJ, line 50: Do these need to be persistent state?

MattRehme (changes requested):
Much nicer workflow with this layout. It will definitely be easier for new users to get started.

Same comments as Nihanth on these points:
* Selecting .png still saves as .tif, both for a single image and still image
* Change "Annotation" to "Colorbar", since Colorbar settings are the only thing available in that tab.


Also, I'm not sure if this would be another PR, but the "set/go to home view" and "view all" buttons don't look good in Mac dark mode due to the aliasing around the icons (see attached). Not sure what could be done there.
thumbnails

apps/vaporgui/ImportTab.h
```diff
@@ -0,0 +1,24 @@
...
+
+public:
+    ImportTab(VAPoR::ControlExec *ce, MainForm *mf);
+    void Update() override;
+
+public slots:
+    void DataImported();
+
+signals:
+    void dataImported();
```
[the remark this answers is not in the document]
sgpearse: Removed.

apps/vaporgui/PCaptureWidget.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+#include <QFileDialog>
+#include <QHBoxLayout>
+
+const std::string CaptureModes::CURRENT = "Current frame";
+const std::string CaptureModes::RANGE   = "Timeseries range";
+
+const std::string TiffStrings::CaptureFileType = "TIFF";
+const std::string TiffStrings::FileFilter = "TIFF (*.tif)";
+const std::string TiffStrings::FileSuffix = "tif";
```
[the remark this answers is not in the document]
sgpearse: I've removed half of them, but the remaining ones are used in multiple places.  Just trying to avoid magic strings.

apps/vaporgui/LeftPanel.cpp
```diff
@@ -1,12 +1,15 @@
...
+#include "vapor/ControlExecutive.h"
 #include "AnnotationEventRouter.h"
-#include "AnimationTab.h"
-#include "ViewpointTab.h"
+#include "vapor/GUIStateParams.h"
 
 #include <QScrollArea>
 
-LeftPanel::LeftPanel(ControlExec *ce)
+LeftPanel::LeftPanel(ControlExec *ce, MainForm *mf) : _ce(ce)
```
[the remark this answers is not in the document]
sgpearse: Will implement in 3.9.5 due to complexity.

lib/params/GUIStateParams.cpp
sgpearse: I think it's convenient for the user.  If they import a data type once, they'll probably want to import the same type again if needed.

apps/vaporgui/PCaptureWidget.cpp
```diff
@@ -0,0 +1,137 @@
...
+PCaptureWidget::PCaptureWidget(VAPoR::ControlExec *ce, MainForm *mf)
+    : PWidget("", _section = new PSection("Image(s)")), _ce(ce), _mf(mf)
+{
+    
+    _section->Add(new PRadioButtons(AnimationParams::CaptureModeTag, {CaptureModes::CURRENT, CaptureModes::RANGE}));
+
+    PTimeRangeSelector *t = new PTimeRangeSelector(_ce);
+    // Need to set ShowBasedOnParam for a numeric value.  See previous comment.
+    t->EnableBasedOnParam(AnimationParams::CaptureModeTag, -1);
+    t->EnableBasedOnParam(AnimationParams::CaptureModeTag, CaptureModes::RANGE);
```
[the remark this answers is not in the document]
sgpearse: Discussed in meeting.  Deferring to setting enums instead of strings.

apps/vaporgui/PFileSelector.cpp
```diff
@@ -46,14 +49,24 @@ void PFileSelector::buttonClicked()
...
+    std::string defaultPath = getDefaultPath();
+    QStringList qSelectedPaths = selectPaths(defaultPath);
+    if (qSelectedPaths.isEmpty()) return;
 
-    setParamsString(qSelectedPath.toStdString());
+    std::vector<std::string> paths;
+    paths.reserve(qSelectedPaths.size());
+    for (const QString& qPath : qSelectedPaths) paths.push_back(qPath.toStdString());
+    getParams()->SetValueStringVec(getTag(), "", paths);
+    emit filesSelected();
```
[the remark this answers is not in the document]
sgpearse: I've removed this.  However, where is this defined?  There are a few PWidgets that emit and receive signals, and we aren't defining this with anything like QWidget::blockSignals(true).

apps/vaporgui/PGroup.h
```diff
@@ -25,6 +25,13 @@ class PGroup : public PWidget, public WidgetGroupWrapper<PGroup, PWidget, VGroup
     void updateGUI() const override;
 };
 
+class PHGroup : public PGroup {
```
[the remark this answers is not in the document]
sgpearse: This was an attempt on something I changed my mind on at a later point.  Removed.  Thanks for the catch.

sgpearse:
@NihanthCW and @MattRehme Thanks for your feedback guys.  Everything is fixed except for 

1) Increasing the font size of the top level tabs.  I gave this a shot but the text was being rendered outside of the tab and didn't look good.  This would need further investigation.

2) Generating new icons for dark mode.  This PR is pretty big as it is and segregating that out for later is preferable.

sgpearse:
@StasJ Thank you for your detailed review.  Individual comments should be replied to or resolved.  Regarding the comments in your primary review:

> If vapor is started with data or a session, or when a session is opened, it should default to the render page.

Done by setting GUIStateParams::DataJustLoadedTag, so we're not passing signals throughout the GUI as you mention.

> This could be a separate PR later on, but since we now have an explicit separation between the render stage and annotation stage, should we move per-renderer annotations under the annotate section?

I think that makes sense.  We can discuss.  I'd prefer doing this in a separate PR.
 
> After pressing "Capture Time series", it says "Saved: <file> At: time" but the file name is not correct.

Fixed.
 
> Should camera controls be under Export since they are related to the viewport outside of exporting? Same with moving domain settings.

The current layout is what was specified in the issue but we can reconsider it with the team. 

> Minor feature request: remember last selected output folder.

Check.
 
> Bug: Under animation enable "loop playback". Press capture time series. Vapor will indefinitely save images.

Removed as we discussed earlier this week.
 
> Bug: Import files does not respect data directory

This now does not accept a data directory.  User needs to navigate in a FileDialog.
 
> The text field next to import files is somewhat confusing: 
> It is not edible
> Once you "select" a dataset it gets imported and then the text field sits populated but doesn't do anything.

Should be addressed now.

> Capture Time series should not be limited by max animation frames per second.
> Same for animation play step size.
 
FPS and step size have been removed from the GUI.  I've kept max animation FPS to the default of 10.

> Bug: Changing the time series range changes the current tilmestep display in the rest of the application but it does not actually change the visualizer timestep. (This is correct, it should not affect the visualizer tilmestep imo).
 
I'm having trouble debugging this due to #3731.  However we should discuss the desired behavior as a team.  IMO, if a user specifies a subset of time to render, and the current time step is outside of that range, we should go to the nearest time step that's within the specified range.  I think the user would want to be presented with a time step that's within the range.

> I believe "Time series" is two words but it is spelled as one word throughout the ui.

Good point.  I've been getting that wrong without noticing.

> Each dataset now has a "Data Projection" menu but this is a global property, not per-dataset.
 
Projection string parameters are now held within the Import tab (and shown after data is loaded) and also accessible in the Tools menu.

> Bug: open a dataset. Create a new session. Click on the render list. Vapor will crash.

Thanks for catching that.  Fixed.

include/vapor/STLUtils.h
sgpearse: Renamed to GetCurrentDateTimestamp()

NihanthCW (changes requested):
Looks good. Just one issue and a question
- Why does the "capture current frame" button look different than the rest? 
- Bug: Open Data Projection window from "tools" -> Close the data projection window. It keeps popping up with every mouse movement in the visualizer despite closing it each instance

https://github.com/user-attachments/assets/ec8814d3-8602-4ae9-8cac-a1498c56b1d3

sgpearse:
> Why does the "capture current frame" button look different than the rest?

That's a Qt implementation issue.  Those buttons are on 2 lines, which is giving the square box.

> Bug: Open Data Projection window from "tools" -> Close the data projection window. It keeps popping up with every mouse movement in the visualizer despite closing it each instance

Thanks for catching this.

I'll look into these issues tomorrow.

apps/vaporgui/PCaptureWidget.cpp
```diff
@@ -16,119 +17,124 @@
...
+    layout->addWidget(_captureTimeSeriesButton,1);
+}
+
+void PCaptureHBox::_dropdownIndexChanged(int index) {
+    int value;
+    if (_enumMap.empty()) {
+        value = index;
+    } else {
+        VAssert(index >= 0 && index < _enumMap.size());
+        value = _enumMap[index];
```
StasJ, line 55: The VAssert is intended to be used in scenarios where something should crash in debug but silently default or alert in release builds but this will crash in both cases.

apps/vaporgui/PCaptureWidget.cpp
```diff
@@ -16,119 +17,124 @@
...
-    _typeCombo->SetValue(ap->GetValueString(AnimationParams::CaptureTypeTag, TiffStrings::CaptureFileType));
+    _typeCombo->Update(ap);
 
     // Format the label for the saved file
-    std::string file = "";
-    std::string time = "";
-    if (ap->GetValueString(AnimationParams::CaptureModeTag, "") == CaptureModes::CURRENT) {
+    std::string file, time;
+    std::string dir = ap->GetValueString(AnimationParams::CaptureFileDirTag, "");
+    if (ap->GetValueLong(AnimationParams::CaptureModeTag, 0) == 0) { // Capture signle frame
```
StasJ, line 67: Enum for capture mode please.

apps/vaporgui/PCaptureWidget.cpp
```diff
@@ -16,119 +17,124 @@
...
-    ap->SetValueString(AnimationParams::CaptureTimeseriesFileNameTag, "Capture animation file name", fileName);
-    ap->SetValueString(AnimationParams::CaptureTimeseriesTimeTag, "Capture file time", STLUtils::GetUserTime());
+    ap->SetValueString(AnimationParams::CaptureFileDirTag, "Capture animation file directory", FileUtils::Dirname(fileName));
+    ap->SetValueString(AnimationParams::CaptureTimeSeriesFileNameTag, "Capture animation file name", FileUtils::Basename(fileName));
+    ap->SetValueString(AnimationParams::CaptureTimeSeriesTimeTag, "Capture file time", STLUtils::GetCurrentDateTimestamp());
 
-    _mf->SetTimeStep(ap->GetStartTimestep());
     bool rc = _mf->StartAnimCapture(fileName, defaultSuffix);  // User may cancel to prevent file overwrite, so capture return code
-    if (rc) _mf->AnimationPlayForward();
+    std::cout << "rc " << rc << std::endl;
```
StasJ, line 121: Debug print left in.

apps/vaporgui/PImportDataButton.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "VLabel.h"
+#include "PButton.h"
+#include "DatasetTypeLookup.h"
+#include "vapor/ControlExecutive.h"
+#include "vapor/GUIStateParams.h"
+#include "vapor/FileUtils.h"
+
+#include <QFileDialog>
+
+PImportDataButton::PImportDataButton(VAPoR::ControlExec* ce, MainForm *mf) : PWidget("", _hBox = new VHBoxWidget()), _ce(ce), _mf(mf) {
```
StasJ, line 13: Please don't pass around main form.

apps/vaporgui/PImportDataButton.h
```diff
@@ -0,0 +1,29 @@
...
+
+namespace VAPoR {
+    class ControlExec;
+}
+class MainForm;
+class VHBoxWidget;
+class VLabel;
+class PButton;
+
+class PImportDataButton : public PWidget {
```
StasJ, line 13: The import data button probably doesn't need to represent state within the database so why make a it params widget?

StasJ (commented):
> > Why does the "capture current frame" button look different than the rest?

> That's a Qt implementation issue. Those buttons are on 2 lines, which is giving the square box.

This could probably just be called "export" and that would solve the issue.

---

Thanks for addressing the comments so far.

apps/vaporgui/PFileSelector.cpp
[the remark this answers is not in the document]
StasJ: I don't believe there are apart from one you've added in your last PR.

apps/vaporgui/LeftPanel.cpp
```diff
@@ -35,22 +34,30 @@ LeftPanel::LeftPanel(ControlExec *ce, MainForm *mf) : _ce(ce)
...
-}
 
-void LeftPanel::UpdateImportMenu() {
-    setEnabled(true);
-    _importTab->setEnabled(true);
-    _importTab->Update();
+    if (gsp->GetValueLong(GUIStateParams::DataJustLoadedTag, 0) != 0) {
+        setCurrentIndex(1); // go to Renderer tab
+        gsp->SetValueLong(GUIStateParams::DataJustLoadedTag, "Unsetting data just loaded tag", 0);
+    }
```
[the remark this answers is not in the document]
sgpearse: Signals and parameters have been removed for this.  The MainForm is taking care of it when opening a session or importing data.

apps/vaporgui/MainForm.cpp
```diff
@@ -1149,6 +1159,17 @@ void MainForm::launchPythonVariables()
...
+void MainForm::launchProjectionDialog()
+{
+    if (_projectionSection == nullptr){
+        _projectionSection = new PProjectionStringSection(_controlExec);
+        _projectionSection->adjustSize();
+        _guiStateParamsUpdatableElements.insert(_projectionSection);
+    }
+    _projectionSection->Update(GetStateParams());
+    _projectionSection->show();
+}
```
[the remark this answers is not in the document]
sgpearse: This is true.  Before I saw your comment I made another non modal implementation that I think is suitable also.

apps/vaporgui/PImportDataButton.cpp
sgpearse: Will be removed in 3.9.5 per meeting disuccsion.

sgpearse:
@stasj, @NihanthCW, and @MattRehme: Comments have been addressed.

I've also fixed #3735, which occurred due to CircleCI upgrading their windows machine-agent version.  We don't need special handling for windows' fun insistence of using backslashes in CPack anymore, and it turns out that our special handling for them will actually break the build.  See VAPOR/CMakeLists.txt for where this logic was removed

NihanthCW (approved):

apps/vaporgui/LeftPanel.cpp
```diff
@@ -35,22 +34,35 @@ LeftPanel::LeftPanel(ControlExec *ce, MainForm *mf) : _ce(ce)
 
 void LeftPanel::Update()
 {
-    bool noDatasetLoaded = _ce->GetParams<GUIStateParams>()->GetOpenDataSetNames().empty();
-    if (noDatasetLoaded) return;
+    GUIStateParams *gsp = _ce->GetParams<GUIStateParams>();
+    // Enable the only import panel when no data is loaded
+    bool noDatasetLoaded = gsp->GetOpenDataSetNames().empty();
+    if (noDatasetLoaded) { 
+        setEnabled(true);
```
StasJ, line 41: MainForm has an explicit mechanism for controlling whether widgets are enabled based on data being loaded and LeftPanel is controlled by this mechanism. Please update the existing mechanism rather than fighting it here.

apps/vaporgui/MainForm.cpp
```diff
@@ -767,22 +734,21 @@ void MainForm::helpAbout()
...
 {
     _paramsMgr->BeginSaveStateGroup("Import Dataset");
     if (name.empty()) name = _getDataSetName(files[0], existsAction);
-    if (name.empty()) return;
     int rc = _controlExec->OpenData(files, name, format);
     if (rc < 0) {
         _paramsMgr->EndSaveStateGroup();
         MSG_ERR("Failed to load data");
-        return;
+        return -1;
```
StasJ, line 745: This change appears to be no longer used.

apps/vaporgui/MainForm.cpp
```diff
@@ -987,10 +957,6 @@ bool MainForm::eventFilter(QObject *obj, QEvent *event)
         _paramsWidgetDemo->Update(GetStateParams(), _paramsMgr);
 #endif
 
-        setUpdatesEnabled(false);
-        _controlExec->SyncWithParams();
-        setUpdatesEnabled(true);
```
StasJ, line 995: Please consider rebasing in the future as merging introduces unrelated changes which can make followup PR reviews more difficult.

StasJ (changes requested):
Bug:
1. Open a dataset with multiple timesteps
2. Under Export > Images select a "time series range" subset to export
It is now impossible to access the remainder of time steps outside of this subset.
---
Bug:
1. Open dataset/enable renderer
2. Open dataset projection tool menu
3. Create new session
4. Import another dataset
The two projection widgets will contradict each other

apps/vaporgui/MainForm.cpp
sgpearse: Agreed.  Sorry about that.

StasJ: No worries, thanks!

apps/vaporgui/LeftPanel.cpp
StasJ: This contradicts `QWidget::setEnabled(bool)`

StasJ (changes requested):
Bug:
1. Open a dataset/enable a renderer
2. Go to export and select a subrange of time steps
3. Press export
You will get double the output images and the old image capture feature will now be enabled where every action is exported.

---

Bug:
1. Open a dataset with multiple timesteps and enable a renderer
2. Click on the export tab.
Vapor will sometimes crash due to a memory error. I was able to reproduce this twice but not in a debugger. I was able to get it to crash the second time by clicking on the other tabs and then it crashed when clicking on the export tab again. I got the following errors:

Crash 1:
```
zsh: segmentation fault  ./bin/vapor
```
Crash 2:
```
libc++abi: terminating due to uncaught exception of type std::bad_alloc: std::bad_alloc
```

apps/vaporgui/LeftPanel.cpp
sgpearse: I don't understand the problem here.  We need to always enable the import tab, even when setEnabled(false) is issued.

sgpearse:
@stasj Thanks for reporting the bugs.  Not sure how I missed the second one, but it looks like it's a regression on main so I submitted #3744.  The first one is fixed.

I don't understand why LeftPanel's setEnabled logic is contradictary because we always need to enable the Import tab, even when setEnabled(false) is called.  Could you please explain?

StasJ:
> I don't understand why LeftPanel's setEnabled logic is contradictary because we always need to enable the Import tab, even when setEnabled(false) is called. Could you please explain?

Please read the [documentation for QWidget::setEnabled](https://doc.qt.io/qt-6/qwidget.html#enabled-prop) which your implementation directly contradicts. Additionally, calling `setEnabled(false)` calls `setEnabled(true)`

StasJ (approved):

PR #3742, opened 2025-04-03
Adds DrawInFront parameter to RenderParams

sgpearse:
This allows the Image, TwoDData, and Barb renderers to be rendered without depth.  I thought these were the only renderers that needed it, but it may be useful for contours as well.

Some of the Visualizer::paintEvent() code was separated into its own function call for readability, and site.local was edited to make building on Casper more easy.


StasJ (commented):
This feature may be confusing for users, especially if they have multiple renderers using it.

PR #3746, opened 2025-04-08
Vapor transfer function widget

ifranda:
Python function for interactive transfer function control.

Expected usage:
`from vapor.utils.histogram import transferFunctionWidget
transferFunctionWidget(ses, ren, 6)`

image


ifranda:
Issues that I'm working on:
1. Not able to find anywidget
2. Opacities between 0 and 1 are temperamental, but I think I have a fix on the way

PR #3759, opened 2025-04-17
Fix import data label upon creating a new session

sgpearse:
Moves label setting logic in the Import panel when creating a new session.


NihanthCW (approved):
Looks good!

StasJ (approved):

Issue #3744, opened 2025-04-07
Capture image sequence saves files twice


sgpearse:
On main, load any dataset and capture an image sequence.  Files will be saved twice.

PR #3770, opened 2025-05-07
Fix Issue3768 - VAPOR freezes after rendering

sgpearse:
Calling install/removeEventFilter when turning animation on or off was leaving the GUI in an unresponsive state.  This was recently done to prevent Render() from being called multiple times during an animation (#3744).  I've removed this logic and am only making a Render() call in a ParamsChangeEvent when animation is turned off.  Render() now gets issued during an AnimationDrawSignal emission during animation, not through the eventFilter.

The AnimationController's timer is also stopped when animation is complete.  Previously, it was issuing timer events to the application when not being used.

### DIFF
--- a/apps/vaporgui/AnimationController.cpp
+++ b/apps/vaporgui/AnimationController.cpp
@@ -135,6 +135,7 @@ void AnimationController::setPlay(int direction)
         // Done animating. Disable timer and send notification
         //
         disconnect(_myTimer, 0, 0, 0);
+        _myTimer->stop();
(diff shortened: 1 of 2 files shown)


StasJ (changes requested):
Thanks for looking into this Scott, but this PR just mitigates the symptoms and does not address the underlying issue. I can look into this further.

sgpearse:
Interesting.  I'm curious what the underlying issue is and to see what you find.

PR #3779, opened 2025-05-20
Issue3778 - Separate MainForm from LeftPanel

sgpearse:
Separates MainForm's image-capture and import-data functionality into the classes CaptureController and DatasetImporter so these tools can be used by the LeftPanel.

Submitted as a draft because of one question: MainForm.cpp:1246 (on main) has the following logic:

```
    string          vizName = p->GetActiveVizName();
    if (vizName != _capturingAnimationVizName) { MSG_WARN("Terminating capture in non-active visualizer"); }
    if (_controlExec->EnableAnimationCapture(_capturingAnimationVizName, false)) MSG_WARN("Image Capture Warning;\nCurrent active visualizer is not capturing images");
```

Is this warning still achievable?  I am not able to produce it when changing visualizers during an image capture, or any other way I can think of.  I've removed it in this PR, but we can keep it if the MainForm communicates _capturingAnimationVizName to the CaptureController whenever it changes.  It was added in 2017 and does not appear to be needed anymore, from what I can tell.

### DIFF
--- a/apps/vaporgui/MainForm.cpp
+++ b/apps/vaporgui/MainForm.cpp
@@ -153,18 +153,15 @@ MainForm::MainForm(vector<QString> files, QApplication *app, bool interactive, s
         _controlExec->SyncWithParams();
         updateUI();
         setUpdatesEnabled(true);
-
         Render(false, true);
     });
 
     _captureController = new CaptureController(_controlExec);
-    //connect(_captureController, SIGNAL(animationCaptureStarted()), _animationController, SLOT(AnimationPlayForward()));
     connect(_captureController, &CaptureController::animationCaptureStarted, [this]() {
-        _animationController->AnimationPlayForward();
-
         GUIStateParams *p = (GUIStateParams*)_paramsMgr->GetParams(GUIStateParams::GetClassType());
         _capturingAnimationVizName = p->GetActiveVizName();
         _animationCapture = true;
+        _animationController->AnimationPlayForward();
     });
 
     _leftPanel = new LeftPanel(_controlExec, _captureController, this);
@@ -273,7 +270,7 @@ int MainForm::RenderAndExit(int start, int end, const std::string &baseFile, int
     auto vpp = _paramsMgr->GetViewpointParams(GetStateParams()->GetActiveVizName());
 
     _paramsMgr->BeginSaveStateGroup("test");
-    StartAnimCapture(baseFileWithTS);
+    //StartAnimCapture(baseFileWithTS);
     ap->SetStartTimestep(start);
     ap->SetEndTimestep(end);
 
@@ -281,11 +278,23 @@ int MainForm::RenderAndExit(int start, int end, const std::string &baseFile, int
     vpp->SetValueLong(vpp->CustomFramebufferWidthTag, "", width);
     vpp->SetValueLong(vpp->CustomFramebufferHeightTag, "", height);
 
-    _animationController->AnimationPlayForward();
+    //_animationController->AnimationPlayForward();
+    GUIStateParams *p = (GUIStateParams*)_paramsMgr->GetParams(GUIStateParams::GetClassType());
+    _capturingAnimationVizName = p->GetActiveVizName();
+    _animationCapture = true;
+    if (_captureController->EnableAnimationCapture(baseFileWithTS)) _captureController->StartAnimationCapture();
+    else {
+        _animationCapture = false;
+        _capturingAnimationVizName = "";
+    }
+    
     _paramsMgr->EndSaveStateGroup();
 
     connect(_animationController, &AnimationController::AnimationOnOffSignal, this, [this]() {
-        endAnimCapture();
+        _captureController->EndAnimationCapture();
+        //endAnimCapture();
+        _animationCapture = false;
+        _capturingAnimationVizName = "";
         close();
     });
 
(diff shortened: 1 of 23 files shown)


apps/vaporgui/PImportDataButton.cpp
```diff
@@ -34,7 +34,7 @@ void PImportDataButton::_importDataset() {
     if (fileNames.empty()) return;
 
     std::string format = GetDatasets()[getParams()->GetValueLong(GUIStateParams::ImportDataTypeTag, 0)].first;
-    _mf->ImportDataset(fileNames, format);
+    _di->ImportDataset(fileNames, format);
```
StasJ, line 37: There are a few objects that are prevalent enough to have a universally understood shorthand name, such as `ce`, however, for general variables, names such as `cc` and `di` are ambiguous.

StasJ (changes requested):
This introduces duplication of application state, for example through `_sessionNewFlag`.

The objects that were added to replace coupling MainForm are themselves tightly coupled to MainForm. Most of their functionality can be stateless and the few things that need to be communicated with MainForm are events and can therefore just use signals.

With regard to the code snippet in your PR description, have you tried reproducing the issue with multiple visualizers?

sgpearse:
Thanks for your review @stasj.  I can fix the code comments no problem.  I do want to talk about this overall approach and the points you raise in your review.

> This introduces duplication of application state, for example through _sessionNewFlag.

This is true.  I think the correct fix is to put these state flags into GUIStateParams.  Do you agree?

> The objects that were added to replace coupling MainForm are themselves tightly coupled to MainForm. Most of their functionality can be stateless and the few things that need to be communicated with MainForm are events and can therefore just use signals.

While the MainForm is coupled to these classes, separating this logic makes MainForm more cohesive.  As we know, MainForm is over a thousand lines long and it violates many OO principles.  Single Responsibility Principle, from SOLID, especially.  It's hard to test, understand, and modify because it's doing too much, making it very fragile.  This was a big pain point in our last release.  Using signals would keep data-import and image-capture logic within the MainForm.

What is the benefit of keeping this logic in the MainForm, versus breaking it out?

> With regard to the code snippet in your PR description, have you tried reproducing the issue with multiple visualizers?

Yes, noted this in the PR submission.  I'm only able to reproduce this on older versions of VAPOR where we need to "End Capture Sequence" manually.

StasJ:
> This is true. I think the correct fix is to put these state flags into GUIStateParams. Do you agree?

`_sessionNewFlag` is ephemeral state and does not need to be persisted, therefore I would keep a single instance as a regular variable, however, putting it in the GUIStateParams would be fine.

> While the MainForm is coupled to these classes, separating this logic makes MainForm more cohesive. As we know, MainForm is over a thousand lines long and it violates many OO principles. Single Responsibility Principle, from SOLID, especially. It's hard to test, understand, and modify because it's doing too much, making it very fragile. This was a big pain point in our last release. Using signals would keep data-import and image-capture logic within the MainForm.
What is the benefit of keeping this logic in the MainForm, versus breaking it out?
With regard to the code snippet in your PR description, have you tried reproducing the issue with multiple visualizers?

There is a difference between having an object contain code for something and that object having ownership over it. I am not advocating for or against putting this code back in the MainForm, rather I am saying they don't need to be objects at all (well they might technically need to be for qt's sake depending on how you do it but that's an implementation detail). For example the dataset importer can be purely a few static functions rather than an object with its own state which needs to be tracked and passed around.

> Yes, noted this in the PR submission. I'm only able to reproduce this on older versions of VAPOR where we need to "End Capture Sequence" manually.

It sounds like the underlying issue has been resolved so that code can be removed.

sgpearse:
@Stasj - To remove state from the Capture/Import utilities, I've added new parameters so state can still be communicated to the MainForm.  These changes tracked in MainForm's event filter during ParameterChangeEvents.

apps/vaporgui/CaptureUtils.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+    // Add zero padding if we are capturing TimeSeries
+    if (ap->GetValueLong(AnimationParams::CaptureModeTag, AnimationParams::SingleImage) == AnimationParams::TimeSeries) {
+        std::ostringstream oss;
+        oss << FileUtils::RemoveExtension(FileUtils::Basename(file)) << std::setw(5) << std::setfill('0') << "." << FileUtils::Extension(file);
+        file = FileUtils::JoinPaths({FileUtils::Dirname(file), oss.str()});
+    }
+
+    // Warn user about overwriting file
+    if (FileUtils::Exists(file)) {
```
StasJ, line 58: This will miss overwriting later timesteps

apps/vaporgui/MainForm.cpp
```diff
@@ -879,7 +853,7 @@ void MainForm::sessionNew()
     _paramsMgr->UndoRedoClear();
 
     _stateChangeFlag = false;
-    _sessionNewFlag = true;
+    GetStateParams()->SetValueLong(GUIStateParams::SessionNewTag, "Toggle SessionNewTag to true for a new session", true);
```
StasJ, line 856: Please double check that extraneous renders are not triggered by the occurrences of replacing _sessionNewFlag by writing to the params database

apps/vaporgui/MainForm.cpp
```diff
@@ -957,6 +931,23 @@ bool MainForm::eventFilter(QObject *obj, QEvent *event)
...
+            return true;
+        }
+
+        // If PCaptureWidget sets the AnimationParams::AnimationStartedTag parameter,
+        // unset it, issue a call to _animationController->AnimationPlayForward(), and reject input
+        AnimationParams* aParams = GetAnimationParams();
+        if (aParams->GetValueLong(AnimationParams::AnimationStartedTag, 0) == true) {
+            aParams->SetValueLong(AnimationParams::AnimationStartedTag, "Disable tag", false);
+            _animationController->AnimationPlayForward();
+            return true;
```
StasJ, line 948: You are using the params database to essentially recreate signals/slots here.

StasJ (commented):
Thanks Scott. Just a couple more comments.

apps/vaporgui/MainForm.cpp
sgpearse: No problem.  The following sequence of events results in 3 calls to MainForm::Render on our main branch as well as this one, issue3778.

1) Launch VAPOR
2) Import any WRF file
3) Create new session

So, no extraneous renders found.

PR #3809, opened 2026-05-21
Fix issue #3614

### LINKED ISSUE #3614: Image file locations do not persist in SPACK generated sessions
To reproduce on Casper

1. Module load vapor
2. import any dataset, create an image renderer
3. save session, close vapor
4. open the session, and Vapor won't be able to find the image location

There seems to be a hash encoded in each SPACK instance that holds the path to or images directory.  This changes each time VAPOR is run.

Screen Shot 2024-05-17 at 4 06 20 PM

sgpearse:
Fix #3614.  The problem with spack also lies with the AppImage implementation, which mounts the VAPOR executable in a temporary directory that's inconsistent between instances.  This PR adds a fallback for when the image cannot be found, to check the currently mounted AppImage location for the image.

### DIFF
--- a/lib/params/ImageParams.cpp
+++ b/lib/params/ImageParams.cpp
@@ -69,16 +69,8 @@ std::string ImageParams::GetImagePath() const
     std::string defaultImage = Wasp::GetSharePath("images/NaturalEarth.tms");
     std::string savedPath = GetValueString(_fileNameTag, defaultImage);
 
-    if (!Wasp::FileUtils::Exists(savedPath)) {
-        std::string appDir = Wasp::GetEnvironmentalVariable("APPDIR");
-        if (!appDir.empty()) {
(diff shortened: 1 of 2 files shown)


lib/params/ImageParams.cpp
```diff
@@ -65,6 +67,18 @@ int ImageParams::Initialize()
 std::string ImageParams::GetImagePath() const
 {
     std::string defaultImage = Wasp::GetSharePath("images/NaturalEarth.tms");
+    std::string savedPath = GetValueString(_fileNameTag, defaultImage);
 
-    return GetValueString(_fileNameTag, defaultImage);
+    if (!Wasp::FileUtils::Exists(savedPath)) {
+        std::string appDir = Wasp::GetEnvironmentalVariable("APPDIR");
```
StasJ: This should be handled inside of `Wasp::GetSharePath`

StasJ (changes requested):

StasJ (approved):

